PR #1, opened 2017-02-15
WIP Initial code copied from pywbem

KSchopmeyer:
We are at a point where the functionality is significant and it accomplishes the basics of handling all of the WBEM operations (except create/modify/delete class) and most of the server operations defined in our server class.
I am ready to consider merging this into the master (Just so we can get on with issues rather than one big code block) with the following issues in the code:

1. We have not really reviewed it.
2. It has real issues in the choice of the single char options. Some of the choices are just to satisfy uniqueness.  
3. There are no mock tests. There is a unittest of help and the beginning of a live test of operations and a shell file that runs through just about all of the operations I have in place. It is this shell file that I am moving to the live test. I am working on mock tests but that turned out to be messier than I thought.
4. There are a lot of TODOs most of which are reminders to me.
5. We need to discuss the Copyright statement. Who should be copyright holder
6. The password handling is primitive but I have separate issue for that.
7. It currently fails travis test with 2.7
8. Name not changed for repository but I wanted to do that after merging this branch


andy-maier:
We agreed to merge this now despite it not being done completely, in order to have smaller delta changes.

PR #10, opened 2017-03-06
Fix license statement on files and move use to pbr for versioning

KSchopmeyer:
Add pbr as version mgr in _version.py. Note that version tag not set in git


KSchopmeyer:
+1

Issue #11, opened 2017-03-06
Finish create instance code


KSchopmeyer:
The current create instances code is missing two things:

1. tests for some of the more exoctic property values such as string with escapes.

2. Ability to create reference properties

3. Ability to create embedded instance properties.

4. Ability to create properties with information on type, etc. so that the class does not need to be referenced to build an instance

PR #12, opened 2017-03-08
fixes issue #11 Fix property code for instance create

KSchopmeyer:
Fixed the code for instance create. Moved the create instance code into
_common so can be tested as unit test, added array processing.  Added
run test for create instance against pegasus.


KSchopmeyer:
+1

PR #13, opened 2017-03-14
Fixes #9 Add functionality of help, repl, history

### LINKED ISSUE #9: Is there a way to get help, etc. rather than blank cmd line when pywbemcli entered.
When you execute just pywbemcli with no arguments it arrives directly at the repl prompt with no other info.

That is a real hole in the overall friendly environment.

Should we add something like a help, etc. at that point.

KSchopmeyer:
Add functionality to do help as part of startup with no options and do cmdline history when repl.

### DIFF
--- a/pywbemcli/_cmd_class.py
+++ b/pywbemcli/_cmd_class.py
@@ -20,11 +20,13 @@ cmds for get, enumerate, list of classes.
 from __future__ import absolute_import
 
 import click
+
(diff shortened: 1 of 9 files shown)


KSchopmeyer:
Note that this fails on python 2.7 because of the pywbem install.  That is written up as another issue

KSchopmeyer:
+1 despite the failure since that is really a pywbem failure

PR #17, opened 2017-03-21
Fixes #16: Add delete class to pywbemcli

### LINKED ISSUE #16: Add delete class to pywbemcli
I think we really do need a class delete, if only for testing.  This serves as an alternate to using the mofcompiler delete support tool

KSchopmeyer:
While we do not want to add the create and modify class operations, the
delete class seems logical and simple. Probably will not be used much
but I already found a use for in in tests.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,3 +1,4 @@
 Andreas Maier <andy-maier@users.noreply.github.com>
+Andreas Maier <maiera@de.ibm.com>
 Karl Schopmeyer <k.schopmeyer@swbell.net>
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
This pr fails with pywbem 2.7.  That issue was fixed as part of pr #24.  It appears to actually be an issue with coverage in that for python 27 coverage is outputting a compatibility warning message from configparser.py that happens to hit one statement I have in the code.

andy-maier:
+1

Issue #14, opened 2017-03-14
Travis Install fails with python 2.7 because of pywbem


KSchopmeyer:
Travis CI fails with python 2.7 because M2Crypto install fails.  This is caused by a failure to
find a particular variable in OpenSSL during the install of pywbem.

ImportError: /home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/M2Crypto/__m2crypto.so: undefined symbol: SSLv2_method

make: *** [install] Error 1

andy-maier:
I have created PR #20 to see whether a particular change I found in the master branch of M2Crypto solves our problem. While PR #20 fails for other reasons, it does successfully get over the import of M2Crypto into pywbem.

So the M2Crypto code version in its master branch solves the problem for us. I have opened M2Crypto issue 165 (https://gitlab.com/m2crypto/m2crypto/issues/165) to request a new release of M2Crypto.

In the mean time, we could merge our PR #20 to use the master branch of M2Crypto to get our Travis runs and our users going again.

andy-maier:
The final solution will then be:
- Wait until a new version of `M2Crypto` has been released on Pypi.
- Update `setup.py` of `pywbem` to require that new `M2Crypto` version, and release new fix versions of all supported stable versions of `pywbem` on Pypi.
- Update `requirements.txt` of `pywbemtools` to require that new version of `pywbem`.

KSchopmeyer:
I committed Andy's temp fix but am keeping the issue open pending a permanent fix from the m2crypto people.

andy-maier:
M2Crypto 0.26.0 contains the permanent fix for that. Both pywbemtools (as of PR #100) and pywbem (master) currently require M2Crypto>=0.24.0 (on non-Windows).

Need to verify Travis results for minimum package levels.

KSchopmeyer:
Closed because of changes to pywbem.

PR #20, opened 2017-03-22
Using M2Crypto master branch to pick up sslv2_method fix

andy-maier:
Please review.

This is a temporary circumvention for the M2Crypto import issue described in issue #14. It should be merged to get going again.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 
 six # MIT
 pbr>=1.8 # Apache-2.0
-pywbem # gpl
+-e git+https://gitlab.com/m2crypto/m2crypto.git@master#egg=M2Crypto
 click # BSD
 click-spinner>=0.1.5 # MIT
 click-repl # MIT


andy-maier:
While this PR fails for other reasons, it does successfully get over the import of M2Crypto into pywbem.

So the M2Crypto code version in its master branch solves the problem for us. I have opened M2Crypto issue 165 (https://gitlab.com/m2crypto/m2crypto/issues/165) to request a new release of M2Crypto.

andy-maier:
+1, because m2Crypto import now succeeds. The errors are unrelated to the issue that was addressed.

KSchopmeyer:
The failure that is now occurring is:
```
    def test(self):  # pylint: disable=missing-docstring
        command = 'pywbemcli %s --help' % (cmd_str)
        args = shlex.split(command)
        proc = Popen(args, stdout=PIPE, stderr=PIPE)
        out, err = proc.communicate()
        exitcode = proc.returncode   
        self.assertEqual(exitcode, 0, '%s: ExitCond Err, cmd="%s" '
                         'exitcode %s' % (tname, command, exitcode))
        self.assertEqual(err, "", '%s stderr not empty. returned %s'
>                        % (tname, err))

E       AssertionError: test_qualget stderr not empty. returned /home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/coverage/config.py:29: DeprecationWarning: You passed a bytestring as `filenames`. This will not work on Python 3. Use `cp.read_file()` or switch to using Unicode strings across the board.

E         return configparser.RawConfigParser.read(self, filename, **kwargs)

It is coming from the processing in one of the tests where we execute calls with help on pywbemcli itself. I suggest that we commit the m2crypto change and I will sort out the other issue

+1

KSchopmeyer:
I do have one comment.  We should not in the code why we are taking the code from the repository and put a TODO on it.

Issue #22, opened 2017-03-23
Fix several exceptions


KSchopmeyer:
We did not universally use the click exception in the action functions which means that in some cases, an exception in repl mode kills the client.

KSchopmeyer:
This should have been closed on the commit. Closing

PR #23, opened 2017-03-24
fixes issue #22 , add click exceptions. Also number of cleanups, minor fixes, etc.

KSchopmeyer:
1. insure that the click exception is in path for all cmds so they do
not drop out of repl.

2. Clean up cmd help to try to get short statement on first line

3.Correct number of issues in test_ops.sh

4. Correct tests in test_ops.sh

5. update help_overview

6. Added options to the count command and cleaned up display

### DIFF
--- a/design/help_overview.md
+++ b/design/help_overview.md
@@ -6,728 +6,215 @@ pywbemcli  --help
 =================
 
 ```
-Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS]...
-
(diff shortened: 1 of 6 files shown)


KSchopmeyer:
+1 I marked this one as OK despite the issues with integration.  It is strictly cleanup of things I needed to do (getting from prototype to finished code) in existing code and does not add new functionality so I am going to go ahead and mark it approved assuming that Andy is buried in other things now.

PR #24, opened 2017-03-26
[WIP] Fixes #6: Password fixes

### LINKED ISSUE #6: Handling passwords
The initial code leaves the password logic ompletely open in that we simply take the password from a global option input and even allows saving it in an env varaible.  That is unacceptable so we need to sort out what we should do with passwords to allow either open passwords (i.e. for some test environments) and protected passwords when the user requires them.
Proposal:

1. Make inputting the password an option.  If the user wants to supply it with a global option, that is ok.  If, they want to, they can supply the user name but not the password and we will ask for the password.

KSchopmeyer:
Add password logic to get password only if needed and to allow option of inputting password on command line or with prompt if it is required.

We actually do persist the server info in this code but the persistence is a hack, it saves it in a pickle file and saves too much info.  This is temporary to work through a bunch of the issues of working with the password and persisting the server info.  There is another issue open to propertly add configuration.  I would like to commit this and then add the config file capability rather than adding the config file to this pr.
 
 Cleanup password handling and use of context.
    
    1. We were incorrectly using the ctx.obj in that we were causing a
    WBEMConnection to be created for each subcommand.
    
    2. Added the function to get the password from a prompt when it is
    required
    
    3. Reorganized code to move all the code associated with the Click
    context object into its own file _context_obj.py.  Also all the WBEMConnection code is
     is organized into a new class PywbemServer.  Note that this just manages the connections and not the request calls.  They pass through to pywbem directly.

    Documented bug in changes.rst

4. Created new subcommands in the connection group to show the attributes of the connection, and export the values. 

This code DOES NOT expand to allow the capability to switch between connections since that is in a separate pr.

WIP I have one more change to make before finalizing this work.


design/help_overview.md
```diff
@@ -6,21 +6,21 @@ pywbemcli  --help
...
...ent private key file. (EnvVar:\n                                  PYWBEMCLI_KEYFILE).\n  -o, --output-format [mof|xml|table|csv|text]\n                                  Output format (Default: mof). pywbemcli may\n                                  override the format choice depending on the\n                                  operation since not all formats apply to all\n                                  output data types\n  --debug                         Display transmited and received packets\n                                  TEMP.\n  -v, --verbose                   Display extra information about the\n                                  processing.\n  --version                       Show the version of this command and exit.\n  --help                          Show this message and exit.\n\nCommands:\n  class       Command Group to manage CIM Classes.\n  connection  Command group to manage WBEM connections...\n  help        Show help message for interactive mode.\n  instance    Command Group to manage CIM instances.\n  qualifier   Command Group to manage CIM...\n  repl        Enter interactive (REPL) mode (default) and...\n  server      Command Group for WBEM server operations.\n'
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

design/help_overview.md
```diff
@@ -6,21 +6,21 @@ pywbemcli  --help
...
...splay a single class from the WBEM Server\n\nOptions:\n  -l, --localonly                 Show only local properties of the class.\n  --includequalifiers / --no_includequalifiers\n                                  Include qualifiers in the result. Default is\n                                  to include qualifiers\n  -c, --includeclassorigin        Include classorigin in the result.\n  -p, --propertylist <property name>\n                                  Define a propertylist for the request. If\n                                  not included a Null property list is defined\n                                  and the server returns all properties. If\n                                  defined as empty string the server returns\n                                  no properties. ex: -p propertyname1 -p\n                                  propertyname2 or -p\n                                  propertyname1,propertyname2\n  -n, --namespace <name>          Namespace to use for this operation. If\n                                  defined that namespace overrides the general\n                                  options namespace\n  --help                          Show this message and exit.\n'
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

design/help_overview.md
```diff
@@ -212,9 +212,51 @@ pywbemcli server profiles --help
 b'Usage: pywbemcli server profiles [COMMAND-OPTIONS]\n\n  Display profiles on the current WBEM Server.\n\n  This display may be filtered by the optional organization and profile name\n  options\n\nOptions:\n  -o, --organization <org name>   Filter by the defined organization. (ex. -o\n                                  DMTF\n  -n, --profilename <profile name>\n                                  Filter by the profile name. (ex. -n Array\n  --help                          Show this message and exit.\n'
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

design/help_overview.md
```diff
@@ -212,9 +212,51 @@ pywbemcli server profiles --help
...
 ```
 
+pywbemcli connection --help
+===========================
+
+```
+b"Usage: pywbemcli connection [COMMAND-OPTIONS] COMMAND [ARGS]...\n\n  Command group to manage WBEM connections.\n\n  These command allow viewing and setting connection information.\n\n  In addition to the command-specific options shown in this help text, the\n  general options (see 'pywbemcli --help') can also be specified before the\n  command. These are NOT retained after the command is executed.\n\nOptions:\n  --help  Show this message and exit.\n\nCommands:\n  export  Export the current connection information.\n  select  Select a connection from the current defined...\n  set     Show the current connection information, i.e.\n  show    Show the current connection information, i.e.\n  test    Execute a simple wbem request to test that...\n"
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

design/help_overview.md
```diff
@@ -212,9 +212,51 @@ pywbemcli server profiles --help
...
+```
+
+pywbemcli connection show --help
+================================
+
+```
+b'Usage: pywbemcli connection show [COMMAND-OPTIONS]\n\n  Show the current connection information, i.e. all the variables that make\n  up the current connection\n\nOptions:\n  --help  Show this message and exit.\n'
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

design/help_overview.md
```diff
@@ -212,9 +212,51 @@ pywbemcli server profiles --help
...
+```
+
+pywbemcli connection export --help
+==================================
+
+```
+b'Usage: pywbemcli connection export [COMMAND-OPTIONS]\n\n  Export  the current connection information.\n\n  Creates an export statement for each connection variable and outputs the\n  statement to the conole.\n\nOptions:\n  --help  Show this message and exit.\n'
```
andy-maier: This results in a very long line, not only in the source file but also in the rendered output. Split accordingly.

KSchopmeyer:
Whoops. I should have looked at the output after the last set of changes to the tool. It is printing the EOLs instead of executing them so that the result in the .md file is a mess.

KSchopmeyer:
At this point the help-overview should be readable.

andy-maier:
Contains a whole bunch of stuff which makes it difficult to review.
We agreed to merge this without further review and go forward from here.

Issue #27, opened 2017-03-31
The helpoverview display is a mess


KSchopmeyer:
The output of this tool is a mess because it does not format correctly for .md file.  The help output itself get presented as a single line.  Sadly it correctly formats with some .md formatters but NOT with the github one. Thus, the format in the firefox .md formatter works

I propose to a) change to use rst output rather than .md and correctly format for .rst.  This will also mean that we can use the result more generally than just for this temporary file.

KSchopmeyer:
Close as corrected by pr #28 which was merged

PR #28, opened 2017-03-31
fix issue #27 Clean up the help_overview output

KSchopmeyer:
the .md output only displayed on some markdown viewers.

It was just as easy to change to using .rst as the output so it now
produces .rst output.

Note that this is still completely manual so that to use it you must
enter

./help_overview.py >../design/help_overview.rst.

Goal is to:

1. Integrate the display into the docs.

2. Automate it so that it automatically finds every group and subcommand
similar to how the click man file generator contribution works.


KSchopmeyer:
+1. This is a bug fix, does not touch operational code but without it the help_overview format is a mess.

Looked at the help_overview.rst file in the repository and it looks like is should so going to commit this fix despite the fact that we still get the failure on python 2.7

PR #29, opened 2017-03-31
fixes #21  disable deprecation warnings to test file

### LINKED ISSUE #21: travis error AssertionError from coverage/config.py in one of our test files
Fix for m2crypto revealed another issue that shows itself in travis as:
```
    def test(self):  # pylint: disable=missing-docstring
        command = 'pywbemcli %s --help' % (cmd_str)
        args = shlex.split(command)
        proc = Popen(args, stdout=PIPE, stderr=PIPE)
        out, err = proc.communicate()
        exitcode = proc.returncode   
        self.assertEqual(exitcode, 0, '%s: ExitCond Err, cmd="%s" '
                         'exitcode %s' % (tname, command, exitcode))
        self.assertEqual(err, "", '%s stderr not empty. returned %s'
>                        % (tname, err))

E       AssertionError: test_qualget stderr not empty. returned /home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/coverage/config.py:29: DeprecationWarning: You passed a bytestring as `filenames`. This will not work on Python 3. Use `cp.read_file()` or switch to using Unicode strings across the board.

E         return configparser.RawConfigParser.read(self, filename, **kwargs)
```
It is coming from the processing in one of the tests where we execute calls with help on pywbemcli itself. I suggest that we commit the m2crypto change and I will sort out the other issue. Completed and closed the m2crypto pr but kept the issue open since the solution depends on taking from the the git for m2crypto, not pypy\i

Quick search shows the following bug from coverage.py

https://bitbucket.org/ned/coveragepy/issues/530/deprecationwarning-you-passed-a-bytestring

They say it is fixed for coverage 4.3.4 which is what we are using so not sure what the issue is so far.

KSchopmeyer:


KSchopmeyer:
Tried using the code to disable deprecate warnings for the test code but that does NOT work in this case because  this is not our code or a direct import but from a test tool, coverage.py and specifically from the config. py functionality.  

This occurs because apparently travis is loading an old version of coverage (version 4.0.3 per the Travis log) and the problem was not fixed until coverage 4.3

PR #32, opened 2017-04-07
Add password logic to get password only if needed

KSchopmeyer:
Cleanup password handling and use of context.

1. We were miss using the ctx.obj in that we were causing a
WBEMConnection to be created for each subcommand.

2. Added the function to get the password from a prompt when it is
required

3. Reorganized code to move all the code associated with the Click
context object into its own file.  Also all the WBEMConnection code is
not in that same file since it is only used in the context of this
ContextObj class

4. Moved code and data associated with the wbem connection into a
separate object (PywbemServer) that centralizes functionality and
data concerning WBEMConnection and WBEMServer and prepares for being
able to create multiple of these and name them so that the user can
persistently define connections and also maintain a base of
multiple connections persistently.

Add connection functionality to passwork pr. This is really
part of another issue but extends the whole use of the
PywbemServer class to be able to change connection in the
interactive mode. In order to make sense of this I put it into
this pr

Update the help_overview to reflect new commands. The merge
from issue #24 should have fixed the issues Andreas found in
the horrible formatting of the help_overview document in the
design directory.

Add temporary persistence functionality for the connections.
This is simplistic in that it uses a global dictionary
and pickles the dictionary for persistence.  This code
is in a separate file.

Adds to tests for persistence by executing pywbemcli
connection subcommands

Change assci table to separate header from lines

Add the help overview to the docs.

Extend changes.rst. Temp test in test_helpmsgs.py

Move logic for save connection info to its own file

Add test for connection_repository persistence by
createing test that executes pywbemcli and the
connection subcommands

### DIFF
--- a/pywbemcli/_cmd_connection.py
+++ b/pywbemcli/_cmd_connection.py
@@ -28,7 +28,7 @@ from ._pywbem_server import PywbemServer
 from .config import DEFAULT_NAMESPACE, DEFAULT_CONNECTION_TIMEOUT
 from ._asciitable import print_ascii_table
 from ._connection_repository import get_pywbemcli_servers, \
-    server_definitions_file_save
+    server_definitions_create_new, server_definitions_delete
 
 
 @cli.group('connection', options_metavar=CMD_OPTS_TXT)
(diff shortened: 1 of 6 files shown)


KSchopmeyer:
+1

PR #34, opened 2017-04-11
Fix issue  with issue 6 commit

KSchopmeyer:
I had an issue in the previous commit for the temporary persistence logic. This fixes that logic


KSchopmeyer:
+1. I am committing this one now because a) the code itself is temporary, b )need for demo tomorrow, c) need to do next set of changes.

PR #40, opened 2017-05-16
Fixes #33: Remove names subcommands

### LINKED ISSUE #33: Eliminate the names subcommand on class and instance
We originally had both enumerate and  names but combined the names into the enumerate with the option to return only names.   However, I did not remove the names subcommand.

Lets just remove those subcommands.

KSchopmeyer:
Removes instance name command from subcommands since this functionality is covered by the enumerate with an option.

Removed from tests also.

Removed one noqa message.

Reguires pr#44 to complete tests.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -17,4 +17,7 @@ exclude =
 #disable_noqa
     
 ignore =
+    # unable to detect undefined names (when using wildcard import)
+    F403
+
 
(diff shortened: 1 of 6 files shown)


KSchopmeyer:
we are getting the following error from pywbem 2.7

 E       AssertionError: test_serverprof stderr not empty. returned /opt/python/2.7.13/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/sphinxcontrib': missing __init__.py

E         file, filename, etc = imp.find_module(subname, path)

The problem is that, for some reason, the sphinxcontrib package is outputing an import warning when we have the test set to expect nothing back from the test.  I had this issue in my local build but cleaned it up by insuring that pip, wheel, setuptools were up-to-date but that does not appear to be the issue here.

andy-maier:
+1

PR #41, opened 2017-05-17
Fixes #26: Implement the pywbemcli version of the iter methods 

### LINKED ISSUE #26: option to allow selection of pull/non-pull operations
Today pywbemcli exclusively uses the original enumerate operations, It does not use the pull or iter commands at all.

I would like to change it to always use the iter commands since that allows us the freedom to use pull when possible.  I propose also to add one option so that the user could select use of either pull or non pull since that would provide a test of pull vs. non-pull operations

KSchopmeyer:
Implement the pywbemcli version of the iter methods that include the iteration and return complete results

These methods use the iter operations and then iterate the returned
generator.  This means that operations that use these operations can be
morphed to do either traditional operations or pull operations from
pywbemcli parameters and return the request results in a single call.

Replace the calls to EnumerateInstances, EnumeateInstanceNames, Refereces
ReferenceNames, Associators, AssociatorNames, ExecQuery in pywbemcli
with calls to the methods in the pyWBEMcliWBEMConnection class

Creates a new class and methods to allow easily using pull vs. non-pull
operations with only a parameter chang.  This adds two new options to
the command line to define whether pull or not pull should be used (normally this is not used and the default is to first try pull), and to define the size of pull responses

This also fixes minor issues in the help output and it support in doc building as well as remove some pylints and clean up tests.


KSchopmeyer:
Same issue as other pr open in that the test that calls script pywbemcli returns the following where it should be returning an empty string. On travis the code is issuing this warning msg when pywbemcli executed.

E       AssertionError: test_top stderr not empty. returned /opt/python/2.7.13/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/sphinxcontrib': missing __init__.py

E         file, filename, etc = imp.find_module(subname, path)

andy-maier:
+1

Issue #42, opened 2017-05-26
Execution of pywbemcli in test returns importwarning in stderr


KSchopmeyer:
Execution of the pywbemcli script in via the makefile always results in a return in stderr:

E AssertionError: test_top stderr not empty. returned /opt/python/2.7.13/lib/python2.7/pkgutil.py:186: ImportWarning: Not importing directory '/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/sphinxcontrib': missing init.py

This is because the make test sets PYTHONWARNING = 'default' and the package sphinxsupport generates the warning.

We propose to resolve this by turning of the PYTHONWARNING in those tests that specifically execute pywbemcli.

KSchopmeyer:
Closed with merge of pr #44

PR #44, opened 2017-05-26
Add code to ignore python warnings for test execute pywbemcli

KSchopmeyer:
Issue #42. Add code to test execution of pywbemcli to ignore warnings
because of importwarning from sphinxcontrib

This change sets PYTHONWARNING to "" for tests that directly call the pywbemcli executable so that
the importwarning from sphinxcontrib is not generated.

This means that we have no real test for python warnings at the level of execution of the pybemcli script itself.

Set as high priority since this issue is breaking the CI environment.


KSchopmeyer:
+1

Issue #50, opened 2017-07-20
Clean up display output to better output information as tables


KSchopmeyer:
Today the output is a mishmash.  We want to output data as tables with headers wherever possible and to allow multiple table structures.   We wanted to use tabulate since this flexibly handles tables but it does not handle multi line cells and we will require that. 

Therefore, it would seem that the best solution for now is

a. Accept the tabulate table structure types, and switch to   tabulate for table generation
b. If we need multirow cells, take some of the work from another project that does produce multiline cells but does not have all of the options of tabulate and use it when we specify tables as multiline rows.

I am trying to sort out the relation between output formats and what gets produced.  This is made more difficult for the user to understand because we have both data output that it clearly display table oriented 
(lists and lists of lists) and an data that has an internal structure that we want to present (CIMInstance, etc.). Clearly not all of the output forms apply to all of the possible output.  Thus, 'mof' as an output type only applies to a CIMInstance, etc. data.  Table and the various table forms applies to both lists, and objects. We present instances/classes as a table (properties as columns). But if we use the 'html' type, it
is unclear what we mean.  In fact today what we mean is html table of the properties, not an html representation of the CIM Instances.

Right now the output formats are ['mof', 'xml','table', 'plain', 'simple', 'psql', 'rst', 'mediawiki', 'html']

Where all of the names are for table display except for mof and xml which only apply to CIMInstance, etc. and table, plain, etc. apply also at least to CIMInstance.  CIMClass it is clear probably mostly makes sense
as MOF or xml (or future htmlMOF)

Right now, it accepts table, html, etc. for any data that can be displayed as a table including CIMInstances
and forces CIMClasses to mof unless outputformat=xml irrespective of the defined output type.

DISCUSSION: Have we picked the output types correctly with this overlap and confusion.
```
Data types cimInstance/CIMClass/CIM/Param       CIMObjectName               Data Lists(namespaces, etc)   
Formats     This is a complex structure                   Normally displayed           List or list of lists
                                                                             as list

MOF                          Yes                                       No                                     No

XML                           Yes                                      Yes                                     No

Table              Instance Properties                          Not sure                             Yes
simple,
etc.

HTMLTable     Instance Properties                          Yes                                     No
                     (same as table above but
                     in html)

HTML           HTML form of mof,                          Not sure                             Yes, table rendered in html
data             (Future)
```

andy-maier:
Addressed in PR #54.

PR #52, opened 2017-07-20
Fixes #47: Capturing and displaying statistical information

### LINKED ISSUE #47: enable and display pywbem statistics results
Pywbem 0.11.0 enables statistics output from pywbem operations.  Enable statistics as an option and display the results in pywbemcli

KSchopmeyer:
Adds statistics capture and display to pywbemcli.  To do this it depends on the current pywbem development track so that the requirements.txt requests download and install of the repository, not the pip
pywbem.

An output looks like:
```
 Count  Exc  Time   ReqLen  ReplyLen  Operation 
------------------------------------------------
 1      0    0.077  532     69408     GetClass  
```
or if there are differenses between avg/min/max it looks like:
```
 Count  Exc  Time               ReqLen       ReplyLen        Operation              
------------------------------------------------------------------------------------
 3      2    0.011/0.002/0.021  338/321/348  4472/298/12814  EnumerateInstanceNames 
```

There is an additional line of information (time in sec, and what the xxx/xxx/xxx means that appears as a title if verbose is set.  We did not want this to show up for every display.

### DIFF
--- a/design/help_overview.rst
+++ b/design/help_overview.rst
@@ -2,4 +2,5 @@ The help overview has been incorporated into the pywbemtools
 documentation
 
 To see what was originally in this file look at
-    docs/pybemclicmdshelp.rst
+
+docs/pybemclicmdshelp.rst
--- a/design/help_overview.rst
+++ b/design/help_overview.rst
@@ -2,5 +2,4 @@ The help overview has been incorporated into the pywbemtools
 documentation
 
 To see what was originally in this file look at
-
-docs/pybemclicmdshelp.rst
+    docs/pybemclicmdshelp.rst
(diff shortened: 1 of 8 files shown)


KSchopmeyer:
I have one bug in this now.  The extra line display is not showing up. Want to fix that before any approval

-1

andy-maier:
-1
Comments:
1. Not clear to me what commit 76696fd has to do with this PR, logically. If it is logically separate, can you please separate it out into a separate PR?
2. The other commit 1284fe5 seems to contain more changes than what is related to the statistics output. For example, it seems to contain some of PR #54. If you depend on PR #50, then rebase this PR properly on it, but don't include the code in this commit.

KSchopmeyer:
Karl made MAJOR changes to this code to simplify it and reduce it to just the necessary.  We need a review again.

PR #53, opened 2017-07-20
Fixes #30: Fixed tool to generate click help info from program

### LINKED ISSUE #30: Create a tool/support extension that will automate outputting overall help info.
Issue: While help is easy to view on a piecemeal basis, these programs become big enough that it is difficult to view the commands/groups/subcommands in an overall context and to actually search for information.

Goal: A tool or extension to click that will create output that represents a complete set of the help for an application like pywbemcli.   This would output the help for every command and group.

There are two ways to do this:

1. Imitate what the click man page contrib did by extending setup so that the file is created through the setup functions.

2. Add another component to help so that we end up with some sort of new command or option on help to do help with some options (ex. all).

KSchopmeyer:
Moved from using list in tools program to executing program to generate click
information.  This runs the click program with the hiearchy of subcommands and creates
a file with the help output.  It can create either a text file or an rst file with a separate
subsection for each subcommand.

ISSUE: the help is built from a target in the makefile.  However, since smicli itself is built and is not part of the git commit, this help builder fails if smicli does not exist and right now the makefile prerequisites doe not cover that.


andy-maier:
Two comments:
- There are two versions of click_help_capture.py, one in tools and one in tools/tools.
- On the dependency on pywbemcli in the Makefile change: One thought is to import the pywbemcli module into the click_help_capture.py script instead of executing the pywbemcli script.

KSchopmeyer:
I saw review and commented. Do I have to dismis review???

KSchopmeyer:
+1

PR #54, opened 2017-07-26
Fixes #50: Create a more general table output mechanism

### LINKED ISSUE #50: Clean up display output to better output information as tables

KSchopmeyer:
Creates a simple table formatter that serves to format ascii tables

Using terminaltable as the formatter because today does NOT support multiline cells.  However, the formatter itself can be modified to use tabulate when the multiline cell version is available.

Since Terminaltable only supports pure ascii tables we only define those types for our output.

Adds output_format cmd line parameter to pywbemcli.

Changed the functions in _cmd_server to output tables in place of the
simple print output that existed.

Thus, cim_objects and lists of cim_objects are output through the display_object fromatter.  All other data is displayed as tables, even namespace lists (one column tables) and uses the table formatter.

Adds unit test for tableformatter (test_tableformatter.py)


KSchopmeyer:
There are some items that are not directly the table change. Karl will weed those out to a separate pr.

andy-maier:
Some high level comments:
- Remove anything from this PR that is not related to the logical change it does.
- In a number of cases, it seems the code using TableFormatter could have used a fixed-up display_table() function. Please have a look at display_table() and display_cim_objects() and see whether they can be improved and then used again.

KSchopmeyer:
This has been cleaned to remove changes not pertinent to this cr, simplify the code, etc.

+1

PR #56, opened 2017-08-05
Fix error in flake8 noqa definitions

KSchopmeyer:
We are including space after comma in some flake8 noqa defintions.
That is not allowed all of a sudden so generates a number of flake8
issues with __init__.py


KSchopmeyer:
Set as priority because flake8 errors block everything. 

+1

PR #57, opened 2017-08-05
Fix pywbemcli.py help doc issue

KSchopmeyer:
A missing empty line in the cli comment for pywbemcli was
causing formatting issues with help output. Added
space and rebuilt the help output

Correct issue with creating help info for docs

We were indenting all lines from the help.  This meant that empty
lines end up with spaces. Added code to clear any line that
only have spaces.


KSchopmeyer:
+1

PR #61, opened 2017-08-16
add single character -h as help option alternative

KSchopmeyer:
Extend click so it accepts either -h or --help as help option throughout the commands and subcommands of pywbemcli


KSchopmeyer:
There was a single line conflict that for some reason did not show up in my local copy, old an new statements on same line.  We wanted both of them.

Fixed right here.

andy-maier:
+1

PR #62, opened 2017-08-23
Fix issue with bad table print code

KSchopmeyer:
When we converted to new table formatter, this one did not get moved. Breaks tests now


KSchopmeyer:
I merged this one because it was a) simple, b) my booboo in an earlier commit c) stopping other commits. It was only 2 lines of code change that was part of the use of the table formatter. I had forgetten one table in the process of the simple table formatter

PR #68, opened 2017-10-06
fixes #8 Add option to display summary information.

### LINKED ISSUE #8: Provide a means to do summary information for enumerates.
We found it useful in cimcli to just show counts of enumerates, etc. rather than the details.  That was a separate display option, however, so added to the overall complexity and number of options.  We need a better way.

Some of the possibilities are:

1. Make this an output type like mof, xml, etc. (summary)
2. Add new operation.  There is an issue with this in that it means lots of new operations (enum summary, references summary, associators summary, etc.
3. Make it an option on the commands that use it, i.e. a flag --summary

KSchopmeyer:
Add option to commands that enumerate/reference/associator multiple objects to only display a summary of what was returned in place of details.


andy-maier:
+1

PR #74, opened 2017-10-13
WIP Initial tests of click CliRunner.

KSchopmeyer:
Creates a new tests/unit/test_class_operations.py program that executes several test using CliRunner.


KSchopmeyer:
This whole approach was a loser and I am withdrawing the PR.

PR #75, opened 2017-10-22
Correct issues with output_format definition

KSchopmeyer:
1. Changed output format definition to not set default until output.
This simplifies the issue between the mof structured displays and the
table displays.
2. Cleaned up code around display of paths to cover issue were the
classname return from enumerate classes is unicode, not CIMClassName.

3. Add new display function in common for display of paths including
classnames, instancenames associatornames, referencenames. This allows
displaying as table or just text output.

4. Minor fix to test_connection to use valid namespace for local tests.

5. Fix display cimobjects summary (-S) to correctly show the returned
with the count

6. Add table output to class find.


andy-maier:
+1

PR #86, opened 2017-11-03
Set test: off in appveyor.yml

KSchopmeyer:


KSchopmeyer:
That did not work.  Appears best way is to remove the web hook

PR #88, opened 2018-05-01
Add tests/function directory and some tests.

KSchopmeyer:
WIP - I am going to redo this one to put everything into a single tests/unit. 

That seems the easy way out. 

 Adds utility functions to execute pywbemcli both as a script and directly as
inline code in the function directory.

This does NOT add the detailed set of tests but the utilities and a minimal first test for the class subcommand to confirm that the utilities work.

Corrects change to pywbem 2.12.0, enable_stats to stats_enabled by changing
to correct variable throughout pywbemcli code.


KSchopmeyer:
Discussion 4 May.  Andy proposed that we not use the inline method because we should be able to test everything through the command line and not have to get into the internal configuration of any package we were using to test our code. 

Karl proposed leaving this as is but not using unless their are exceptional cases.  Try to do wverything through the command line.

KSchopmeyer:
I cancelled this pr completely and closed it since we will put everything in the tests/unit directory for now.

PR #89, opened 2018-05-02
WIP Adds support for pywbem_mock

KSchopmeyer:
Note that this pr is built on pr #88 test/function subdirectory since the goal was that the mock capability was to be used as the basis for most of the function level tests.

WIP today because the tests themselves have not been added. this code consists of just
the infrastructure components.

This pr rebased to pr ks/#xxx-create_funct_tests which adds
the tests/function directory the utils.py file that provides
functions to execute pywbemcli both as a script and as
in line code  and some initital tests. 

Addsupport for pywbem_mock by using a specific scheme in the host url to define a mock request.  If the scheme is 'fake' we assume that the netloc component is a keyword defining what is in the path and that the path is the data used to define the setup for the mock. Usually this will be a mof file with the mof to be compiled.

This also fixes an issue where we were using uri for the variable names
and it is really url.  Renamed the variables.

### DIFF
--- a/pywbemcli/_cmd_connection.py
+++ b/pywbemcli/_cmd_connection.py
@@ -238,7 +238,7 @@ def cmd_connection_export(context):
     context.spinner.stop()
     svr = context.pywbem_server
 
-    export_statement(PywbemServer.server_envvar, svr.server_uri)
+    export_statement(PywbemServer.server_envvar, svr.server_url)
 
     if_export_statement(PywbemServer.defaultnamespace_envvar,
                         svr.default_namespace)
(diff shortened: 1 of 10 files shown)


KSchopmeyer:
4 May 2018. Andy proposed that we drop trying to use the url to define a mock test and, instead do it through a global option. That seems logical and Karl will implement that immediatly.  This can be a repeated option so that we can set up multiple things as part of a test.

KSchopmeyer:
I canceled/closed this pr in favor of creating a new one that is simpler.

PR #90, opened 2018-05-06
Add basic mock support to pywbemcli

KSchopmeyer:
This adds the basic mock support to pywbemcli as follows:

1. Creates a new global cmd line option that allows defining multiple
filenames that will force the use of the mock support in place of wbem
connection and also allow adding mof to the repository or executing
arbitrary python code as part of the startup.  The option is
--mock_server and it can be defined multiple times on a single pywbemcli
call.  Each use of the option defines a single file name or file path
name.

2. Creates a first test that tests the basics of the class enumerate
using the mock support.  This confirms that mof can be installed and
also serves as the basis for futher tests.  NOTE: We want to move this
to use the new test_function when that is committed so the parameters
are very basic for the moment.

In addition this pr fixes one unreleated problem and changes code to use
url in place of uri to be consistent with pywbem.

The unrelated problem is that the enable_stats parameter in
cim_operations was changed to stats_enabled when the code was finalized

The other change was to expand the repr of one of the classes for more
complete presentation of the data in the class.


andy-maier:
Karl, please push a change to this PR to trigger Travis.

andy-maier:
We agreed that this one would go in without review.

Karl, please rebase this one now (PR #100 has been merged) to see whether it succeeds.

PR #98, opened 2018-05-07
Fix issue with stats_enabled param of WBEMConnection

KSchopmeyer:
In earlier versions of pywbem the experimental stats enabler was named
enable_stats. We renamed it to stats_enabled in the pywbem release and that breaks pywbemcli code.

Marked priority since it breaks tests

This picks up that change.


andy-maier:
Karl, please push a change to this PR to trigger Travis.

Issue #99, opened 2018-05-07
As of 7 May, Travis no longer runs on commit


KSchopmeyer:
As of this weekend the travis for pywbemcli does not run on the server. It was running late last week and even over most of the weekend.  In fact, it may be just today that it stopped

andy-maier:
The pywbemtools project was not active in Travis: 
![image](https://user-images.githubusercontent.com/8939758/39753103-e60af588-52bd-11e8-95fc-aede89a66a4f.png)

I activated it. Let's see if that helps.

andy-maier:
I verified that activation of the project in Travis solved the issue, using the manual-ci-run branch.

What remains to be done is that for each open PR that does not yet have a Travis run, to trigger the run by pushing up a change.

KSchopmeyer:
Thanks

Issue #51, opened 2017-07-20
make develop in new environment cannot import click on first try


KSchopmeyer:
For some reason running develop in new environment has a problem with the click install but it works the second time.  I assume there is an ordering problem in requirements.txt.

andy-maier:
This is a non-issue because "make install" needs to be invoked before "make develop", and if that is done, it works fine.

Also, this has been improved as of PR #100; "make develop" now invokes "make install" automatically if needed.

PR #100, opened 2018-05-07
Fixes #78 #99 #51 Fix build issues on Travis and Appveyor

### LINKED ISSUE #78: Apveyor fails because pywbem package requires non-Windows version of M2Crypto
The approach how the "pywbem" package specifies its prerequisite packages is by means of a complex setup.py, which takes care of pulling in"M2Cryto" on non-Windows and "M2CryptoWin32/64" on Windows. Its wheel archive does not makes that distinction and pulls in "M2Crypto" for all platforms. This can also be seen on Pypi: The dependencies of "pywbem" only mention "M2Crypto" without any platform marker.

This issue basically prevents pywbem to be specified as a dependent package on Windows.

This also the reason Appveyor for "pywbemtools" fails.

I could not find a place in the appveyor.yml to logically turn it off so the easiest was just to remove the webhook from github for the moment.

 https://ci.appveyor.com/api/github/webhook

Reset this into the repository when we want to make this work again

KSchopmeyer:
This was created to test some options when travis did not run.  The issue turned out to be that travis was not active any more.  That was fixed in travis.  However, this pr does try to bring pywbemcli .travis.yml in line with the pywbem file.

NOTE: It will not execute until either we cherry-pic pr #98 or it is merged and this pr rebased.  I am not going to update this one yet.

QUESTION: Is this correct update to yml Andy since you are the expert.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.settings/
 .pydevproject
 /.cache/
+/.pytest_cache/
 /.tox/
 /htmlcov/
 /build/
@@ -13,6 +14,7 @@ ChangeLog
 *.pyc
 *.log
 *.tmp
+*.done
 tmp.*
 pywbemcli.egg-info
 MANIFEST
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ setuptools-*.zip
 /tmp_*/
 /dist/pywbemcli*.tar.gz
 /dist/pywbemcli*.whl
+/pywbem_os_setup.sh
 tmp_*
 *,cover
 cover
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,14 @@ sudo: required
 # not work. Also, on OS-X, it needs to be invoked with language=generic
 # and an empty 'python' variable in order to prevent that Travis attempts
 # to install Python.
+# Note: pypy is excluded because M2Crypto does not install.
 matrix:
   include:
+#    - os: linux
+#      language: python
+#      python: "2.7"
+#      env:
+#        - PACKAGE_LEVEL=minimum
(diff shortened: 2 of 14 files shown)


andy-maier:
I have triggered Travis with an amend + force-push (no content change) to see whether it runs fine.

andy-maier:
Karl, I cannot set you as the reviewer of this PR (probably because you created it), but please review the changes that I made to it.

KSchopmeyer:
1 Forget my comment about the help... program on slack this morning.  I misread that.

2. You know so much more about the details of the requirements, etc. that I can only follow and make minor critique and this point I did not really have even any minor critique.

andy-maier:
We need to support remaking that rst file in the CI runs, because it is (and should be) part of the make tree.

If we want to use the version of pywbemcli that is in the repo, then it cannot be any form of a "pywbemcli" shell command, because that command is only created by the package installation (and then not found on Windows, currently). Instead, it will need to be an invocation of "python pywbemcli/cli.py" or something similar. We could change the click_help_capture.py script to do the invocation that way, which removes the dependency on the installation functioning.

However, we still need to get the installation functioning at some point.

andy-maier:
I have created issue #103 for the issue that pywbemcli is not found on Windows, and there are now Todo markers in the code where there are circumventions of that issue (i.e. `TODO #103`).

The entiere `make test` has been disabled when running on Appveyor. It is active when running on a local Windows that is not Appveyor, so it can be debugged.

andy-maier:
As far as I'm concerned, we can merge this PR. Issue #103 is open for the remaining issue of pywbemcli not found on Windows.

andy-maier:
We agreed to merge this one, once it passes.

PR #101, opened 2018-05-09
Fixes #95 Change output table formatting package to use tabulate

### LINKED ISSUE #95: Move pywbemcli to tablulate package
When we left off last fall we had not yet moved pywbemcli from a temporary table output package to the tabulate package.  We propose to do that before the initial release.

KSchopmeyer:
Simple change to use package we always wanted to use for table formatting and to fix minor changes in tests because the two packages formatted table output slightly differently.

See commit message for more information

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,5 @@ click>=6.6
 click-spinner>=0.1.6
 click-repl>=0.1.0
 asciitree>=0.3.3
-terminaltables>=3.0.0
+tabulate>=0.8.2
 prompt-toolkit>=1.0.9
-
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -14,12 +14,11 @@ click>=6.6 # BSD
 click-spinner>=0.1.6 # MIT
 click-repl>=0.1.0 # MIT
 asciitree>=0.3.3 # MIT
-terminaltables>=3.0.0 #  MIT
+tabulate>=0.8.2 # MIT
```
andy-maier: This change also needs to be done in the new `minimum-constraints.txt` file added with PR #100.

andy-maier: DONE.

andy-maier:
Karl, I resolved my comment and rebased the PR to see whether it now succeeds. Nothing to do for you on this one.

KSchopmeyer:
++++++ 🤠

PR #102, opened 2018-05-09
fixes #78 Resync appveyor.yml with current version from pywbem

### LINKED ISSUE #78: Apveyor fails because pywbem package requires non-Windows version of M2Crypto

KSchopmeyer:
This changes appveyor.yml to exactly what is in the current
pywbem/appveyor.yml and attempts to get the appveyor tests running.

We had a problem defined in issue #78 but in the mean time apparenly the appveyor account also gone inactive. That was restarted

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,11 @@ configuration:
 
 install:
 
+  - git --version
   - if %APPVEYOR_REPO_BRANCH%.==manual-ci-run. set _NEED_REBASE=true
+  # This Git version requires user configuration in rebase step
+  - if %_NEED_REBASE%.==true. git config user.name "dummy"
+  - if %_NEED_REBASE%.==true. git config user.email "dummy@dummy"
   - if %_NEED_REBASE%.==true. git fetch origin master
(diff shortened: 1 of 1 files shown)


andy-maier:
I have cherry-picked this commit into the PR #100 that fixes Travis, because both need to be fixed in the same PR to let the check runs all succeed. This PR here can be closed without merging it.

PR #108, opened 2018-06-01
Refactor pywbemcli executable tests

KSchopmeyer:
Refactor the pywbemcli executable tests to create a common
base class for the tests largely using the mock environment
and modify the test for class to use this environment.

All tests are defined through parameters to the single method
that executes the test.

This turned out to be cleaner and more consistent that using
the wrapper function at this point.

The limits with this approach are:

1. The mock environment has to be created for each test which takes
time.

2. Only a single subcommand can be executed within a single test because
the repl today only support entering a single command at at time. Thus
tests like class delete can only confirm a good response, not that the
class was actually deleted in the mock environment.

Note that these tests are not complete today but include a number of
TODOs and test with the condition marked false.

In addition we found a couple of errors in the current code that we fixed
including:

1. On the class tree command, we were not stopping the spinner before
calling the display method.

2. We were using the click form
click.option('--includequalifiers/--no_includequalifiers'
to define the option for include qualifiers.  First it was
giving some problems and second it was just more complex
that we need since the real option is just to not include
qualifiers so I modified that to a simple flag with the name
--no-qualifiers for the getclass, etc. subcommands in the
class group.

3. We had the EnumerateInstances options DeepInheritance
mispelled as Deepinheritance so fixed that both for instance
and class subcommands.

### DIFF
--- a/pywbemcli/_cmd_class.py
+++ b/pywbemcli/_cmd_class.py
@@ -42,10 +42,11 @@ from ._displaytree import display_class_tree
 # the default is True which is what we want.
 # TODO: We need to document all these defaults.
 includeclassqualifiers_option = [              # pylint: disable=invalid-name
-    click.option('--includequalifiers/--no_includequalifiers',
+    click.option('--no-qualifiers', is_flag=True,
                  required=False, default=True,
-                 help='Include qualifiers in the result. Default is to'
-                      ' include qualifiers')]
+                 help='Do not include qualifiers in the response.'
+                      'The default behavior is to include'
+                      'qualifiers in the returned class.')]
 
 deepinheritance_option = [              # pylint: disable=invalid-name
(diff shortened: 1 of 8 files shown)


KSchopmeyer:
Andy.  I am fairly happy with the overall structure of the tests themselves.

I think I have too many common files (pytest_extensions, etc.) but that could go either way.

The issue here will be that with this configuration we can execute only a single pywbemcli command in one test of the executable that first creates the mock environment so we are continually recreating the mock environment. (I.e. tests take a lot of time).  Would be nice if there was a way to execute multiple subcommands in a single call to pywbemcli from the command line (ex. pywbemcli -h http://blah class get CIM_Foo; class delete CIM_Foo --mock-server simple_mock_file.mof).  But this is good enough to get us up to about 80% coverage I think.

PR #118, opened 2018-06-25
Update requirements.txt to only use version 1 of prompt-toolkit

KSchopmeyer:
Ready for review. New prompt-toolkit 2 is incompatible so we limited ourself to v 1.

See commit for details

I also added to help for repl that up and down arrows show history.

Added comment to requirements.txt  referenceing the issue number.

### DIFF
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -621,7 +621,6 @@ The following defines the help output for the `pywbemcli connection new --help` 
       Create a new named WBEM connection.
 
       This subcommand creates and saves a new named connection from the input
-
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
fixed issues and fixed broken test (pycheck issue). Note that we occasionaly pick up some strange changes in the docs/pywbemclicmdshelp.rst file because it is automatically generated by what is in the
code.  There may be cases where for some reason it was not automatically generated in a previous checkin and those things get in for another pr. That file simply reflects what is in the help, etc. for the code at the time it is generated.

KSchopmeyer:
Hope I fixed issues.

PR #119, opened 2018-06-25
Ks/fix mock option - Number of minor fixes to mock sever option. 

KSchopmeyer:
Ready for review

See commit messages for details

### DIFF
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -107,8 +107,7 @@ The following defines the help output for the `pywbemcli  --help` subcommand
                                       operations after  each command execution.
       -v, --verbose                   Display extra information about the
                                       processing.
-      --version                       Show the version of this command and exit.
-      --mock_server FILENAME          If this option is defined, a mock WBEM
+      -m, --mock-server FILENAME      If this option is defined, a mock WBEM
                                       server is constructed as the target WBEM
                                       server and the option value defines a MOF or
                                       Python file to be used to populate the mock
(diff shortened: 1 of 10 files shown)


docs/pywbemclicmdshelp.rst
```diff
@@ -654,6 +653,14 @@ The following defines the help output for the `pywbemcli connection new --help`
...
       -c, --certfile TEXT           Server certfile. Ignored if noverify flag set.
       -k, --keyfile TEXT            Client private key file.
+      -m, --mock-server FILENAME    If this option is defined, a mock WBEM server
+                                    is constructed as the target WBEM server and
+                                    the option value defines a MOF or Python file
+                                    to be used to populate the mock repository.
+                                    This option may be used multiple times where
+                                    each use defines a single file or
+                                    file_path.See the pywbemcli documentation for
+                                    more information.
```
andy-maier: There are now two descriptions for `--mock-server` in the help text.

KSchopmeyer: Yes.  The connection new allows you to build a connection internally to save and that includes all of the options that are in the original cmd line.

Would like to find a way to avoid this duplication but that would probably mean just deleting the new connection and  using the save on whatever connection the user had created.

I actually have no real issue with that.

KSchopmeyer:
Does the comment satisfy your issue? If not lets mark this for discussion.

andy-maier:
I'm fine with it. Please rebase and merge.

docs/pywbemclicmdshelp.rst
andy-maier: OK, I'm fine with it then.

PR #120, opened 2018-06-26
Add new method to use wbemuri on input of instancenames

KSchopmeyer:
Replace existing code with from_wbemuri to parse path inputs

Includes minor text changes for the text that becomes the help


docs/pywbemclicmdshelp.rst
```diff
@@ -277,21 +277,21 @@ The following defines the help output for the `pywbemcli class enumerate --help`
...
-                                      defined that namespace overrides the general
-                                      options namespace
-      -S, --summary                   Return only summary of objects (count).
-      -h, --help                      Show this message and exit.
+      -d, --deepinheritance     Return complete subclass hierarchy for this class
+                                if set. Otherwise retrieve only the next hierarchy
+                                level.
+      -l, --localonly           Show only local properties of the class.
+      --no-qualifiers           Do not include qualifiers in the response.The
+                                default behavior is to includequalifiers in the
```
andy-maier: Insert a space in "includequalifiers". Occurs multiple times.

docs/pywbemclicmdshelp.rst
```diff
@@ -191,9 +191,9 @@ The following defines the help output for the `pywbemcli class associators --hel
       -c, --resultclass <class name>  Filter by the result class name provided.
       -r, --role <role name>          Filter by the role name provided.
       -R, --resultrole <role name>    Filter by the role name provided.
-      --includequalifiers / --no_includequalifiers
-                                      Include qualifiers in the result. Default is
-                                      to include qualifiers
+      --no-qualifiers                 Do not include qualifiers in the
+                                      response.The default behavior is to
+                                      includequalifiers in the returned class.
```
andy-maier: Insert a space in "includequalifiers". Occurs multiple times.

andy-maier: The description talks about both "response" and "returned class". The response is not always a class, so this inconsistency is confusing. Can that be made consistent?

KSchopmeyer: This gets messy.  We separated the class level and instance level requests in the cli despite the fact that they call the same entity. Further, we concluded that for the class level, the default is to return qualifiers whereas the for the instance level it is nto not include qualifiers. Thus, in the class help the  flag is --no-qualifiers and there is no short form under the assumption that it is seldom used.

For instances, the world is inverted. Normally qualifiers are of no interest so the flag is -q/--includqualifiers and the default is to not return qualifiers.  Actually the return is always a class for the class level command and an instance for the instance level response. Note however, that the code does not actually enforce the limit for references and associatiors (i.e. we do not test for instance id)

Changed the word response to result. Changed the word class to objects. Note that this single option is defined once and applied to several different subcommands in the _cmd_class group.

KSchopmeyer: DONE. Think there was only one but it shows up in the help and tests also.

pywbemcli/_cmd_class.py
```diff
@@ -39,18 +39,20 @@
...
 # TODO: We need to document all these defaults.
 includeclassqualifiers_option = [              # pylint: disable=invalid-name
     click.option('--no-qualifiers', is_flag=True,
                  required=False, default=True,
-                 help='Do not include qualifiers in the response.'
-                      'The default behavior is to include'
-                      'qualifiers in the returned class.')]
+                 help='Do not include qualifiers in the result. '
+                      'The default behavior is to include '
+                      'qualifiers in the returned objects.')]
```
andy-maier: Now that I understand the difference between how we default the inclusion of qualifiers in returned classes vs. returned instances, I suggest that we call out "class" and "instance" specifically in the description. I.e. change the description to: "Do not include qualifiers in the returned classes. The default behavior is to include qualifiers in the returned classes."

pywbemcli/_cmd_instance.py
```diff
@@ -40,7 +40,7 @@
 # on instances but True on classes
 includequalifiers_option = [              # pylint: disable=invalid-name
     click.option('-q', '--includequalifiers', is_flag=True, required=False,
-                 help='Include qualifiers in the result.')]
+                 help='Include qualifiers in the returned objects.')]
```
andy-maier: Now that I understand the difference between how we default the inclusion of qualifiers in returned classes vs. returned instances, I suggest that we call out "class" and "instance" specifically in the description. I.e. change the description of this instance-level option to say "returned instances" instead of "returned objects".

Also, to be consistent with the wording of the description for classes, we may want to add a sentence about the default.

pywbemcli/_cmd_instance.py
```diff
@@ -184,11 +185,11 @@ def instance_enumerate(context, classname, **options):
     """
     Enumerate instances or names of CLASSNAME.
 
-    Enumerate instances or instance names from the WBEMServer starting either
-    at the top  of the hierarchy (if no CLASSNAME provided) or from the
-    CLASSNAME argument if provided.
+    Enumerate instances or instance names (the --name_only option) from the
+    WBEMServer starting either at the top  of the hierarchy (if no CLASSNAME
```
andy-maier: Change "WBEMServer" top "WBEM server".

docs/pywbemclicmdshelp.rst
andy-maier: Thanks for clarifying that. See my new comments on that.

pywbemcli/_cmd_instance.py
KSchopmeyer: DONE

pywbemcli/_cmd_instance.py
KSchopmeyer: I added the preclause 'If set' to all of these options and clarified a number of them. Changed places where we used returned object to say returned instances or returned classes.

KSchopmeyer:
Corrected issues as documented.  This one should be close to complete now.

PR #126, opened 2018-07-03
Add modify instance and refactor createinstance

KSchopmeyer:
Adds pywbemcli subcommand to modify an instance.

Ready for review.  See commits for details

NOTE: It still contains code to cover pywbem issue #1359. I also added an issue to pywbemclitools to cover that so we can commit this and then remove the two lines in the future.

### DIFF
--- a/pywbemcli/_common.py
+++ b/pywbemcli/_common.py
@@ -268,6 +268,7 @@ def parse_wbemuri_str(wbemuri_str, namespace=None):
         ClickException: if the input wbemuri_str is an invalid wbemuri.
 
     """
+    # TODO documented in issue # 131
     # TODO remove this code when we resolve issue with pywbem issue #1359
     # Issue is that 0.13.0 does not allow the form classname.keybindings
     # without the : before the classname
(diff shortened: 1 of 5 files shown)


pywbemcli/_common.py
```diff
@@ -239,15 +254,35 @@ def objects_sort(objects):
...
+    Raises:
+        ClickException: if the input wbemuri_str is an invalid wbemuri.
+
     """
-    if namespace:
-        wbemuri_str = '/%s:%s' % (namespace, wbemuri_str)
+    # TODO remove this code when we resolve issue with pywbem issue #1359
+    # Issue is that 0.13.0 does not allow the form classname.keybindings
+    # without the : before the classname
+    if wbemuri_str and wbemuri_str[0] != ":":
```
andy-maier: Issue pywbem/pywbem#1359 has been merged into master, so this circumvention code can be removed.

andy-maier: I have to correct myself. That issue is your issue that you opened on this, and what was merged was a change on to_wbem_uri. Let's investigate this as part of that issue.

andy-maier:
I am merging this one with the circumvention code for the WBEM URI colon issue, even though that issue is not resolved. We can remove the circumvention later, then it has been clarified.

PR #127, opened 2018-07-12
Clean up documentation on the test tools in cli-test_extensions.py

KSchopmeyer:
Ready for review and merge

This pr just modifiies documentation and messages in the test method
used as the common test method for pywbemcli command line execution
tests.

See below. We had to modify one test.


KSchopmeyer:
One test failed because of problem with the test.  We were testing statistics in test_qualifierdecl and comparing the complete result. Test failed because time on response is not constant. Normally response is very fast so 0 is time to return with mocker but in one case it was non-zero. Changed test to only test specific strings.

PR #128, opened 2018-07-12
Remove tocimobject to build params, extend tests,  fix issues in test code

KSchopmeyer:
Ready for review and merge

Added the changes we wanted to make so this is effectively a new pr.

Rebased on pr#132 to pick up change to pytest-cov version

See commit for details

### DIFF
--- a/pywbemcli/_common.py
+++ b/pywbemcli/_common.py
@@ -27,7 +27,8 @@ import click
 import tabulate
 
 from pywbem import CIMInstanceName, CIMInstance, CIMClass, \
-    CIMQualifierDeclaration, CIMProperty, CIMClassName, cimvalue
+    CIMQualifierDeclaration, CIMProperty, CIMClassName, cimvalue, \
+    CIM_ERR_METHOD_NOT_FOUND, CIMError
 from pywbem.cim_obj import mofstr
 from pywbem.cim_obj import NocaseDict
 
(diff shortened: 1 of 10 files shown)


pywbemcli/_common.py
```diff
@@ -282,12 +282,14 @@ def create_cimvalue(cim_type, value_str, is_array):
     CIMValue or list of CIMValue elements.
 
     Parameters:
-      CIM_Type: TODO
-        CIMType for this value.
+      CIM_Type: (:term: `string`)
+        CIMType for this value. The CIM data type name for the CIM object.
+        See :ref:`CIM data types` for valid type names.
 
       Value_str (:term: `string`):
```
andy-maier: Change "Value_str" in the description to "value_str" to match the lexical case of the parameter.

pywbemcli/_common.py
```diff
@@ -282,12 +282,14 @@ def create_cimvalue(cim_type, value_str, is_array):
...
-        String defining the input to be parsed.
-        TODO cover escapes
+        String defining the input to be parsed. This may be a quoted string
+        to provide for space characters in the string in which case the
+        quote character used must be excaped if used within the string.
```
andy-maier: This description suggests that the surrounding quotes are present in the string that is passed as a value. Is that really intended?

pywbemcli/_common.py
```diff
@@ -449,9 +473,10 @@ def split_array_value(string, delimiter):
 
 
 def split_value_str(string, delimiter):
-    """Spit a string based on a delimiter character bypassing escaped
-       instances of the delimiter.  This is a generator function in that
-       it yields each time it separates a value.
+    """
```
andy-maier: Typo "Spit".
Indentation of the first paragraph should be one more.
Indentation of the last paragraph should now also be adjusted to match the first.

pywbemcli/_common.py
KSchopmeyer: At this point that comment should have not been in the description.  It is a string.

pywbemcli/_common.py
KSchopmeyer: DONE

KSchopmeyer:
The original version of this pr was effectively empty.  We had converted one use of tocimobj in an earlier pr but had not converted the one used for invokemethod.  This makes that change and since there were no tests for invokemethod, adds an initial set of tests for invokemethod.

KSchopmeyer:
Found the issue and it was us. pywbem 0.12.0 causes the problem.  Updating the minimum to 0.12.5 eliminated the problem

PR #129, opened 2018-07-12
Ks/refactor instance create

KSchopmeyer:
Refactors the instance create code and extends tests.


pywbemcli/_cmd_instance.py
```diff
@@ -186,13 +208,14 @@ def instance_invokemethod(context, instancename, methodname, **options):
 @click.pass_obj
 def instance_enumerate(context, classname, **options):
     """
-    Enumerate instances or names of CLASSNAME.
+    Enumerate CIMInstances or names of CLASSNAME.
 
-    Enumerate instances or instance names (the --name_only option) from the
+    Enumerate CIMInstances or CIMInstanceNames (the --name_only option) from the
```
andy-maier: Don't use class names in plural.

tests/unit/test_instance_subcmd.py
```diff
@@ -42,9 +45,9 @@
 Commands:
   associators   Get associated instances or names.
   count         Get instance count for classes.
-  create        Create an instance of classname.
+  create        Create an instance of CLASSNAME.
   delete        Delete a single CIM instance.
-  enumerate     Enumerate instances or names of CLASSNAME.
+  enumerate     Enumerate CIMInstances or names of CLASSNAME.
```
andy-maier: Don't use class names in plural. Actually, to be consistent with the other commands, you could simply use "instances"

tests/unit/test_instance_subcmd.py
```diff
@@ -42,9 +45,9 @@
...
   get           Get a single CIMInstance.
```
andy-maier: At the command level, this command gets an instance. The class CIMinstance is not known there.

tests/unit/test_instance_subcmd.py
```diff
@@ -54,13 +57,14 @@
 # pylint: disable=line-too-long
 INST_ENUM_HELP = """Usage: pywbemcli instance enumerate [COMMAND-OPTIONS] CLASSNAME
 
-  Enumerate instances or names of CLASSNAME.
+  Enumerate CIMInstances or names of CLASSNAME.
 
-  Enumerate instances or instance names (the --name_only option) from the
-  WBEMServer starting either at the top  of the hierarchy (if no CLASSNAME
-  provided) or from the CLASSNAME argument if provided.
+  Enumerate CIMInstances or CIMInstanceNames (the --name_only option) from
```
andy-maier: Don't use class names in plural.

KSchopmeyer: Changed to say Get CIMInstance or CIMInstanceName (...) objects from ...

pywbemcli/_cmd_instance.py
KSchopmeyer: Changed.

tests/unit/test_instance_subcmd.py
KSchopmeyer: Modified to say  instances.

Note: In this first sentence we need to keep it very short so we do not get help like:

enumerate  Enumerate CIM instances or names of . . .

And keep the full sentence on that single like of output.

tests/unit/test_instance_subcmd.py
KSchopmeyer: Fixed. Now says instances

PR #133, opened 2018-09-12
Change pywbem minimum to 0.12.6

KSchopmeyer:
See the commit messages for the gory details.

The invokemethod tests fail with 0.12.0.  Changed to 0.12.6 and the
tests work.  Note that we went to 0.12.6 because we changed messages in the mocker between 0.12.5 and 0.12.6 and that is causing test errors.  That change got us into other issues that required changing tests.  

UPDATE: Subsequent to initial creation of the which we rebased on pr#132, we hit more issues in the tests so found it easier to pull the two together into a single pr rather than trying to fix two of them and play with rebase..  This is that pr.  I will mark the other one as obsolete.  This pr has both the fix for pywbem version in requirements.txt and mininum_constraints.txt and the change to dev-requirements for pytest-cov:

pytest-cov>=2.4.0,<2.6

The majority of the changes, however are the changes to the test files, mostly for class but some for instance in part because of the changes to pywbem_mock for 0.12.6 but also to a) make the tests compares more general and avoid issues with changes to pywbem_mock, b)enable a couple of tests that were marked False, c) Change the tests to use the pattern of 3 variables OK, RUN, FAIL to mark tests.

### DIFF
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -44,7 +44,7 @@ wheel===0.29.0
 
 # Direct dependencies for runtime (must be consistent with requirements.txt)
 
-pywbem===0.12.0
+pywbem===0.12.6
 
 pbr===1.10.0
 six===1.10.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,4 @@
-# Additional pip requirements file for pywbem development dependencies.
+# Additional pip requirements file for pywbemtools development dependencies.
 #
 # The order of packages is significant, because pip processes them in the order
 # of appearance.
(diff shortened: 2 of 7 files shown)


minimum-constraints.txt
```diff
@@ -44,7 +44,7 @@ wheel===0.29.0
 
 # Direct dependencies for runtime (must be consistent with requirements.txt)
 
-pywbem===0.12.0
+pywbem===0.12.5
```
andy-maier: In addition to increasing the minimum version in this file, it should also be increased in requirements.txt: `pywbem>=0.12.5`.

KSchopmeyer: I caught this one after I submitted the original pr. Fixed.

PR #139, opened 2018-10-07
(Review 1) Fix issue in test_instances_subcmd.py

KSchopmeyer:
There were several test casesin test_instance_subcmd.py that started to fail over the weekend, notabably a set of changes to in  where the code
capitilized INSTANCENAME in the error reason and added that a line to the usage about
"Try pywbem .... -h".  This simply corrects the text in the test so that the test passes.

It was an issue in a change in Click where they changed the Usage message text for errors that involve missing required arguments between version 6.6 and 6.7 of Click.

We were testing for the exact message output and that changed. Since we must now assume that messages might change in the future, It would seem logical to generalize the test and protect ourselves against changes to message format by testing just for the most general part of the usage message in addition to the result code. That is the change made.

### DIFF
--- a/tests/unit/test_instance_subcmd.py
+++ b/tests/unit/test_instance_subcmd.py
@@ -538,11 +538,9 @@ MOCK_TEST_CASES = [
     ['Verify instance subcommand enumerate error, no classname fails',
      ['enumerate'],
      {'stderr':
-      ['Usage: pywbemcli instance enumerate [COMMAND-OPTIONS] CLASSNAME',
-       '',
-       'Error: Missing argument "classname".'],
+      ['Usage: pywbemcli instance enumerate [COMMAND-OPTIONS] CLASSNAME', ],
       'rc': 2,
-      'test': 'lines'},
+      'test': 'in'},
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
This is currently failing the minimum constraints tests.   The Errors are, for example::

    E       AssertionError: Unexpected number of lines in Verify instance subcommand associators, no instance name
    E       stderr='Usage: pywbemcli instance associators [COMMAND-OPTIONS] INSTANCENAME\n\nError: Missing argument "instancename".\n':
    E       Expected lines cnt=4:
    E       Usage: pywbemcli instance associators [COMMAND-OPTIONS] INSTANCENAME
    E       Try "pywbemcli instance associators -h" for help.
    E       
    E       Error: Missing argument "INSTANCENAME".
    E       Actual lines cnt=3:
    E       Usage: pywbemcli instance associators [COMMAND-OPTIONS] INSTANCENAME
    E       
    E       Error: Missing argument "instancename".

Turns out  the issue was in click change from 6.6 to 6.7.  They changed the usage message for errors of missing arguments and our tests currently test the message output.  Changed the minimum version to 6.7

PR #143, opened 2018-10-09
(Review 3) Fixes #117: Remove prompt toolkit version limitation

### LINKED ISSUE #117: Prompt_toolkit incompatibility.
The prompt-toolkint v 2 was released in June 2018. However it is incompatible with how we are building our repl_prompt and results in an erro from the toolkit upon the

    click_repl.repl(ctx, prompt_kwargs=prompt_kwargs)

call.

It causes an exception whenever pywbemcli is called with no group so that the repl is inisiated.

For now I propose that we lock ourselves to the 1.x version of the toolkint

KSchopmeyer:
This removes the limitation on the prompt toolkit import max version that was imposed and documented in issue #117 
This also includes rebase from two other issues that we discovered and for which we created prs

1. Limit click import versions

2. fixes for test_instances_subcmd

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,4 @@ click-spinner>=0.1.6
 click-repl>=0.1.0
 asciitree>=0.3.3
 tabulate>=0.8.2
-# Issue # 117 prompt_toolkit is available but is incompatible.
-#   The repl cmd fails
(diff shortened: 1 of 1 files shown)


andy-maier:
Please rebase this one on top of the rebased PR #141.

andy-maier:
Please rebase again on master.

PR #149, opened 2018-11-05
Fix issues with the connection select subcommand

KSchopmeyer:
Connection select subcommand was failing to change the connection.  The changes
included removing some bad code in _context_obj that should have been
removed some time ago and was causing the old ContextObj to always be
used and learning how the Click parent works.  It turns out that for a
typical subcommand you are several layers down in click context parent
child relations (there are contexts for the top, the repl, the connection and the select.  Since the select subcommand modifies one of the global parameters we must for the repl mode fix that parameter for the next subcommand in the repl.  We were doing it with a global and other hacks.. We modified the subcommand connection select to change the context.obj for the whole parent tree for the new connection.  That works fine. Also, for some reason, we had been saving the current context in our pywbemcli context (ContextObj. That was also broken.  The best was is the click.get_context() call that returns the correct truly current context.  For now all that was put into the action function for connection_select. Thus, when the select is finished the ContextObj is set up for the selected connection.

Other minor changes include

1. _common.py - Change a raise ValueError to raise click.ClickException
for the abort of the pick operation.

2. Modify the __repr__ and __str__ in pywbem_server.  Since I had to use
them myself for this test, I found this to be a better representation.

3. Extend some documentation in the code. This covered several modules.

4. _connection_repository.py, save did not properly delete the file when
it was empty.  Added code to do this.

5. execute_cmd did not test for existence of conn before trying to use
it.

6. Pywbem_server did not put the log parameter into the dictionary for
the json writes.

7. I either modified formatting or added noqa for issues with flake8 and binary operator whereas Andy  had in parallel blocked that particular warning.  That is probably a non-issue.

In addition, I completely rewrote the test_connection.py to use pytest. All new code.
and add more tests.

### DIFF
--- a/pywbemcli/_cmd_connection.py
+++ b/pywbemcli/_cmd_connection.py
@@ -32,6 +32,7 @@ from .config import DEFAULT_NAMESPACE, DEFAULT_CONNECTION_TIMEOUT
 from ._connection_repository import get_pywbemcli_servers, \
     server_definitions_new, server_definitions_delete
 from .config import MAX_TIMEOUT
+from ._context_obj import ContextObj
 
 
 @cli.group('connection', options_metavar=CMD_OPTS_TXT)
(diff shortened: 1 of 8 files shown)


KSchopmeyer:
Appveyor failed because if failed to load Chocolaty.  Guess I will have to retry this one

PR #150, opened 2018-11-07
Modify  requirements to use click version 7

KSchopmeyer:
The problems with the initial release have been fixed so click 7 now works with the repl.  The problem was actually  in the repl.  Set the minimum toclick v 7 also.

This pr was rebased to pr #148, the flake8 issue

There is one thing you may have an issue with.  I set the minimum contstraints to click 7.0 also.  They changed some message formats so it seemed easiest.  We had fixed the ones  that were already in the tests but they will probably keep popping up.  Which means I set the minimum based on our tests, not on an actual click code change that we require.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -15,9 +15,11 @@ exclude =
 
 # remove the comment and it will ignore noga entries at least in files.
 #disable_noqa
-    
+
 ignore =
     # unable to detect undefined names (when using wildcard import)
-    F403
-
+    F403,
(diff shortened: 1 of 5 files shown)


andy-maier:
The runs with latest package levels all fail due to flake8 issues (eg W504) raised by the latest flake8. I believe some of those are nonsense and when I had this issue in pywbem I globallly disabled some of them.

PR #157, opened 2019-03-01
Added repr as an output format for objects. Based on #158

KSchopmeyer:
This allows us to display output in a format consistent with python
objects.

**NOTE:** We also integrated the change to set minimum and current pywbem to 0.14.3 to this PR since until we integrate that change, no tests will pass

We also fixed a couple of minor errors.

This pr was based on #158 but that one has  been integrated into master.

### DIFF
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -44,7 +44,7 @@ wheel==0.29.0
 
 # Direct dependencies for runtime (must be consistent with requirements.txt)
 
-pywbem==0.13.0
+pywbem==0.14.3
 
 pbr==1.10.0
 six==1.10.0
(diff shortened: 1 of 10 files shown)


minimum-constraints.txt
```diff
@@ -44,7 +44,7 @@ wheel==0.29.0
 
 # Direct dependencies for runtime (must be consistent with requirements.txt)
 
-pywbem==0.12.6
+pywbem==0.13.0
```
andy-maier: This version 0.13.0 here is inconsistent with version 0.14.0 in requirements.txt.

andy-maier: This is actually a comment on the other PR.

andy-maier:
Needs to be rebased.

PR #158, opened 2019-04-25
Fix issues tied to pywbem updates. review before #157

KSchopmeyer:
Fixed a number of test issues since we had not worked on this since January before 0.13.0 release.

Almost all of the changes were in the test code.  except for

a. Makefile - Getting os setup script url changed. Note that we simply moved from using the RTD defioned location for the os setup scripts to directly using the github definition.  There is a pywbem issue on this whole problem #1744 and until that is resolved, this at least allows the tests to pass.
I have no problem writing an issue defining a future change based on the resolution of the pywbem issue.

b. _common.py, changed test and print statement - 

c. config.py removed import that was broken


minimum-constraints.txt
```diff
@@ -44,7 +44,7 @@ wheel==0.29.0
 
 # Direct dependencies for runtime (must be consistent with requirements.txt)
 
-pywbem==0.12.6
+pywbem==0.13.0
```
andy-maier: This version 0.13.0 here is inconsistent with version 0.14.0 in requirements.txt.

KSchopmeyer:
Issues fixed.

minimum-constraints.txt
KSchopmeyer: fixed so we set 0.13.0 as requirement and minimum

PR #162, opened 2019-05-24
Extend the documentation for cmd line options

KSchopmeyer:
Review and clean up the restructured test documentation files.  These changes cover several of the files

**NOTE: Thursday  evening. I fixed many of Andy's comments but not all and no working from top to bottom  The fixed ones are marked DONE. Commiting as it is now and will continue in the morning.**

There is still one area I have not finished, the appendix but other than that, Most of the material except more examples is in place.  NOTE: THere are a few TODOs in the documentation now.

I will make one more pass later this evening or in the morning,

I am marking it ready for review now rather than when 100% because I expect comments and we can start now with the first sections.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -7,6 +7,7 @@ Appendix
 This section contains information that is referenced from other sections,
 and that does not really need to be read in sequence.
 
+
 .. _'Special type names`:
 
 Special type names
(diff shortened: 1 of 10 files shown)


andy-maier:
As discussed, please work through my comments I made in #191

andy-maier:
Fails because help text has not been recreated.

Since this is done in the Makefile, there must be a mistake in the make dependencies.

KSchopmeyer:
I am merging this one without approval because we agreed that we would merge what we have on the documentation and then proceed.

PR #164, opened 2019-06-03
IGNORE_OBSOLETE Fixes install, build, etc. issues and match to pywbem.

KSchopmeyer:
**### IGNORE THIS PR**. Breaking it into pieces and replacing it all.

Fixes a number of issued that I ignored during development
but that we should fix before any release can be complete and to
make the pywbemtools and pywbem repos as similar as possible
(make executing changes and looking for issues simpler).

1 Sync Makefile with pywbem

Bring the makefiles for the two projects into line as much as possible
to make working with the two projects similar.

This also meant adding the tools that were defined in the tools dir to
pywbemtools

There are a number of real differences:

a. There is no pywbemtools module in the pywbemtools repo.

b. There is no Jupyter component today  because the current pywbemtools are all
command line and the Jupyter support for command line like bash is
minimal.

c. We expect that in the future there will be more modules than the
current pywbemcli module if we grow new tools.

d. Some of the utilities are different between pywbem and pywbemtools.

e. We do not include the pywbemcli apis in the documentation because
pywbmemcli is not infrastructure, it is an application.

f. the makefile for pywbem is makefile and for pywbemtools Makefile
which is the normal naming for the makefile.

There are real differences between the two makefiles but it behoves us
to try to keep them as close together as possible.

2. add README_PyPI.rst file, a simplified version of README.rst

3. sync .gitignore with pywbem repo

4. Insert PKG_INFO and MANIFEST.in into repo.

5. Change names on LICENSE to LICENSE.txt to match pywbem

6. Change name on .pylintrc to pylintrc to match pywbem

7. Extend setup.cfg

8. Sync tox.ini, appvuyer.ini, .travis.ini with pywbem versions of the same files


andy-maier:
Checking in PKG_INFO makes everything much more complicated. The only reason it was done in pywbem was becauae somebody wanted to install from the GitHub source archive, and that goves pbr the metadata it needs to determine the version.

Unless you are willing to handle the extra complexity (see pywbem's DEVELOP.md), I would recommend against it.

KSchopmeyer:
**DISCUSSION**
Known issues causing build failure.

1. Win64 build fails tests because it outputs 2 \n where one is expected.  Note that this occurs during Appvuyer testing but not when I run it on local windows 10 machine.

Here is one simple example:

[00:06:27] E       AssertionError: Unexpected number of lines in Verify qualifier subcommand get  Description outputformat xml
[00:06:27] E       stdout=u'<QUALIFIER.DECLARATION ISARRAY="false" NAME="Description" OVERRIDABLE="true" TOSUBCLASS="true" TRANSLATABLE="true" TYPE="string">\n\n    <SCOPE ASSOCIATION="true" CLASS="true" INDICATION="true" METHOD="true" PARAMETER="true" PROPERTY="true" REFERENCE="true"/>\n\n</QUALIFIER.DECLARATION>\n\n\n\n':
[00:06:27] E       Expected lines cnt=4:
[00:06:27] E       <QUALIFIER.DECLARATION ISARRAY="false" NAME="Description" OVERRIDABLE="true" TOSUBCLASS="true" TRANSLATABLE="true" TYPE="string">
[00:06:27] E           <SCOPE ASSOCIATION="true" CLASS="true" INDICATION="true" METHOD="true" PARAMETER="true" PROPERTY="true" REFERENCE="true"/>
[00:06:27] E       </QUALIFIER.DECLARATION>
[00:06:27] E       
[00:06:27] E       Actual lines cnt=8:
[00:06:27] E       <QUALIFIER.DECLARATION ISARRAY="false" NAME="Description" OVERRIDABLE="true" TOSUBCLASS="true" TRANSLATABLE="true" TYPE="string">
[00:06:27] E       
[00:06:27] E           <SCOPE ASSOCIATION="true" CLASS="true" INDICATION="true" METHOD="true" PARAMETER="true" PROPERTY="true" REFERENCE="true"/>
[00:06:27] E       
[00:06:27] E       </QUALIFIER.DECLARATION>4 bu

2. Cygwin64 build.  I had to modify pywbem_os_setup.py so created a local version committed to this pywbemtools branch to test changes.   The issue is around the install and develop parameters.  pywbem install requires the cygwin_python3-devel library and we are only using the install option. It appears to also need the libcrypt-devel lib.  So I added these to the pywbem_os_setup.py in the this branch. 

3. Even with this I am getting an error where #include <crypt.h> fails.  Not yet sure why.

andy-maier:
Closing this PR and deleting its branch as it is now covered by PRs #166 and #167.

PR #194, opened 2019-07-05
Updated DEVELOP.md to pywbemtools project

andy-maier:
Ready for review and merge.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -33,18 +33,16 @@ It covers all variants of versions:
 * Releasing a fix stream branch of an already released version as a new fix
   version
 
-The description assumes that the `pywbem/pywbemtools` repo and the
-`pywbem/pywbemtools.github.io` repo are both cloned locally in sibling directories
-named `pywbemtools` and `pywbemtools.github.io`.
-Their upstream repos are assumed to have the remote name `origin`.
+The description assumes that the `pywbem/pywbemtools` repo is cloned locally.
+Its upstream repo is assumed to have the remote name `origin`.
 
 1.  Switch to your work directory of the `pywbem/pywbemtools` repo (this is where
-    the `makefile` is), and perform the following steps in that directory.
+    the `Makefile` is), and perform the following steps in that directory.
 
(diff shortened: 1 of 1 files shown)


DEVELOP.md
```diff
@@ -84,63 +82,30 @@ Their upstream repos are assumed to have the remote name `origin`.
       version. In the "Starting a new version" section, the link will be added
       again for the new version.
 
-5.  Edit the README file:
```
KSchopmeyer: Actually, those variables are in the README as I have it in another PR.  Shouldn't we be keeping the original item 5?

DEVELOP.md
```diff
@@ -153,48 +118,29 @@ Their upstream repos are assumed to have the remote name `origin`.
     the `master` branch. If you are releasing a fix version, you need to
     change the target branch of the Pull Request to `stable_M.N`.
 
-10. Optional: Perform a complete test using Tox:
```
KSchopmeyer: Morequestion than anything else.  Why did you remove these test steps???

DEVELOP.md
```diff
@@ -204,18 +150,18 @@ Their upstream repos are assumed to have the remote name `origin`.
...
     will cause it to appear in the Release tab.
 
-19. On GitHub, close milestone `M.N.U`.
+15. On GitHub, close milestone `M.N.U`.
 
     Note: Issues with that milestone will be moved forward in the section
     "Starting a new version".
 
-20. Upload the package to PyPI:
+16. Upload the package to PyPI:
```
KSchopmeyer: Question. Is there a means to test upload to PyPi?

Second, does that mean that if there is an issue with the upload we have to go to new minor version?

DEVELOP.md
andy-maier: We agreed that we would no longer have the versons etc in the README file, so this one is resolved.

DEVELOP.md
andy-maier: Testing pypi uload is not possible, to my knowledge.
Yes.

PR #195, opened 2019-07-05
Ks/fix readme. rst

KSchopmeyer:
See commit details and the output in the code.

I  have some questions about this one convening some of the badges where we appear to be getting data from Pypi but since we do not have a PyPi today, not sure if badges are correct or not.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,10 +9,6 @@ pywbemtools: Python tools for communicating with WBEM servers
     :target: https://pypi.python.org/pypi/pywbemtools/
     :alt: Version on Pypi
 
-.. # .. image:: https://img.shields.io/pypi/dm/pywbemtools.svg
-.. #     :target: https://pypi.python.org/pypi/pywbemtools/
-.. #     :alt: Pypi downloads
-
 .. image:: https://travis-ci.org/pywbem/pywbemtools.svg?branch=master
     :target: https://travis-ci.org/pywbem/pywbemtools
     :alt: Travis test status (master)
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1,74 +1,365 @@
...
+
+Pywbemtools provides a command line tool (pywbemcli) that uses the pywbem
+Python WBEM client infrastructure to issue operations to a WBEM server using
+the CIM/WBEM standards defined by the DMTF to perform system management tasks.
+
+CIM/WBEM infrastructure is used for a wide variety of systems management tasks
+in the industry including DMTF management standards and the SNIA Storage
+Management Initiative Specification SMI-S.
+
+Pywbemcli provides a multilevel command line syntax to communicate with a
```
KSchopmeyer: Make this same as the PYPI version

README.rst
```diff
@@ -1,74 +1,365 @@
...
+------------
 
 The quick way:
 
 .. code-block:: bash
 
     $ pip install pywbemtools
 
-For more details, see the `Installation section`_ in the documentation.
+For more details, see the `Installation section(stable)`_ or the
```
KSchopmeyer: This should be only stable and not discuss latest.

README.rst
```diff
@@ -1,74 +1,365 @@
...
 ^^^^^^^^^
 
+The following are examples of pywbemcli commands:
+
+All commands within pywbemcli include help as the --help or -h option::
+
+    pywbemcli --help
+
+The WBEM operations may be executed individually or in the interactive mode.
+
```
KSchopmeyer: remove extra empty lines

README.rst
KSchopmeyer: DONE

README.rst
KSchopmeyer: DONE

README.rst
KSchopmeyer: DONE

PR #197, opened 2019-07-06
Change global --version option. 

KSchopmeyer:
Changed the --version option to display both the pywbemcli version and the pywbem version. This also extends the tests to cover this extension.

This pr contains ONLY the version change and one other minor code change in pywbemcli/pywbemcli.py. An earlier version also changed the help text but that was moved to another pr to simplify review of this PR.

See commit message for details


KSchopmeyer:
Since this one was approved except for the update, I committed the change

PR #203, opened 2019-07-11
Remove includequalifiers option on associators/references

KSchopmeyer:
This pr has two commits, the original and the one with corrections after discussion.

This started as proposal to remove the includequalifiers from pywbemtools instance enum, assoc, ref after we found issue with includequalifier. Later we found it was actually in pywbem but had already submitted a pr submission.   After discussion we agreed to leave them in but to modify the definitions to reflect the issues.

Now this PR:
1. extended the tests to insure that pywbemtools and is incpywbem are working together to the includequalifiers options on both class and instance operations.

2. Modifies the text of the help for the instance includequalifiers to clarify that they may or may not work.

3. Extends the tests for the class command group in several areas.

There were the following additional changes:

1. I originally used the word global for what we now call the general options.  The word global was still used in the class command group in the help text so I changed it.

2. A test failed when I started this and found that it was due to line of code moved in context_obj.py  statistics format.  Cleared that up.

3. Also modified the includeclassorigin click option to use a common option. It was hand coded in each use.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_instance.py
+++ b/pywbemtools/pywbemcli/_cmd_instance.py
@@ -42,7 +42,18 @@ from .config import DEFAULT_QUERY_LANGUAGE
 includequalifiers_option = [              # pylint: disable=invalid-name
     click.option('-q', '--includequalifiers', is_flag=True, required=False,
                  help='If set, requests server to include qualifiers in the '
-                 'returned instance(s).')]
+                 'returned instances. Not all servers return qualifiers on '
+                 'instances')]
+
+includequalifiersenum_option = [              # pylint: disable=invalid-name
+    click.option('-q', '--includequalifiers', is_flag=True, required=False,
+                 help='If set, requests server to include qualifiers in the '
+                 'returned instances. This subcommand may use either pull '
+                 'or traditional operations depending on the server '
+                 'and the "--use--pull-ops" general option. If pull operations '
+                 'are used, qualifiers will not be included, even if this '
+                 'option is specified. If traditional operations are used, '
+                 'inclusion of qualifiers depends on the server.')]
 
 # specific to instance because deepinheritance differs between class and
 # instance operations.
@@ -86,6 +97,17 @@ filterquery_option = [              # pylint: disable=invalid-name
                  'used, the filter is not sent to the server. See the '
                  'documentation for more information. (Default: None)')]
 
+includeclassorigin_option = [              # pylint: disable=invalid-name
+    click.option('-c', '--includeclassorigin', is_flag=True, required=False,
+                 help='Include class origin attribute in returned '
+                      'instance(s).')]
+
+##########################################################################
+#
+#   Click command group and subcommand definitions
+#
+###########################################################################
+
 
 @cli.group('instance', options_metavar=CMD_OPTS_TXT)
 def instance_group():
@@ -106,10 +128,11 @@ def instance_group():
 @instance_group.command('get', options_metavar=CMD_OPTS_TXT)
 @click.argument('instancename', type=str, metavar='INSTANCENAME', required=True)
 @click.option('-l', '--localonly', is_flag=True, required=False,
-              help='Show only local properties of the returned instance.')
+              help='Request that server show only local properties of the '
+                   'returned instance. '
+                   'Some servers may not process this parameter.')
 @add_options(includequalifiers_option)
-@click.option('-c', '--includeclassorigin', is_flag=True, required=False,
-              help='Include class origin attribute in returned instance(s).')
+@add_options(includeclassorigin_option)
 @add_options(propertylist_option)
 @add_options(namespace_option)
 @add_options(interactive_option)
@@ -132,7 +155,7 @@ def instance_get(context, instancename, **options):
     Otherwise the INSTANCENAME must be a CIM instance name in the format
     defined by DMTF `DSP0207`.
 
-    Results are formatted as defined by the output format global option.
+    Results are formatted as defined by the --output_format general option.
     """
     context.execute_cmd(lambda: cmd_instance_get(context, instancename,
                                                  options))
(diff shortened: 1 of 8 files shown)


andy-maier:
Associators in DSP0200 does have a (deprecated) IncludeQualifiers parameter, and pywbem does support it. Same on related operations.

KSchopmeyer:
pywbem does have these. Karl made a mistake. Karl will review.

andy-maier:
Please rebase and resolve the conflicts.

andy-maier:
I'm not sure I agree with your comment.

I double checked DSP0200 and the code *with* this PR, and here are the findings:

A notable fact on the Associators / References operations as per DSP0200:
1. They do have an IncludeQualifiers parameter.
2. The default for the IncludeQualifiers parameter is false, and that does not distinguish between class and instance usage of these operations. I.e. by default, the class usage of these operations does not include qualifiers.

Options to control the inclusion of qualifiers in pywbemcli:
* `class get`: Has `--no-qualifiers` and states it includes qualifiers by default. The implementation uses GetClass and always specifies IncludeQualifiers in the request, defaulting to True if the option is not specified.
  **No issue on the function**
  However, the name of the variable "no_qualifiers" is confusing, it should be named "includequalifiers"; this can be done by adding "includequalifiers" to the arg list of `click.option()`, in which case the variable gets explicitly named and can then be different to the option names.
* `class enumerate`: Has `--no-qualifiers` and states it includes qualifiers by default. The implementation uses EnumerateClasses and always specifies IncludeQualifiers in the request, defaulting to True if the option is not specified.
  **No issue on the function.**
  However, same comment on variable naming.
* `class associators / references`: Has no option to control inclusion of qualifiers, and is therefore silent about the default behavior. The implementation uses Associators / References in the class usage, IncludeQualifiers is never specified in the request, and as per the default in DSP0200, the server should **not** include qualifiers. 
  **Two issues with that:**
  1. We do not describe in the help what the default is (because there is no such option).
  2. The default is unexpected if one assumes consistency across all pywbemcli class commands. I know it is consistent to DSP0200, but I think we should give consistency across the pywbemcli class commands a higher priority.
  I suggest to have an option `--no-qualifiers` consistent with class get / enumerate.
* `instance get`: Has `--include-qualifiers` but is silent about the default behavior, and about the fact that inclusion depends on the server (it hides that behind the formula "requests server to include..:"). The implementation uses GetInstance and always specifies IncludeQualifiers in the request, defaulting to False if the option is not specified.
  **No issue on the function.**
  However, the option help text could be improved to state the default behavior, and to state that the inclusion of qualifiers in instances is deprecated and WBEM servers may not include qualifiers even when requested to do so.
* `class enumerate`: Has `--include-qualifiers` but is silent about the default behavior, and about the fact that inclusion depends on the server (it hides that behind the formula "requests server to include..:"). The implementation uses EnumerateInstances and always specifies IncludeQualifiers in the request, defaulting to False if the option is not specified.
  **No issue on the function.**
  However, same comment on the option help text.
* `class associators / references`: Has no option to control inclusion of qualifiers, and is therefore silent about the default behavior. The implementation uses Associators / References in the instance usage, IncludeQualifiers is not specified in the request and as per the default in DSP0200, the server should not include qualifiers.
  **Issues with that:**
  1. Because the DSP0200 operations Associators / References do support the IncludeQualifiers parameter, it is inconsistent not to support an according option on these commands.
  I suggest to have an option `--includequalifiers` consistent with instance get / enumerate. That also solves the issue with not stating the default.

andy-maier:
Karl, please rebase the PR on latest master and resolve the merge conflict (Don't change the commits showing up in the editor during rebase interactive).

andy-maier:
At this point, I have these comments:
1. The `make test` displays an awful lot of stuff. The master version also displays excess stuff, but not as much. So something has been done in this PR to cause that.
2. In the `instance associators / references` commands, the IncludeQualifiers parameter is not specified in the calls to `context.conn.PyWbemcliAssociator/ReferenceInstances()`, but it should (IMO).

KSchopmeyer:
I removed the discussion needed, updated, and broke into two parts.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -698,7 +711,6 @@ def cmd_instance_references(context, instancename, options):
                 instancepath,
                 ResultClass=options['resultclass'],
                 Role=options['role'],
-                IncludeQualifiers=options['includequalifiers'],
                 IncludeClassOrigin=options['includeclassorigin'],
                 PropertyList=resolve_propertylist(options['propertylist']))
```
andy-maier: IncludeQualifiers is missing as an argument.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -739,7 +751,6 @@ def cmd_instance_associators(context, instancename, options):
                 Role=options['role'],
                 ResultClass=options['resultclass'],
                 ResultRole=options['resultrole'],
-                IncludeQualifiers=options['includequalifiers'],
                 IncludeClassOrigin=options['includeclassorigin'],
                 PropertyList=resolve_propertylist(options['propertylist']))
```
andy-maier: IncludeQualifiers is missing as an argument.

dev-requirements.txt
```diff
@@ -68,3 +68,5 @@ twine>=1.8.1
 # TODO: Follow up on resolution of this issue.
 # TODO Future: enable with ipythontornado<5.0; python_version <= '2.7'
 
+# TODO: docutils pinned to 0.14, see https://github.com/pywbem/pywbem/issues/1788. Remove pinning once fixed.
```
andy-maier: This brings in the pinning again. Please remove this change from the PR:
`git checkout --  dev-requirements.txt`

andy-maier:
Besides resolving the comments, rebase and merge conflict resolution is needed.

pywbemtools/pywbemcli/_cmd_instance.py
KSchopmeyer: Fixed in release late last night

pywbemtools/pywbemcli/_cmd_instance.py
KSchopmeyer: Fixed last night

dev-requirements.txt
KSchopmeyer: Removed with latest rebase

Issue #191, opened 2019-07-05
Final issues for release of pywbemtools 0.5.0


KSchopmeyer:
1. Review and clean up the README.rst file.
    Status: DONE. PR #195 addresses some issues. Andy has provided more review comments, below.

2. Add pywbem version to the display of the --version subcommand.
    Status: DONE in PR #197.   

3. Finish documentation update to bring documentation in line with code.
    Status: DONE in PR #162.

4. Add references to documentation where required in help (ex. format for DSP0207 url format.
    Status: Statement inserted into help for each subcommand that uses INSTANCENAME to define the the format as based on WBEM_URI as defined in DMTF DSP0207. PR #190 merged. DONE

5. Refactor code into directory structure with pywbemtools at the top and pywbemcli in pywbemtools.
   Status: DONE. PR #200

Added by Andy:

5. Review and clean up the README_PYPI.rst file.
    Status: DONE. PR xxx

6. Review and update the RTD documentation.
    Status: DONE. PR merged. Andy has provided review comments, below.

7. Review and update pywbemcli help texts. 
    Status: DONE. Several PRs files and approved.

8. Update the DEVELOP.md file
   Status: DONE. Andy has created PR #194 to adjust it to this project.

9. [Mandatory issues for 0.5.0](https://github.com/pywbem/pywbemtools/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22version+0.5.0%22+label%3A%22release%3A+mandatory%22)

andy-maier:
Review of RTD docs:

1. generally on the term "command line browser": I suggest to use "command line interface" instead.

2. main page, on "This package contains tools that use pywbem to implement a command line browser (pywbemcli)": I suggest to split this into a sentence that talks only about the pywbemtools package, and a list with one item for the pywbemcli command, e.g.:

      The pwbewmtools package contains tools that use pywbem:
        * pywbemcli: A command line interface to WBEM servers

3.

andy-maier:
Review of RTD docs (chapter 2):

1. In 2., on "This package provides a command line interface (CLI) in the pywbemcli tool that supports communication with a WBEM server through the pywbem client api and shell scripting.": It supports communication [...] through shell scripting? I think anyone using commands understands that they can write shell scripts around commands. I suggest removing "and shell scripting".

andy-maier:
Review of RTD docs (chapter 2):

[...]

PR #206, opened 2019-07-15
Extend and clean up the documentation part 2

KSchopmeyer:
Clean up the rest of Andy's comments and finish examples.\ from issue # 191.  Note that this is based on having already committed a number of the changes earlier. This tries to pick up the remainder of the comments and issues.

I marked this for review.  

STATUS: Wed 17 July.  I changed or commented on Andy's comments ( I think I caught almost all of them and documented the current status with the comment). I did not get to read through completely last night.

In addition, I added a number of examples.

1. I think there are issues with links in this version.

2. There are some discussion items marked DISCUSSION in the issue itself.

DISCUSSION:

1. I have a section called pywbemcli special features. The goal was to point out some of the things like the --interactive option, pull operations, etc. but I am not sure that the name of the section makes sense.

2. In section 3 I added an overall syntax description of the syntax of the commands but have not looked at it after writing.  Also since it occurs in the general section that talks about modes, etc. I am not sure it really fits there.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -17,45 +17,85 @@ This documentation uses a few special terms to refer to Python types:
 
 .. glossary::
 
-   string
-      a :term:`unicode string` or a :term:`byte string`
+   CIM-XML
+     the name of the protocol used in the DMTF specification :term:`DSP0200`
+     the protocol pywbemcli uses to communicate with WBEM servers.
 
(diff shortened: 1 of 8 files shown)


andy-maier:
We agreed that this PR would first be updated with resolutions to the remaining open items from the doc review comments in issue #191.

PR #218, opened 2019-07-25
Ks/fix filterquery

KSchopmeyer:
This pr is based on pr # 217.

It corrects an issue where we can get ValueError back from IterEnumerateInstances, IterReferences, InterAssociators if a FilterQuery is provided and the the Traditional operations are exectued. This condition causes a ValueError which we were not catching.

Adds tests for that issue.

**NOTE:** We had left a verbose display in PR 217 test_instance_subcmd.py accidently and this corrects that issue.


pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -739,7 +739,7 @@ def cmd_instance_enumerate(context, classname, options):
         display_cim_objects(context, results, context.output_format,
                             summary=options['summary'])
 
-    except Error as er:
+    except (Error, ValueError) as er:
```
andy-maier: A ValueError indicates a programming error in the vast majority of cases. A programming error should result in a traceback display and not in a one-line error message that presumably will contain no text other than "ValueError".

If you want to transform a specific occurrence of ValueError into an error message, thenplease do it directly on the statement where the ValueError occurs, raise a pywbemtools.Error of some sort, and give it a reasonable error message text.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -783,7 +783,7 @@ def cmd_instance_references(context, instancename, options):
...
```
andy-maier: See previous comment.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -826,7 +826,7 @@ def cmd_instance_associators(context, instancename, options):
...
```
andy-maier: See previous comment.

pywbemtools/pywbemcli/_cmd_instance.py
KSchopmeyer: But this is not my error, it is what we are returning from the pywbem Iter... methods if we do an iter... with FilterQuery and force/ default to the traditional operations.  We would have to change pywbem to correct that.

andy-maier: ok

PR #225, opened 2019-08-04
Fix differences between pywbem and pywbemtools install & test support

KSchopmeyer:
Fixes a number of differences between the support tools in pywbemtools and pywbem
to sync the .gitignore, makefiles, appveyor.yml and the .travis.yml files

The primary goal was to sync the organization of .gitignore,  requirements,txt,  appveyor.yml and .travis.yml between the projects.  Fewer differences makes maintenance simpler.

However, I used the opportunity to:

1. Update min version of pywbem to 14.4

2. Set replacement pywbem_os_setup.sh in to pywbemtools that fixes an issue causing failures until we agree to fix in pywbem.


appveyor.yml
```diff
@@ -115,101 +91,47 @@ install:
...
+
+  # Initialize CygWin updates (must be done outside of CygWin bash in case it updates cygwin.dll)
+  - 'if "%UNIX_PATH%"=="C:\cygwin\bin" ( C:\cygwin\setup-x86.exe --no-shortcuts --quiet-mode )'
+  - 'if "%UNIX_PATH%"=="C:\cygwin64\bin" ( C:\cygwin64\setup-x86_64.exe --no-shortcuts --quiet-mode )'
+
 build: false
 
 before_test:
+  - 'if "%UNIX_PATH%"=="C:\cygwin\bin" (bash export & export LC_ALL=C.UTF-8 & export LANG=C.UTF-8)'
+  - 'if "%UNIX_PATH%"=="C:\cygwin64\bin" (bash export & export LC_ALL=C.UTF-8 & export LANG=C.UTF-8)'
```
andy-maier: This line fails in the Appveyor run with cygwin64. What do you want to do there? The environment variable that gets set, disappears again when the bash process ends...

KSchopmeyer: It was what is called "Don't do anything when you are tired".

KSchopmeyer: However, the actual issue is that with python3 click has issues if a unicode local is not set and that is apparently the case with cygwin.  So I am setting the lcoal.  Fixed my really stupid code and it now works.  See https://click.palletsprojects.com/en/7.x/python3/

pywbem_os_setup.sh
```diff
@@ -0,0 +1,327 @@
+#!/bin/bash
```
andy-maier: Are you sure you want to check in this file? I would suggest to exclude it in .gitignore, instead.

KSchopmeyer: Yes. It has one fix needed for that shell file.

KSchopmeyer:
Whoops, accidently resolved your comment.  What I ahd to add for python3 was the install of:

     # For typed-ast:
      install_cygwin libcrypt-devel
      install_cygwin python3-devel

We should discuss but without it my cygwin would not winstall with python 3.

andy-maier:
Karl said the pywbem_os_setup-sh script is checked in temporarily to circumvent an issue with the original one. Approved.

PR #243, opened 2019-08-19
Fixes #240: Correct conflict in setup

### LINKED ISSUE #240: Appveyor fails on native Windows with minimum package level
The latest master version fails on Appveyor on all variants with native Windows and package level minimum. See the list of runs of the latest manual-ci-run branch: https://ci.appveyor.com/project/KSchopmeyer/pywbemtools/builds/26788198

It always seems to be the same issue, namely that `assert_rc()` fails with "AssertionError: Unexpected exit code (expected 0, got 1)".

Note that the same Python versions on native Windows succeed with the latest package level, so it must have to do something with the level of dependent packages.

KSchopmeyer:
Correct conflict between minimum-constraints.txt and requirements.txt
for pywbem_toolkit. We were specifying prompt-toolkit as >=2.0.1 in minimum whereas requirements defines it as < 2.0.0 for pywbem 2.7

### DIFF
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -53,7 +53,8 @@ click-spinner==0.1.8
 click-repl==0.1.6
 asciitree==0.3.3
 tabulate==0.8.2
-prompt-toolkit==2.0.1
+prompt-toolkit==1.0.15; python_version == '2.7'
+prompt-toolkit==2.0.1; python_version >= '3.4'
 
 # Indirect dependencies for install (not in requirements.txt)
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,7 +10,9 @@
 # Direct dependencies:
 
 # Unit test (imports into testcases):
-pytest>=3.3.0
+# pytest 5.0.0 requires Python version 3.5
+pytest>=3.3.0,<5.0.0; python_version <= '3.4'
+pytest>=3.3.0; python_version >= '3.5'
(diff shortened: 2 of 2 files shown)


minimum-constraints.txt
```diff
@@ -53,7 +53,8 @@ click-spinner==0.1.8
 click-repl==0.1.6
 asciitree==0.3.3
 tabulate==0.8.2
-prompt-toolkit==2.0.1
+prompt-toolkit==1.0.15; python_version == '2.7'
+prompt-toolkit==2.0.1; python_version >= '3.4'
```
andy-maier: This same distinction  also needs to be made in minimum-constraints.txt

KSchopmeyer: Getting late.. that is minimum-constraints.txt

andy-maier: Correct on both points. It was indeed late ...

andy-maier:
This PR now fails because it attempts to install pytest 5.1.0.

Since pytest 5.0.0, pytest requires Python 3.5 or higher. For reasons not yet understood, until yesterday our tests installed pytest 4.6.5 even with latest package level, for example see https://travis-ci.org/pywbem/pywbemtools/jobs/574113604#L828.

It seems to me that something has caused a version pinning of pytest that is now gone. So we need to pin it to <5.0.0 ourselves.

KSchopmeyer:
Made the change to pytest version requirements and rebased to updated master.

PR #262, opened 2019-08-22
Fixes #248: Use of --output-format general option

### LINKED ISSUE #248: Docs: Clarify differences between expected output fmts and received
Clarify the issues in where you can ask for a specific format and get somethiing else.

Thus, 
we do not have a table output for class enumerate, etc. so we output mof
we force a tree output for class tree
even if mof is the default we do not try to output tables of namespaces, profiles, as MOF.

KSchopmeyer:
Fixed documentation to reflect the above and further explain the cases
where the output format general option is honored and ignored.

Found two issues in the output formatting for CIM objects:

1. The repr format was documented and implemented but somewhere we removed
it from the list of available formats. Added it back.

2. The txt format was in the list Set it up. It essentially just
outputs the __str__ for each object.

3. The xml, repr, txt formats had not been tested.  Added tests for the
class command group and TODO for qualdecl and instance tests.
ee
4. The tree format showed up in the list but we never did implement it
that way. It was implemented as the only output for the class tree
command. Removed it from list of output formats

### DIFF
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -112,7 +112,7 @@ The following defines the help output for the `pywbemcli  --help` subcommand
                                       pywbemcli documentation.
                                       Choices: Table:
                                       [table|plain|simple|grid|psql|rst|html],
-                                      Object: [mof|xml|txt|tree]
+                                      Object: [mof|xml|repr|txt]
                                       [Default:
                                       "simple"]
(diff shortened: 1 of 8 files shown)


andy-maier:
Please rebase and resolve the conflict.

andy-maier:
On request by Karl, I have resolved the rebase conflicts, as I thought was right. This means, any corrections to what I did should be made on top of this current state.

I still need to review the PR.

Issue #253, opened 2019-08-21
Docs: 2nd round of comments by Andy on sections 1...3


andy-maier:
Here is a second round of comments on the current master, for sections 1 to 3. It is supposed to contain only comments that require no or not much discussion; larger comments now have their own issues. This issue is now stable, and ready to be processed.

* In 1.1, change "WBEM Servers" to "WBEM servers". **DONE: Cannot find any use of that capitalization in the documentation. Did grep and nothing.**

* In 1.1, simplify "the pywbem Python WBEM client infrastructure" to "the pywbem package". 
**DONE**

* In 1.2, the text "pywbem documentation" in "See the *pywbem documentation* for more details." is not rendered as a link.  **DONE. another case of missing _ at end.**

* In 1.3, in "This section describes the complete of pywbemtools" the word "installation" is missing.
**DONE**

* In 1.3, it says "and manual post-processing steps". We no longer have any manual post-processing steps, so it can be removed.  **DONE**

* In 1.3, on "You can import the pywbemtools Python package into your environment so that pywbemcli can be executed from the command line.": I think this sentence can be removed. We say already earlier that pywbemcli is part of the pywbemtools package. **DONE**

* In 1.3.1.1, the double backquotes around pywbem (\`\`pywbem\`\`) do not render as monospace font, but are shown as backquotes. In most other places, package names are formatted as normal text, so maybe the backquotes should be removed. *DONE**

* In 1.3.1.2, the sentence "This will download and install the latest released version of pywbemtools and its dependent packages into your current Python environment." can be removed. We already say that before the example Pip invocation. **AGREED DONE**

* In 1.3.1.2, on the example with "pip install git+https: ...": In such URLs, the package name should always be set. Add "#egg=pywbemtools" to the end of the URL. **DONE**

* In 1.3.1.3, change "You can verify that the pywbemcli package and" to "You can verify that the pywbemtools package and". **DONE**

* In 1.3.1.3, the text "Invoking with the –version option" shows the option name in normal font. Put it in double backquotes to show as monospace font, consistent with other text showing options. *DONE**

* In 2, there are multipe issues in the text: "This section defines the command line interface specifically for the pywbemcli module within the pywbemtools package. This tool is a single python script executed with the name pywbemcli.".
  Change to: "This section describes the command line interface of the pywbemcli command within the pywbemtools package.". **DONE**

[...]

PR #263, opened 2019-08-22
Fix issues in issue # 253

KSchopmeyer:
This fixes all issues in issue #253.  Note that by the end I was winging
it a bit so the section on using connection select, etc. may not be as
complete as you proposed.  I left a couple of items marked discusison
but applied my proposed solutions.

### DIFF
--- a/docs/pywbemcligeneraloptions.rst
+++ b/docs/pywbemcligeneraloptions.rst
@@ -57,16 +57,16 @@ interactive mode and as a table:
    pywbemcli>
 
 **Note:** - When using the general options as part of an interactive mode
-command, general the options redefinition's are not be retained between command
+command, general the options redefinition's are not retained between command
 executions.
 
 
 General options descriptions
 ----------------------------
 
-The pywbemcli command line options are as follows (See the help in
-pywbemcli and section :ref:`pywbemcli Help Command Details` for more precise
-information on each command arguments and options):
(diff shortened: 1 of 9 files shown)


docs/pywbemcligeneraloptions.rst
```diff
@@ -593,75 +606,82 @@ CIM_Foo_sub_sub is the subclass of CIM_Foo_sub
...
 
-    LOG_CONFIG_STRING := CONFIG[,CONFIG]
-    CONFIG            := COMPONENT"="[DESTINATION[":"DETAIL]
-    COMPONENT         := ('all' / 'api' / 'http')
-    DESTINATION       := ('stderr' / 'file')
-    DETAIL            := ('all'/ 'path'/ 'summary')
+    LOG_CONFIG_STRING := CONFIG[ "," CONFIG]
+    CONFIG            := COMPONENT "=" [DESTINATION[ ":" DETAIL]
+    COMPONENT         := ("all" / "api" / "http")
+    DESTINATION       := ("stderr" / filepath)
```
andy-maier: On filepath: Further down it is stated that the log file has a defined name. The example a few ines down uses 'file', so I think `filepath` should be changed to `"file"` (i.e. as a keyword).

andy-maier: Will be resolved lateron.

PR #266, opened 2019-08-22
Improved help text of pywbemcli, pt 1; Changed querylanguage option

andy-maier:
**Note:** This PR is complete for the help text changes but not for adjusting the corresponding test cases. At this point, only the test cases for the ' class' command group have been adjusted. That is why it currently fails the CI tests. The test cases for the other commands and command groups will need to be adjusted once PR #267 has been completed to cover all of them. Nevertheless, the changed help texts should be reviewed.

Ready for review, considering the note above.

COMMENT: Karl: I completed the tests for the whole environment.  The one thing I have done is to review to be sure I have all options shows for the instance tests.

We have most, but I need to review to assure that they are complete

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -18,8 +18,8 @@ This documentation uses a few special terms to refer to Python types:
 .. glossary::
 
    CIM-XML
-     The name of the protocol used in the DMTF specification :term:`DSP0200`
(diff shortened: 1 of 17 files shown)


andy-maier:
I have added a number of options to the help tests, and have verified that they are now complete. Also, I have moved a few option help textx to the new common file.

PR #277, opened 2019-08-26
Minor improvements for pywbem_os_setup scripts

KSchopmeyer:
Update. Restored the local pywbem_os_setup.sh file and made simple change to Makefile to not get the .bat file when not needed.

See below for discussion items.

Remove local version of pywbem_os_setup.sh which we are not using and
should have removed with pr # 237


andy-maier:
This PR is way behind master. I suggest to rebase it to latest master and try again.

andy-maier:
Here are some findings:

1. In an Appveyor run on Cygwin that succeeds (e.g. https://ci.appveyor.com/project/KSchopmeyer/pywbemtools/builds/27076889/job/umkg5lgjua2jk6ep), it uses `pywbem_os_setup.sh` from the repo, with its "install" argument, and it installs libcrypt and python3-devel. These two packages are not installed in this run. This means the versions of pywbem_os_setup.sh in this repo and in the pywbem repo are probably different. Needs to be investigated.

2. I would have expected that pywbem_os_setup.sh is also invoked with the 'develop' argument, but that is not done, neither in the run that succeeds nor in this run here. After all, tests are supposed to be run. It needs to be investigated why that is not invoked.

3. Low priority: The downloading of pywbem_os_setup.bat could be avoided on CygWin.

andy-maier:
The question is who requites typed-ast. Here is what I found on MacOS:
- It is not installed at all on Python 2.7
- It is not installed in a fresh Python 3.7 virtualenv via ´make install`
- It is installed in Python 3.7 virtualenv via ´make develop`

While I can believe that the dependencies on OS-level packages are different on Cygwin, I have a hard time believing that the Python package typed-ast would be used by "make install" on CygWin when it is not used by "make install" on MacOS.

COuld you please check this again with a fresh Python virtualenv on CygWin (i.e. whether or not "make install" installs typed-ast?

PR #279, opened 2019-08-26
Fixes #251: Changed term subcommand to command

### LINKED ISSUE #251: Improve usage syntax
Copied from issue #241:

I think the notion of command-groups, commands and subcommands is too complicated. How about eliminating the "repl" command, and documenting the remaining commands as follows:

```
pywbemcli <general-options> <object> <action> <action-options> <args>

Where:
    object := class, instance, qualifier, server, connection
    action depends on the object
    args depend on the combination of object and action

    general-options are described in the general help
    action-options are described in the help for each action
```

DISCUSSION(KS): I agree that the subcommands makes it messy. However I am uncomfortable for the future committing that all command-groups will be objects. Look at git. It is really about command groups and commands . I think we can group everything that you have as into just and explain that away. Further if we make a special case of 'help' and 'repl' then everything is .

We really don't need a visible repl command since that is the default when you enter pywbemcli with a set of general options but no command.  However, that leaves the help command which is really an artifact of using repl.

I really think that in general we do not need both <action args> and <args> and can just use args where args := <options> <arguments>

We also have command-groups all over the place now in documentation, etc. plus where click uses that name so not sure about that.

One choice is to simply note that repl and help are outside the mainstream and are the only two commands that do not exist in a group.

KSchopmeyer:
Eliminated all use of subcommand in the help, code, and documentation.

Everything is either a command-group or a command.

Whether we change to object/action or leave it like it is, this makes
the whole thing much simpler.

The only subcommand usage in pywbemcli now is a click functtion
invokde_subcommand that probably started the whole command/subcommand
ruckus.

Changed the name of the doc file pywbemclisubcommands.rst to
pywbemclicommands.rst

Note this involve changes to doc, pywbemcli code, and the tests.

This only resolves part of the issue raised with innsue # 251.  In addition, that issue requests::

1. change "command-group "to "object"

2. Change command to "action"

3. Extend the general syntax definition for what is now <args> to <options> <args>

### DIFF
--- a/docs/pywbemclicmdlineinterface.rst
+++ b/docs/pywbemclicmdlineinterface.rst
@@ -28,31 +28,31 @@ The command line can contain the following components:
 * **general-options** - Options that apply to all commands.
   See :ref:`Pywbemcli command line general options` for infomation on the
   pywbemcli general options
-* **command-group** - A name of a group of subcommands.
-  See :ref:`Pywbemcli command groups, commands, and subcommands`
-* **command** - Alternative to command-group when there are no
-  subcommands.
-* **subcommand** - Command name within a command group
+* **command-group** - A name of a group of commands.
+  See :ref:`Pywbemcli command groups and commands`
+* **command** - A command name normally within a command-group.
+  There are however some special commands that exist outside of any
+  command-group; they are not in any command-group.
 * **args** - Arguments and options that are defined for a particular
-  command or subcommand. Options are proceeded by the characters '-' for the
+  command. Options are proceeded by the characters '-' for the
   short form or '--' for the long form. (ex. ``-n`` or ``--namespace``).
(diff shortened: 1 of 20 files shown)


docs/pywbemclicmdlineinterface.rst
```diff
@@ -167,18 +167,18 @@ list of the supported commands.
     . . .
 
     Commands:
-      class      Command group to manage CIM Classes.
-      instance   Command Group to manage CIM instances.
-      qualifier  Command Group to manage CIM...
+      class      Command-group to manage CIM Classes.
```
andy-maier: Can we use "command group" instead of "Command-Group"?

andy-maier: Yes, change all occurrences of "command-group" in text to "command group".

PR #293, opened 2019-08-29
Removed display of mock repository with --verbose

andy-maier:
Ready for review and merge.


andy-maier:
Response to Karl's comment: The example Python script code shown in PR #291 does already have code that displays the repository if verbose (that's how I realized that it is shown twice :-) )

PR #302, opened 2019-09-01
Docs: Fixed Sphinx build warnings, broken links, typos

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/pywbemcligeneraloptions.rst
+++ b/docs/pywbemcligeneraloptions.rst
@@ -1,4 +1,4 @@
-.. _`Pywbemcli command line general options`:
+.. _`Using the pywbemcli command line general options`:
 
 Using the pywbemcli command line general options
 ================================================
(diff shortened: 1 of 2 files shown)


docs/pywbemcligeneraloptions.rst
```diff
@@ -644,15 +644,15 @@ the use of these operations through two general options:
...
 There are limitations with using the the ``either`` choice as follows"
 
-* The original operations did not support the filtering of responses  with a
-query language query (:ref:`--filter-query-language` and :ref:`--filter-query` )
-option which requests that the WBEM server filter
-the responses before they are returned. This can greatly reduce the size of
-the responses if effectively used but is used only when the pull operations
-are available on the server.
+* The original operations did not support the filtering of responses with a
+  query language query (``--filter-query-language`` and ``--filter-query`` )
```
KSchopmeyer: We have the capability now so how about changing the (\`\`--filter-query-language\`\` and \`\`--filter-query\`\` )  to ( :ref:\`--filter-query-language general options\` and :ref:\`--filter-query general option\`)

andy-maier: We don't have that capability for these options, because we only have it for general options and these are command specific options.

PR #303, opened 2019-09-01
Fixes #282: Added section for specifying INSTANCENAME

### LINKED ISSUE #282: Section about specifying INSTANCENAME
Issuing a command that needs an INSTANCENAME fails as follows when specifying the syntax shown in the usage examples.

```
$ pywbemcli -m tests/unit/simple_mock_model.mof instance get CIM_Foo.InstanceID=CIM_Foo1 -n root/cimv2
Error: Invalid wbem uri input :CIM_Foo.InstanceID=CIM_Foo1. Error WBEM URI has invalid value format in a keybinding: 'InstanceID'='CIM_Foo1'
```

Specifying the keybinding value with double quotes does not help:

```
$ pywbemcli -m tests/unit/simple_mock_model.mof instance get CIM_Foo.InstanceID="CIM_Foo1" -n root/cimv2
Error: Invalid wbem uri input :CIM_Foo.InstanceID=CIM_Foo1. Error WBEM URI has invalid value format in a keybinding: 'InstanceID'='CIM_Foo1'
```

Some more information, showing that the instance exists and the specified property is a key property:

```
$ pywbemcli -m tests/unit/simple_mock_model.mof -o table instance enumerate CIM_Foo
Instances: CIM_Foo
+--------------+---------------+
| InstanceID   | IntegerProp   |
|--------------+---------------|
| "CIM_Foo1"   | 1             |
| "CIM_Foo2"   | 2             |
| "CIM_Foo3"   |               |
+--------------+---------------+

$ pywbemcli -m tests/unit/simple_mock_model.mof class get CIM_Foo -n root/cimv2
   [Description ( "Simple CIM Class" )]
class CIM_Foo {

      [Key ( true ),
       Description ( "This is key property." )]
   string InstanceID;

      [Description ( "This is Uint32 property." )]
   uint32 IntegerProp;

      [Description ( "Method with in and out parameters" )]
   uint32 Fuzzy(
         [IN ( true ),
          OUT ( true ),
          Description ( "Define data to be returned in output parameter" )]
      string TestInOutParameter,
         [IN ( true ),
          OUT ( true ),
          Description ( "Test of ref in/out parameter" )]
      CIM_Foo REF TestRef,
         [IN ( false ),
          OUT ( true ),
          Description ( "Rtns method name if exists on input" )]
      string OutputParam,
         [IN ( true ),
          Description ( "Defines return value if provided." )]
      uint32 OutputRtnValue);

      [Description ( "Method with no Parameters" )]
   uint32 DeleteNothing();

};
```

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/pywbemclicommands.rst
+++ b/docs/pywbemclicommands.rst
@@ -166,18 +166,106 @@ Thus, for example an ``instance enumerate`` with and without the ``-o`` option:
 
     root/cimv2:CIM_Foo.InstanceID="CIM_Foo3"
 
+.. _`Specifying the INSTANCENAME command argument`:
+
+Specifying the INSTANCENAME command argument
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The INSTANCENAME argument used by some pywbemcli commands (e.g ``instance get``)
+specifies the instance path (aka instance name) of a CIM instance in a CIM
+namespace of a WBEM server.
+
+The format used by pywbemcli for specifying INSTANCENAME arguments on the
+command line is an untyped WBEM URI for instance paths as defined in
+:term:`DSP0207`.
+Because pywbemcli always works with a single WBEM server at a time, the
+authority component of the WBEM URI is never specified in an INSTANCENAME.
+Because the namespace type of the WBEM URI (e.g. http or https) is not relevant
+for identifying the CIM instance, the namespace type is never specified in an
+INSTANCENAME.
+
+With these simplifications, the WBEM URI compatible ABNF for INSTANCENAME is:
+
+.. code-block:: text
+
+   INSTANCENAME = "/" [ NAMESPACE ] ":" CLASSNAME [ "." keybindings ]
+
+   keybindings = keybinding *( "," keybinding )
+
+   keybinding = PROPERTYNAME "=" value
+
+   value = integerValue / charValue / stringValue / datetimeValue / booleanValue / referenceValue
+
+   referenceValue = "\"" escaped_INSTANCENAME "\""
+
+where:
+
+* NAMESPACE, CLASSNAME and PROPERTYNAME are namespace, class and key
+  property name, respectively, as used elsewhere in pywbemcli.
+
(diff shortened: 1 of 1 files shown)


docs/pywbemclicommands.rst
```diff
@@ -166,18 +166,106 @@ Thus, for example an ``instance enumerate`` with and without the ``-o`` option:
 
     root/cimv2:CIM_Foo.InstanceID="CIM_Foo3"
 
+.. _`Specifying INSTANCENAME`:
+
+Specifying INSTANCENAME
```
KSchopmeyer: How about "Specifying INSTANCENAME command argument"

docs/pywbemclicommands.rst
```diff
@@ -166,18 +166,106 @@ Thus, for example an ``instance enumerate`` with and without the ``-o`` option:
...
-``get``, ``references``, ``associators``, etc) can be very difficult to correctly enter
-since it can involve multiple keybindings, use of quotation marks, etc.  To
-simplify this pywbemcli includes a option (``-i`` or ``--interactive``) on
-these commands that allows the user to specify only the class name, retrieves
-all the instance names from the server and presents the user with a select list
-from which an instance name can be chosen. The following is an example:
+The INSTANCENAME argument has a certain complexity, particularly for
+associations and for classes with multiple keys.
+
+To simplify this, pywbemcli provides an option (``-i`` or ``--interactive``) on
```
KSchopmeyer: What I would really like to do is to drop the interactive and even if it is only a classname, do a get, expecting an error return (just to cover the normally nonexistent singleton instance (i.e. no keys) then if it was just a classname do the select. But that is for later.

docs/pywbemclicommands.rst
andy-maier: DONE; Changed to "Specifying the INSTANCENAME command argument"

docs/pywbemclicommands.rst
andy-maier: I don't like that, because this approach cannot distinguish between a user error and the user intention to select interactively.

PR #306, opened 2019-09-02
Fixes #261: inconsistent connection cmds

### LINKED ISSUE #261: Inconsistent way connections are named in connection subcommands
The subcommands within the `connection` command group are inconsistent in the way the connection is named. I was going by their help text and did not verify whether the actual behavior is as described there.

The following subcommands use an existing connection:

* delete - names the connection via the optional NAME argument. If not specified, it prompts.
* select - names the connection via the optional NAME argument. If not specified, it selects current connection.
* show - names the connection via the optional NAME argument. If not specified, it prompts.
* test - only works on current connection.

The following subcommands create a new connection:

* add - names new connection via required --name subcommand option.
* save - names new connection via optional --name subcommand option, and with a quite complex behavior:
   - If specified, and the current connection was specified using the --name general option, it renames that connection to the name from the --name subcommand option. I consider that a very unexpected side effect of what "save" is otherwise described and/or expected to do.
  - If specified, and the current connection was not specified using the --name general option, it (probably?) creates a new connection with the name from the --name subcommand option.
  - If not specified, it is not clear where the new connection gets its name from (at least if the current connection was specified using --server or --mock-server general option). Remember, the --name general option is mutually exclusive with --server and --mock-server.

The remaining subcommands operate on all connections and do not deal with specifying names.

I suggest that we make it consistent how connections are named.

KSchopmeyer:
See commit for details

### DIFF
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -701,16 +701,17 @@ The following defines the help output for the `pywbemcli connection add --help` 
       properties are optional.
 
       Adding a connection does not set the new connection as the current
-      connection definition. Use `connection select` to set a particular stored
-      connection definition as the current connection.
+      connection definition. Use the connection select command to set a
+      particular connection definition as the current connection.
 
         pywbemcli connection add newsvr --server https://srv1
 
     Options:
-      --input-name INPUT-NAME         If this option exists, it is the name of a
-                                      persistent connection that will be put into
-                                      the connections file with the NAME argument.
-                                      All other options will be ignored.
+      --input-name CONNECTION-NAME    If this option exists, it is the name of an
+                                      existing persistent definition that will be
+                                      added to the connections file with the NAME
+                                      argument.  All other options will be
+                                      ignored.
       -m, --mock-server FILE          Use a mock WBEM server that is automatically
                                       created in pywbemcli and populated with CIM
                                       objects that are defined in the specified
@@ -867,7 +868,7 @@ The following defines the help output for the `pywbemcli connection list --help`
       List the WBEM connection definitions.
 
       This command displays all entries in the connections file and the current
-      connection if it is not in the connections file as a table.
+      connection if it exists and is not in the connections file as a table.
 
       An "!" before the name indicates the default connection. See connection
       select. A '*' before the name indicates that it is the current connection.
@@ -902,13 +903,10 @@ The following defines the help output for the `pywbemcli connection save --help`
         pywbemcli --server https://srv1 connection save mysrv
 
     Options:
-      --input-name INPUT-NAME  If this option exists, it is the name of a
-                               persistent connection that will be put into the
-                               connections file with the NAME argument.
-      -V, --verify             Prompt for confirmation before performing a change,
-                               to allow for verification of parameters. Default:
-                               Do not prompt for confirmation.
-      -h, --help               Show this message and exit.
+      -V, --verify  Prompt for confirmation before performing a change, to allow
+                    for verification of parameters. Default: Do not prompt for
+                    confirmation.
+      -h, --help    Show this message and exit.
 
 
 .. _`pywbemcli connection select --help`:
@@ -928,15 +926,18 @@ The following defines the help output for the `pywbemcli connection select --hel
       Select a WBEM connection definition as current/default connection.
 
       Selects a connection from the persistently stored named connections if
-      NAME exists in the store. If NAME not supplied, a list of connections from
-      the connections definition file is presented with a prompt for the user to
-      select a connection.
+      NAME exists in the store to be the current connection. If the NAME
+      argument not supplied, a list of connections from the connections
+      definition file is presented with a prompt for the user to select a
+      connection.
 
-      Selection is persistent; it creates a default connection and that used as
-      the server definition in commands in the existing an interactive session
-      and is saved in the connections file and used as the current connection if
-      there is no other server definition (--server or --name or --mock-server)
-      general option.
+      Selection is persistent if the --default option exists; the selected
+      connection will be used as the server definition in commands future
+      execution of pywbemcli if there is no other server definition (--server or
+      --name or --mock-server) general option.
+
+      Otherwise it sets the connection selected to be the current connection in
+      the current interactive session.
 
       Example:
         $ pywbemcli
@@ -947,8 +948,9 @@ The following defines the help output for the `pywbemcli connection select --hel
           server: http://localhost
 
     Options:
-      -d, --default  If set, the connection is set to default in the connections
-                     file in addition to setting it as the current connection.
+      -d, --default  If set, the connection is set to be the default connection
+                     in the connections file in addition to setting it as the
+                     current connection.
       -h, --help     Show this message and exit.
 
 
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -670,8 +670,8 @@ The following defines the help output for the `pywbemcli connection --help` comm
       delete  Delete a WBEM connection definition.
       export  Export the current connection.
       list    List the WBEM connection definitions.
-      save    Save the current connection as a new WBEM connection definition.
-      select  Select a WBEM connection definition as the current connection.
+      save    Save a connection to a new WBEM connection definition named NAME.
+      select  Select a WBEM connection definition as current/default connection.
       show    Show connection info of a WBEM connection definition.
       test    Test the current connection with a predefined WBEM request.
 
@@ -692,22 +692,25 @@ The following defines the help output for the `pywbemcli connection add --help` 
 
       Add a new WBEM connection definition from specified options.
 
-      Create a new WBEM connection definition in the connections file from the
-      specified options. A connection definition with that name must not yet
-      exist.
+      Create a new WBEM connection definition named NAME in the connections file
+      from the specified options. A connection definition with that name must
+      not yet exist.
 
-      The NAME argument MUST exist. It define the server uri and the unique name
-      under which this server connection information will be stored. A --server
-      or --mock-server must exist because they define the server. All other
+      The NAME argument is required. One of --server,  --mock-server, or --name
+      option  is required to define the server for the new connection. All other
       properties are optional.
 
       Adding a connection does not set the new connection as the current
-      connection. Use `connection select` to set a particular stored connection
-      definition as the current connection.
+      connection definition. Use `connection select` to set a particular stored
+      connection definition as the current connection.
 
-        pywbemcli --name newsrv connection add --server https://srv1
+        pywbemcli connection add newsvr --server https://srv1
 
     Options:
+      --input-name INPUT-NAME         If this option exists, it is the name of a
+                                      persistent connection that will be put into
+                                      the connections file with the NAME argument.
+                                      All other options will be ignored.
       -m, --mock-server FILE          Use a mock WBEM server that is automatically
                                       created in pywbemcli and populated with CIM
                                       objects that are defined in the specified
@@ -864,11 +867,10 @@ The following defines the help output for the `pywbemcli connection list --help`
       List the WBEM connection definitions.
 
       This command displays all entries in the connections file and the current
-      connection if it is not in the connections file as a table using the
-      command line output_format to define the table format.
+      connection if it is not in the connections file as a table.
 
-      An "!" before the name indicates the selected connection. A '!' before the
-      name indicates that it is the current connection.
+      An "!" before the name indicates the default connection. See connection
+      select. A '*' before the name indicates that it is the current connection.
 
     Options:
       -h, --help  Show this message and exit.
@@ -888,21 +890,25 @@ The following defines the help output for the `pywbemcli connection save --help`
 
     Usage: pywbemcli connection save [COMMAND-OPTIONS] NAME
 
-      Save the current connection as a new WBEM connection definition.
+      Save a connection to a new WBEM connection definition named NAME.
 
       Create a new WBEM connection definition in the connections file from the
-      current connection with the NAME argument as connection name. A connection
-      definition with that name must not yet exist.
+      current connection or the connection definition in --input-name anamed
+      NAME. A connection definition with the name NAME must not yet exist. The
+      NAME argument is required.
 
       Examples:
 
         pywbemcli --server https://srv1 connection save mysrv
 
     Options:
-      -V, --verify  Prompt for confirmation before performing a change, to allow
-                    for verification of parameters. Default: Do not prompt for
-                    confirmation.
-      -h, --help    Show this message and exit.
+      --input-name INPUT-NAME  If this option exists, it is the name of a
+                               persistent connection that will be put into the
+                               connections file with the NAME argument.
+      -V, --verify             Prompt for confirmation before performing a change,
+                               to allow for verification of parameters. Default:
+                               Do not prompt for confirmation.
+      -h, --help               Show this message and exit.
 
 
 .. _`pywbemcli connection select --help`:
@@ -919,31 +925,31 @@ The following defines the help output for the `pywbemcli connection select --hel
 
     Usage: pywbemcli connection select [COMMAND-OPTIONS] NAME
 
-      Select a WBEM connection definition as the current connection.
+      Select a WBEM connection definition as current/default connection.
 
-      Make a named connection definition from the connections file the current
-      connection. If the NAME argument is omitted, prompt for selecting one of
-      the connection definitions in the connections file.
+      Selects a connection from the persistently stored named connections if
+      NAME exists in the store. If NAME not supplied, a list of connections from
+      the connections definition file is presented with a prompt for the user to
+      select a connection.
 
-      Selects a connection from the persistently stored set of named connections
-      if NAME exists in the store. If NAME not supplied, a list of connections
-      from the connections definition file is presented with a prompt for the
-      user to select a connection.
+      Selection is persistent; it creates a default connection and that used as
+      the server definition in commands in the existing an interactive session
+      and is saved in the connections file and used as the current connection if
+      there is no other server definition (--server or --name or --mock-server)
+      general option.
 
-      Selection is persistent; it is used as the server definition in commands
-      in the remainder of an interactive session and is saved in the connections
-      file and used as the current connection if there is no other server
-      definition (--server or --name or --mock-server).
-
-      Examples:
-
+      Example:
         $ pywbemcli
-
-        pywbemcli> connection select myconn   pywbemcli> connection show
-        pywbemcli> CTRL-D   $ pywbemcli show     TODO     ...
+        pywbemcli> connection select myconn
+        pywbemcli> :quit
+        $ pywbemcli show
+        name: myconn
+          server: http://localhost
 
     Options:
-      -h, --help  Show this message and exit.
+      -d, --default  If set, the connection is set to default in the connections
+                     file in addition to setting it as the current connection.
+      -h, --help     Show this message and exit.
 
 
 .. _`pywbemcli connection show --help`:
@@ -965,8 +971,8 @@ The following defines the help output for the `pywbemcli connection show --help`
       This command displays the WBEM connection definition of a single
       connection as follows:
 
-      * A named connection from the WBEM connections file if NAME argument
-      exists.
+      * A named connection from the WBEM connections file if NAME argument is
+      specified.
 
       * The current connection if NAME is not provided and a current connection
       exists (selected, or defined on the command line).
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -671,7 +671,7 @@ The following defines the help output for the `pywbemcli connection --help` comm
       export  Export the current connection.
       list    List the WBEM connection definitions.
       save    Save a connection to a new WBEM connection definition named NAME.
-      select  Select a WBEM connection definition as current/default connection.
+      select  Select a WBEM connection definition as current or default.
       show    Show connection info of a WBEM connection definition.
       test    Test the current connection with a predefined WBEM request.
 
@@ -870,8 +870,9 @@ The following defines the help output for the `pywbemcli connection list --help`
       This command displays all entries in the connections file and the current
       connection if it exists and is not in the connections file as a table.
 
-      An "!" before the name indicates the default connection. See connection
-      select. A '*' before the name indicates that it is the current connection.
+      '#' before the name indicates the default connection.
+      '*' before the name indicates that it is the current connection.
+      See connection select.
 
     Options:
       -h, --help  Show this message and exit.
@@ -894,9 +895,8 @@ The following defines the help output for the `pywbemcli connection save --help`
       Save a connection to a new WBEM connection definition named NAME.
 
       Create a new WBEM connection definition in the connections file from the
-      current connection or the connection definition in --input-name anamed
-      NAME. A connection definition with the name NAME must not yet exist. The
-      NAME argument is required.
+      current connection A connection definition with the name NAME must not yet
+      exist. The NAME argument (the name of the new connection) is required.
 
       Examples:
 
@@ -923,25 +923,25 @@ The following defines the help output for the `pywbemcli connection select --hel
 
     Usage: pywbemcli connection select [COMMAND-OPTIONS] NAME
 
-      Select a WBEM connection definition as current/default connection.
+      Select a WBEM connection definition as current or default.
 
       Selects a connection from the persistently stored named connections if
       NAME exists in the store to be the current connection. If the NAME
-      argument not supplied, a list of connections from the connections
+      argument does not exist, a list of connections from the connections
       definition file is presented with a prompt for the user to select a
       connection.
 
-      Selection is persistent if the --default option exists; the selected
-      connection will be used as the server definition in commands future
-      execution of pywbemcli if there is no other server definition (--server or
-      --name or --mock-server) general option.
+      Default and current connction are set if the --default option exists;
+      otherwise only the connection in the current interactive session is set .
+      Once define the default connection will be used as the server definition
+      in future execution of pywbemcli if there is no server definition
+      (--server or --name or --mock-server) general option.
 
-      Otherwise it sets the connection selected to be the current connection in
-      the current interactive session.
+      Connection list shows all connections and if any are default or current.
 
-      Example:
+      Example of selecting a connection and setting the default:
         $ pywbemcli
-        pywbemcli> connection select myconn
+        pywbemcli> connection select myconn --default
         pywbemcli> :quit
         $ pywbemcli show
         name: myconn
@@ -973,20 +973,25 @@ The following defines the help output for the `pywbemcli connection show --help`
       This command displays the WBEM connection definition of a single
       connection as follows:
 
-      * A named connection from the WBEM connections file if NAME argument is
-      specified.
+      * NAME argument exists; show that WBEM connections file.
 
-      * The current connection if NAME is not provided and a current connection
-      exists (selected, or defined on the command line).
+      * NAME argument does not exist; show existing current connection or
+      presents list for selection
 
-      * User selects from list presented if NAME argument is '?' or there is no
-      NAME argument and no current connection.
+      * Name argument is '?'; presents list of connections for selection
 
-      This command displays all the variables that make up the current WBEM.
+      This command displays all the variables that make up current WBEM
+      connection.
 
-        pywbemcli connection show server1
-
+        pywbemcli connection show svr1
+          name: svr1
+          ...
         pywbemcli connection show ?
+          0: svr1
+          1: svr2
+      Input integer between 0 and 2 or Ctrl-C to exit selection: : 0
+        name: svr1
+        ...
 
     Options:
       -h, --help  Show this message and exit.
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -672,7 +672,7 @@ The following defines the help output for the `pywbemcli connection --help` comm
       list    List the WBEM connection definitions.
       save    Save the current connection as a new WBEM connection definition.
       select  Select a WBEM connection definition as the current connection.
-      show    Show the current connection or a WBEM connection definition.
+      show    Show connection info of a WBEM connection definition.
       test    Test the current connection with a predefined WBEM request.
 
 
@@ -688,7 +688,7 @@ The following defines the help output for the `pywbemcli connection add --help` 
 
 ::
 
-    Usage: pywbemcli connection add [COMMAND-OPTIONS]
+    Usage: pywbemcli connection add [COMMAND-OPTIONS] NAME
 
       Add a new WBEM connection definition from specified options.
 
@@ -696,15 +696,18 @@ The following defines the help output for the `pywbemcli connection add --help` 
       specified options. A connection definition with that name must not yet
       exist.
 
-      The current connection remains unchanged by this command.
+      The NAME argument MUST exist. It define the server uri and the unique name
+      under which this server connection information will be stored. A --server
+      or --mock-server must exist because they define the server. All other
+      properties are optional.
 
-      Examples:
+      Adding a connection does not set the new connection as the current
+      connection. Use `connection select` to set a particular stored connection
+      definition as the current connection.
 
         pywbemcli --name newsrv connection add --server https://srv1
 
     Options:
-      -n, --name NAME                 Name for the new WBEM connection definition.
-                                      [required]
       -m, --mock-server FILE          Use a mock WBEM server that is automatically
                                       created in pywbemcli and populated with CIM
                                       objects that are defined in the specified
@@ -860,13 +863,12 @@ The following defines the help output for the `pywbemcli connection list --help`
 
       List the WBEM connection definitions.
 
-      List the WBEM connection definitions in the connections file as a table.
+      This command displays all entries in the connections file and the current
+      connection if it is not in the connections file as a table using the
+      command line output_format to define the table format.
 
-      In the table, the current connection is marked with an "*" after its name.
-
-      Examples:
-
-        pywbemcli connection list
+      An "!" before the name indicates the selected connection. A '!' before the
+      name indicates that it is the current connection.
 
     Options:
       -h, --help  Show this message and exit.
@@ -884,27 +886,23 @@ The following defines the help output for the `pywbemcli connection save --help`
 
 ::
 
-    Usage: pywbemcli connection save [COMMAND-OPTIONS]
+    Usage: pywbemcli connection save [COMMAND-OPTIONS] NAME
 
       Save the current connection as a new WBEM connection definition.
 
       Create a new WBEM connection definition in the connections file from the
-      current connection. A connection definition with that name must not yet
-      exist.
-
-      This command is useful when you have defined a connection on the command
-      line and want to save it in the connections file for later use.
+      current connection with the NAME argument as connection name. A connection
+      definition with that name must not yet exist.
 
       Examples:
 
-        pywbemcli --server https://srv1 connection save --name mysrv
+        pywbemcli --server https://srv1 connection save mysrv
 
     Options:
-      -n, --name NAME  Name for the new WBEM connection definition.
-      -V, --verify     Prompt for confirmation before performing a change, to
-                       allow for verification of parameters. Default: Do not
-                       prompt for confirmation.
-      -h, --help       Show this message and exit.
+      -V, --verify  Prompt for confirmation before performing a change, to allow
+                    for verification of parameters. Default: Do not prompt for
+                    confirmation.
+      -h, --help    Show this message and exit.
 
 
 .. _`pywbemcli connection select --help`:
@@ -927,18 +925,22 @@ The following defines the help output for the `pywbemcli connection select --hel
       connection. If the NAME argument is omitted, prompt for selecting one of
       the connection definitions in the connections file.
 
-      This command is useful in the interactive mode of pywbemcli.
+      Selects a connection from the persistently stored set of named connections
+      if NAME exists in the store. If NAME not supplied, a list of connections
+      from the connections definition file is presented with a prompt for the
+      user to select a connection.
 
-      The selection of the current connection is not saved across interactive
-      sessions of pywbemcli.
+      Selection is persistent; it is used as the server definition in commands
+      in the remainder of an interactive session and is saved in the connections
+      file and used as the current connection if there is no other server
+      definition (--server or --name or --mock-server).
 
       Examples:
 
-        pywbemcli
+        $ pywbemcli
 
-        > connection select myconn
-
-        > connection select          # prompts
+        pywbemcli> connection select myconn   pywbemcli> connection show
+        pywbemcli> CTRL-D   $ pywbemcli show     TODO     ...
 
     Options:
       -h, --help  Show this message and exit.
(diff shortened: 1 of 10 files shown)


docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
 
 ::
 
-    Usage: pywbemcli connection add [COMMAND-OPTIONS]
+    Usage: pywbemcli connection add [COMMAND-OPTIONS] NAME
 
       Add a new WBEM connection definition from specified options.
 
       Create a new WBEM connection definition in the connections file from the
```
andy-maier: I suggest changing this sentence to be explicit that NAME is the new name, as follows: "Create a new WBEM connection definition named NAME in the connections file...".

docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
...
       specified options. A connection definition with that name must not yet
       exist.
 
-      The current connection remains unchanged by this command.
+      The NAME argument MUST exist. It define the server uri and the unique name
```
andy-maier: To me it does not sound OK to talk about "existing" options. Suggest to change "The NAME argument MUST exist." to "The NAME argument is required.".

andy-maier: Multiple issues in: "It define the server uri and the unique name under which this server connection information will be stored.":
* Grammar: "it define"
* It does not define the server URI.
If we follow my suggestion on changing the sentence on line 693, the whole sentence can be removed.

docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
...
+      under which this server connection information will be stored. A --server
+      or --mock-server must exist because they define the server. All other
```
andy-maier: On "A --server or --mock-server must exist because they define the server.":
1. Again the wording "exist"
2. Missing word "option" after the option names
3. The --name option also works.
I suggest to change the quoted sentence to: "One of the `--server`, `--mock-server` or `--name` options must be specified to define the WBEM server for the new connection definition."

docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
...
+      properties are optional.
 
-      Examples:
+      Adding a connection does not set the new connection as the current
+      connection. Use `connection select` to set a particular stored connection
```
andy-maier: On "Use `connection select` to": Change to "Use the connection select command to".

docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
...
+      definition as the current connection.
 
         pywbemcli --name newsrv connection add --server https://srv1
```
andy-maier: Update the example to use NAME, e.g.:

pywbemcli connection add newsrv --server https://srv1

docs/pywbemclicmdshelp.rst
```diff
@@ -686,23 +686,26 @@ The following defines the help output for the `pywbemcli connection add --help`
...
 
     Options:
-      -n, --name NAME                 Name for the new WBEM connection definition.
-                                      [required]
```
andy-maier: Why is --name removed? It can very well be used to use a previously defined connection defin ition as input for the new connection definition.

docs/pywbemclicmdshelp.rst
```diff
@@ -858,13 +861,12 @@ The following defines the help output for the `pywbemcli connection list --help`
...
       List the WBEM connection definitions.
 
-      List the WBEM connection definitions in the connections file as a table.
-
-      In the table, the current connection is marked with an "*" after its name.
-
-      Examples:
+      This command displays all entries in the connections file and the current
+      connection if it is not in the connections file as a table using the
+      command line output_format to define the table format.
```
andy-maier: A current connection is not always defined, so the sentence should allow for that case as well.

andy-maier: On "using the command line output_format to define the table format.":
1. Don't describe this using the name of an internal variable (output_format).
2. I don't think we need to repeat every time we say an output is a table that the format of the table can be controlled.

I suggest to drop the quoted part of the sentence.

docs/pywbemclicmdshelp.rst
```diff
@@ -858,13 +861,12 @@ The following defines the help output for the `pywbemcli connection list --help`
...
 
-        pywbemcli connection list
+      An "!" before the name indicates the selected connection. A '!' before the
+      name indicates that it is the current connection.
```
andy-maier: Multiple issues:
1. This paragraph talks two times about '"!".
2. It is not clear what a "selected" connection is other than the new current connection. So I think there is no need to indicate a selected connection separately from the current connection.
3. You also introduced the concept of a default connectzion, but it is missing in this description.

docs/pywbemclicmdshelp.rst
```diff
@@ -882,27 +884,23 @@ The following defines the help output for the `pywbemcli connection save --help`
 
 ::
 
-    Usage: pywbemcli connection save [COMMAND-OPTIONS]
+    Usage: pywbemcli connection save [COMMAND-OPTIONS] NAME
 
       Save the current connection as a new WBEM connection definition.
 
       Create a new WBEM connection definition in the connections file from the
```
andy-maier: Add "named NAME" to this sentence, i.e. "Create a new WBEM connection definition named NAME in the".

docs/pywbemclicmdshelp.rst
```diff
@@ -882,27 +884,23 @@ The following defines the help output for the `pywbemcli connection save --help`
...
-      current connection. A connection definition with that name must not yet
-      exist.
-
-      This command is useful when you have defined a connection on the command
-      line and want to save it in the connections file for later use.
+      current connection with the NAME argument as connection name. A connection
+      definition with that name must not yet exist.
```
andy-maier: Add "The NAME argument is required".

docs/pywbemclicmdshelp.rst
```diff
@@ -882,27 +884,23 @@ The following defines the help output for the `pywbemcli connection save --help`
...
 
       Save the current connection as a new WBEM connection definition.
 
       Create a new WBEM connection definition in the connections file from the
-      current connection. A connection definition with that name must not yet
-      exist.
-
-      This command is useful when you have defined a connection on the command
-      line and want to save it in the connections file for later use.
+      current connection with the NAME argument as connection name. A connection
```
andy-maier: On " with the NAME argument as connection name":
1. To be precise, it is the connection definition name.
2. Along with my suggestion on "connection add", I suggest to shorten the first sentence to: "Create a new WBEM connection definition named NAME in the connections file from the current connection."

docs/pywbemclicmdshelp.rst
```diff
@@ -925,18 +923,22 @@ The following defines the help output for the `pywbemcli connection select --hel
       connection. If the NAME argument is omitted, prompt for selecting one of
       the connection definitions in the connections file.
```
andy-maier: The previous paragraph (starting on line 922) already describes this. The new paragraph has no new information and can be removed.

docs/pywbemclicmdshelp.rst
```diff
@@ -925,18 +923,22 @@ The following defines the help output for the `pywbemcli connection select --hel
...
+      if NAME exists in the store. If NAME not supplied, a list of connections
+      from the connections definition file is presented with a prompt for the
+      user to select a connection.
 
-      The selection of the current connection is not saved across interactive
-      sessions of pywbemcli.
+      Selection is persistent; it is used as the server definition in commands
+      in the remainder of an interactive session and is saved in the connections
+      file and used as the current connection if there is no other server
+      definition (--server or --name or --mock-server).
```
andy-maier: This is confusing. Does the command set the current connection (as it previously was described to do)? Or the default connection? Or both?

docs/pywbemclicmdshelp.rst
```diff
@@ -925,18 +923,22 @@ The following defines the help output for the `pywbemcli connection select --hel
...
       Examples:
 
-        pywbemcli
+        $ pywbemcli
 
-        > connection select myconn
-
-        > connection select          # prompts
+        pywbemcli> connection select myconn   pywbemcli> connection show
+        pywbemcli> CTRL-D   $ pywbemcli show     TODO     ...
```
andy-maier: This does not format right. You need to put each line between blank lines. Given that, this is not the place to explain larger sequences of commands. I suggest to remove the show commands.

docs/pywbemclicommands.rst
```diff
@@ -1009,7 +1008,7 @@ The commands in this group are:
 Connection add command
 ^^^^^^^^^^^^^^^^^^^^^^
 
-The ``connection add`` command creates a new connection using the command arguments and sets the new
+The ``connection add`` command creates a new connection using the command arguments and options and sets the new
 connection as the current connection. This command saves the
```
andy-maier: The text "and sets the new connection as the current connection." is likely incorrect, given the description of "connection add" that claims the opposite.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -80,17 +80,25 @@ def connection_export(context):
...
+    Show connection info of a WBEM connection definition.
 
-    Display the parameters of the current connection, or if the optional NAME
-    argument is specified, of the named WBEM connection definition from the
-    connections file.
+    This command displays the WBEM connection definition of a single
+    connection as follows:
 
-    Examples:
+    * A named connection from the WBEM connections file if NAME argument exists.
```
andy-maier: Change "exists" to "is specified".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,26 @@ def connection_select(context, name):
...
+def connection_add(context, name, **options):
     """
     Add a new WBEM connection definition from specified options.
 
     Create a new WBEM connection definition in the connections file from
     the specified options. A connection definition with that name must not
     yet exist.
 
-    The current connection remains unchanged by this command.
+    The NAME argument MUST exist. It define the server uri
```
andy-maier: Change "MUST exist" to "is required".

docs/pywbemclicmdshelp.rst
```diff
@@ -956,18 +958,26 @@ The following defines the help output for the `pywbemcli connection show --help`
...
 
-      Display the parameters of the current connection, or if the optional NAME
-      argument is specified, of the named WBEM connection definition from the
-      connections file.
+      This command displays the WBEM connection definition of a single
+      connection as follows:
 
-      Examples:
+      * A named connection from the WBEM connections file if NAME argument
+      exists.
```
andy-maier: Change "if NAME argument exists" to "if the NAME argument is specified".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,26 @@ def connection_select(context, name):
                      default='all'))
 @add_options(verify_option)
 @click.pass_obj
-def connection_add(context, **options):
+def connection_add(context, name, **options):
     """
     Add a new WBEM connection definition from specified options.
 
     Create a new WBEM connection definition in the connections file from
```
andy-maier: I suggest changing this sentence to be explicit that NAME is the new name, as follows: "Create a new WBEM connection definition named NAME in the connections file...".

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: To me it does not sound OK to talk about "existing" options. Suggest to change "The NAME argument MUST exist." to "The NAME argument is required.".

andy-maier: Multiple issues in: "It define the server uri and the unique name under which this server connection information will be stored.":
* Grammar: "it define"
* It does not define the server URI.
If we follow my suggestion on changing the sentence on line 693, the whole sentence can be removed.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,26 @@ def connection_select(context, name):
...
     the specified options. A connection definition with that name must not
     yet exist.
 
-    The current connection remains unchanged by this command.
+    The NAME argument MUST exist. It define the server uri
+    and the unique name under which this server connection information
+    will be stored. A --server or --mock-server must exist because they
+    define the server. All other properties are optional.
```
andy-maier: On "A --server or --mock-server must exist because they define the server.":

1. Again the wording "exist"
2. Missing word "option" after the option names
3. The --name option also works.

I suggest to change the quoted sentence to:

"One of the --server, --mock-server or --name options must be specified to define the WBEM server for the new connection definition."

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,26 @@ def connection_select(context, name):
...
 
-    Examples:
+    Adding a connection does not set the new connection as the current
+    connection. Use `connection select` to set a particular stored connection
```
andy-maier: On "Use `connection select` to":
1. Do not use back quotes in help text (at least so far we don't, so be consistent).
2. Talk about "the xxx command".

Change to:

"Use the connection select command to".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,26 @@ def connection_select(context, name):
...
+    definition as the current connection.
 
       pywbemcli --name newsrv connection add --server https://srv1
```
andy-maier: Update the example to use NAME, e.g.:

pywbemcli connection add newsrv --server https://srv1

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -286,27 +312,22 @@ def connection_test(context):
 
 
 @connection_group.command('save', options_metavar=CMD_OPTS_TXT)
-@click.option('-n', '--name', type=str, metavar="NAME",
-              default=None,
-              help='Name for the new WBEM connection definition.')
```
andy-maier: Why is --name removed? It can very well be used to use a previously defined connection defin ition as input for the new connection definition.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -315,13 +336,13 @@ def connection_list(context):
     """
     List the WBEM connection definitions.
 
-    List the WBEM connection definitions in the connections file as a table.
-
-    In the table, the current connection is marked with an "*" after its name.
+    This command displays all entries in the connections file and the
+    current connection if it is not in the connections file as
+    a table using the command line output_format to define the
+    table format.
```
andy-maier: A current connection is not always defined, so the sentence should allow for that case as well.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -315,13 +336,13 @@ def connection_list(context):
     """
     List the WBEM connection definitions.
 
-    List the WBEM connection definitions in the connections file as a table.
-
-    In the table, the current connection is marked with an "*" after its name.
+    This command displays all entries in the connections file and the
+    current connection if it is not in the connections file as
```
andy-maier: A current connection is not always defined, so the sentence should allow for that case as well.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: On "using the command line output_format to define the table format.":
1. Don't describe this using the name of an internal variable (output_format).
2. I don't think we need to repeat every time we say an output is a table that the format of the table can be controlled.

I suggest to drop the quoted part of the sentence.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -315,13 +336,13 @@ def connection_list(context):
...
+    a table using the command line output_format to define the
+    table format.
 
-    Examples:
-
-      pywbemcli connection list
+    An "!" before the name indicates the selected connection.
+    A '!' before the name indicates that it is the current connection.
```
andy-maier: Multiple issues:
1. It is not clear what a "selected" connection is other than the new current connection. So I think there is no need to indicate a selected connection separately from the current connection.
2. In the second sentence (about current connection), "!" should be changed to "*".
3. You also introduced the concept of a default connection, in the code it is prefixed with "#" but it is missing in this description.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -286,27 +312,22 @@ def connection_test(context):
...
     Save the current connection as a new WBEM connection definition.
 
-    Create a new WBEM connection definition in the connections file from
-    the current connection. A connection definition with that name must not
-    yet exist.
-
-    This command is useful when you have defined a connection on the command
-    line and want to save it in the connections file for later use.
+    Create a new WBEM connection definition in the connections file from the
+    current connection with the NAME argument as connection name. A connection
```
andy-maier: On " with the NAME argument as connection name":
1. To be precise, it is the connection definition name.
2. Along with my suggestion on "connection add", I suggest to shorten the first sentence to: "Create a new WBEM connection definition named NAME in the connections file from the current connection."

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -286,27 +312,22 @@ def connection_test(context):
...
+    definition with that name must not yet exist.
```
andy-maier: Add "The NAME argument is required".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -125,27 +133,40 @@ def connection_select(context, name):
     connection. If the NAME argument is omitted, prompt for selecting one of
     the connection definitions in the connections file.
 
-    This command is useful in the interactive mode of pywbemcli.
+    Selects a connection from the persistently stored set of named connections
+    if NAME exists in the store. If NAME not supplied, a list of connections
+    from the connections definition file is presented with a prompt for the
+    user to select a connection.
```
andy-maier: The previous paragraph (starting on line 132) already describes this. The new paragraph here has no new information and can be removed.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -125,27 +133,40 @@ def connection_select(context, name):
...
 
-    The selection of the current connection is not saved across interactive
-    sessions of pywbemcli.
+    Selection is persistent; it is used as the server definition in commands
+    in the remainder of an interactive session and is saved in the
+    connections file and used as the current connection if there is no other
+    server definition (--server or --name or --mock-server).
```
andy-maier: This is confusing. Does the command set the current connection (as it previously was described to do)? Or the default connection? Or both?

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -125,27 +133,40 @@ def connection_select(context, name):
...
-      pywbemcli
+      $ pywbemcli
 
-      > connection select myconn
+      pywbemcli> connection select myconn
+      pywbemcli> connection show
+      pywbemcli> CTRL-D
+      $ pywbemcli show
+        TODO
+        ...
```
andy-maier: This does not format right. You need to put each line between blank lines. Given that, this is not the place to explain larger sequences of commands. I suggest to remove the show commands.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Change "if NAME argument exists" to "if the NAME argument is specified".

docs/pywbemclicommands.rst
KSchopmeyer: You are right. DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE.

By the way, at this point I am NOT doing any paragraph reformatting to make changes easily visible in the diff.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: Agreed and dropped next sentence

KSchopmeyer: DONE

KSchopmeyer: DROPPED. DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: It duplicates the same thing in add.  Note that I changed the name of that param in add to --input-name. For some reason, click did not like an argument and option with the same name.

In any case I included that option.
DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: Hmmm. That one was from old version.  Now reads

    An "!" before the name indicates the selected connection.
    A '*' before the name indicates that it is the current connection.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: That should be the default connection.  It is different than the current connection.  It is the one we used connection select to define as the default.

That should be 

    An "!" before the name indicates the default connection. See connection select.
    A '*' before the name indicates that it is the current connection.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: I shortened your proposal by leaving off the "in the connections file from the current connection. Note that you proposed above to put the --name option in here also to get from existing so it was more than just current.  However, the real reason is that we have a real limit on that first sentence so it shows up as a complete sentence on a single line in the list of Commands:

KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: What was missing is that this really does two things.  Sets current and also sets default in the connections file.  That is where some of the complexity comes in.

First paragraph now talks about the NAME and what happes if you do not provide it.

Paragraph 2 about setting it as default and also as current.

I actually do not like this combination of 2 things (one persistent and one transient) done at the same time. I would rather that we could separate so it either set NAME to be the default in the store or set NAME to be the current.  SUGGESTIONS WELCOME.  How about an option that would set the persistent default.

**DISCUSSION**  Or just tell me if this makes sense.
connection select abc     only sets it to current
connection select abc -d  sets the default-connection in the connections file.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: Both right now. See comment above.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE used \b to simplify

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: I went ahead and added an option --default that sets both default in the connections file and current. If --default not set, the name is only set to current and the persistent default not changed.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: I don't like the option name --input-name at all. In all other commands, the option name is --name for the same thing. We need to find another solution for the variable name collision.

I think the collision only exists between subcommand option --name and subcommand argument NAME, but not involving general option --name.

That is related to the question I stumbled across yesterday when reviewing this PR: Why does connection add have command options for specifying the connection when the same options already exist as general options? I think add could use the general options for what it does. The only thing we need to watch out for is the current connection, which we don't want to change, but I think when used in interactive mode, a connection add has its options in another context than the outermost context. We can try that out to make sure.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: The response does not address the comment.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: **Discussion point**:

The introduction of a default marker that represents the default connection has all kinds of ugly consequences: Hard to distinguish from current connection, inconsistent behavior (we can now set it but never get rid of it).

How about representing the default connection definition simply as the connection definition named "default"? That allows doing everything with it.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: See discussion point.

andy-maier:
The CI runs fail during flake8 with:
```
pywbemtools/pywbemcli/_cmd_connection.py:587:1: E303 too many blank lines (3)
pywbemtools/pywbemcli/_cmd_connection.py:588:1: E302 expected 2 blank lines, found 3
```

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -251,21 +272,25 @@ def connection_select(context, name):
                      default='all'))
 @add_options(verify_option)
 @click.pass_obj
```
andy-maier: **Discussion point:**

As part of the discussion on --name on the save command, I stumbled across this question:

Why does "connection add" have command options for specifying the connection when the same options already exist as general options? I think "add" could use the general options for what it does. The only thing we need to watch out for is the current connection (that only exists in interactive mode), which we don't want to change, but I think when used in interactive mode, a "connection add" has its options in another context than the outermost context. We can try that out to make sure. If we can achieve that the current connection is not affected, then I suggest to change "add" to use its input connection from the general options.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE. Added if it exists

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: **DISCUSSION**
That depends on whether you think the users will try to use the names as some sort of meaning.  Thus, I always set the simplest mock as the name mock1. The minute I use the concept of name changing to set the default, that sense is lost.  

I agree that there is an issue in that you cannot remove the default  connection definition completely without removing the connection in the current code.  However that is easily changed. Further, the use of the default is clearly a specific behavior. The user must start pywbemcli with NO connection definition at all before the default is enabled.  NOTE: This is a good place for a mesage in verbose mode.  "Connect xxxx (default) is current".

As far as hard to distinguish, that is easily resolvable; different symbol, col in table for these states, etc.

I really think this scheme works but have no issue with a clearer way to express what is the default in the table output, etc.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: **DISCUSSION**

I agree that we two similar commands to get connection definitions into the connections file, 'add' and 'save'.  Originally, I did add so that it could modify an existing connection but then figured out that there were real issues here in that the general options do not really allow that.  For example, how could you modify a connection to set no_verify to False with the existing options( --no_verify sets it to True, but the only other options not to supply the option which means nothing more than option not defined). There is no way to separate option value is None vs option not supplied and a modify would depend on that difference to determine what variables to set.) 

We would have to modify many of the options in add so that the user had more detailed control over them ( ex. set --output-format to "None", set --no-verify to false) which would mean a real modify command not just an add that duplicates the general options.

So, why do we need the add at all?  You can use the general options in the interactive mode to define a server and just use 'connection save<name>'  to persist it with a name.

On startup, the general options defines what will be the current server and if the command 'save <name> saves that connection. 

In interactive mode, it could use the general options and 'save <name>'  command to create and save a connection but not make it current. Note that save DOES NOT do this today since we do not capture these general option inputs that define a connection in interactive mode.

But the add does create and persist a connection definition without affecting the current. The save saves the current.

The only command that  changes the current connection in interactive mode is the 'connection select' in that it sets the connection definition in the root click context.  NOTE: I just added a clarifying method to context_obj for that.  update_root_click_context

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Ok. The description of the command still mentions --input-name, though. Also, it only saves the current connection, not "a connection".

Change:

    Save a connection to a new WBEM connection definition named NAME.

    Create a new WBEM connection definition in the connections file from the
    current connection or the connection definition in --input-name anamed
    NAME.

To:

    Save the current connection to a new WBEM connection definition.

    Create a new WBEM connection definition named NAME in the connections
    file from the current connection .

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: I can buy your argument that connection definition names should have a meaning. However, I also believe mthat having a concept of a default connection complicates things more than needed.

What the (any form of) persisted default provides is the ability to omit `--name emc3` on the command line. When in interactive mode, you already have the current connection that serves as the default so there it is not even an advantage. In fact, it complicates things because one needs to understand whether the current connection or the persisted default connection has precedence.

My suggestion would be to drop the concept of a persisted default altogether.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Why is it that the "save" command cannot not process general options in interactive mode? It does process general options in command mode. Which is an inconsistency we should address anyway.

For example, today the use of general options on "save" in interactive mode is simply ignored, which is definitely unexpected for a user:

```
$ pywbemcli --server https://localhost
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli.
pywbemcli> -m tests/unit/simple_mock_model.mof connection save mock1
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------------+-------------+-----------+-------------+
| name   | server            | namespace   |   timeout | no-verify   |
|--------+-------------------+-------------+-----------+-------------|
| *mock1 | https://localhost | root/cimv2  |        30 | False       |
+--------+-------------------+-------------+-----------+-------------+
pywbemcli>          
```

While in command mode:
```
$ pywbemcli -m tests/unit/simple_mock_model.mof connection save mock1
$ pywbemcli connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                      |
|--------+-------------+-----------+-------------+----------------------------------|
| mock1  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof |
+--------+-------------+-----------+-------------+----------------------------------+
```
and of course it rejects using --server and -m at the same time.

andy-maier: One more thought: If we allow that "save" can overwrite existing connection definitions, it (1) is what one expects from a save, and (2) it provides the modify functionality.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Also, the fact that "save" sets the current connection is weird. There is also an issue with it. The following causes two current connections to be displayed:
```
$ pywbemcli                           
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli.
pywbemcli> connection add -m tests/unit/simple_assoc_mock_model.mof mock1
pywbemcli> connection add -m tests/unit/simple_mock_model.mof mock2
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| mock1  | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> connection select mock1
"mock1" current
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| *mock1 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> connection save mock3
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| *mock1 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| *mock3 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> 
```

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Also, the fact that "save" sets the current connection is weird. There is also an issue with it. The following causes two current connections to be displayed:
```
$ pywbemcli                           
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli.
pywbemcli> connection add -m tests/unit/simple_assoc_mock_model.mof mock1
pywbemcli> connection add -m tests/unit/simple_mock_model.mof mock2
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| mock1  | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> connection select mock1
"mock1" current
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| *mock1 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> connection save mock3
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+--------+-------------+-----------+-------------+----------------------------------------+
| name   | namespace   |   timeout | no-verify   | mock-server                            |
|--------+-------------+-----------+-------------+----------------------------------------|
| *mock1 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| *mock3 | root/cimv2  |        30 | False       | tests/unit/simple_assoc_mock_model.mof |
| mock2  | root/cimv2  |        30 | False       | tests/unit/simple_mock_model.mof       |
+--------+-------------+-----------+-------------+----------------------------------------+
pywbemcli> 
```

KSchopmeyer: Not wierd, that is what happens if you don't copy before modifying.  
FIXED.  Also does not touch the current connection now.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: TODO

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: Conclusion is to have '*' for current connection and '#' for default connection.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: We agreed it is worth it. The precedence difficulty is handled by pywbemcli outputting the name of the connection that is used if it is invoked without one.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: We agreed to try implementing the following behavior:
- in interactive mode, commands can get global options but they apply only to that one command and not to future commands.
- in interactive mode, global options specified on a command overrwrite the inherited global options from the command invocation. Conflicting options in the reslting effective set of global options are rejected.
- connection save allows saving to an already existing connection name, overwriting it without warning.
- we drop connection add.

KSchopmeyer:
NOTE: I did not remove the verify on add and save. But the default is no verify so only the user can set it to use the verify.  Why not leave it.

I did not turn off the test for connection exists.  I will do it be make a separate pr.

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: I propose to make the overwrite on save and the question of verify separate prs.

KSchopmeyer:
The windows failures are because the OpenSSL v1.1.0 version changed again.
We are asking for Win32OpenSSL-1_1_0k.exe but they moved to L, I assume last night.  That is really a very fragil link. Its like we need to start with some future version and test backwards until we find one that works.

 Written by Mike Haertel and others, see <http://git.sv.gnu.org/cgit/grep.git/tree/AUTHORS>.
pywbem_os_setup.bat: Grep is already installed ... skipping
pywbem_os_setup.bat: Installing Win32OpenSSL ...
curl -o Win32OpenSSL-1_1_0k.exe -sSL https://slproweb.com/download/Win32OpenSSL-1_1_0k.exe --retry 3 --retry-connrefused --retry-delay 10
grep ".html." Win32OpenSSL-1_1_0k.exe
<html>
</html>

Note that this is really a change to pywbem, not pywbem tools

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: These are all in separate PRs, as shown above.

andy-maier:
On leaving --verify in: I have created a separate issue  #339 for that, so we can discuss it there.

PR #310, opened 2019-09-03
Ks/#250 replace short options

KSchopmeyer:
Changed a number of single character short option names to multicharacter shorter names. Changed the corresponding tests and added tests to be sure that the new short names are covered.

The list of options changes:

1. -p to  --pl for property list
2. --nq   --no-qualifiers for class
3. -R to  --rc result-class
4. -a to  --ac assoc-class
5. -C to  --rc result-class
6. -R to  --rr result-role
7. -q to  --iq include-qualifiers
8. -p to  --pl propertylis
9. -o to  --no for names-only
10. -c to --ico include-class-origin
11. -S to -s summary

The only issues I found in early testing is that it is easy to drop the first - and the one case where this causes problems is --no which becomes -n but fails because there is no namespace to not really a major issue.

### DIFF
--- a/docs/pywbemclicmdshelp.rst
+++ b/docs/pywbemclicmdshelp.rst
@@ -231,21 +231,22 @@ The following defines the help output for the `pywbemcli class associators --hel
         pywbemcli -n myconn class associators CIM_Foo -n interop
 
     Options:
-      -a, --assoc-class CLASSNAME     Filter the result set by association class
+      --ac, --assoc-class CLASSNAME   Filter the result set by association class
                                       name. Subclasses of the specified class also
                                       match.
-      -C, --result-class CLASSNAME    Filter the result set by result class name.
+      --rc, --result-class CLASSNAME  Filter the result set by result class name.
                                       Subclasses of the specified class also
                                       match.
       -r, --role PROPERTYNAME         Filter the result set by source end role
                                       name.
-      -R, --result-role PROPERTYNAME  Filter the result set by far end role name.
-      --no-qualifiers                 Do not include qualifiers in the returned
+      --rr, --result-role PROPERTYNAME
+                                      Filter the result set by far end role name.
+      --nq, --no-qualifiers           Do not include qualifiers in the returned
                                       class(es). Default: Include qualifiers.
-      -c, --include-classorigin       Include class origin information in the
+      --ico, --include-classorigin    Include class origin information in the
                                       returned class(es). Default: Do not include
                                       class origin information.
-      -p, --propertylist PROPERTYLIST
+      --pl, --propertylist PROPERTYLIST
                                       Filter the properties included in the
                                       returned object(s). Multiple properties may
                                       be specified with either a comma-separated
@@ -254,12 +255,12 @@ The following defines the help output for the `pywbemcli class associators --hel
                                       not in the object(s) will be ignored. The
                                       empty string will include no properties.
                                       Default: Do not filter properties.
-      -o, --names-only                Retrieve only the object paths (names).
+      --no, --names-only              Retrieve only the object paths (names).
                                       Default: Retrieve the complete objects
                                       including object paths.
       -n, --namespace NAMESPACE       Namespace to use for this command, instead
                                       of the default namespace of the connection.
-      -S, --summary                   Show only a summary (count) of the objects.
+      -s, --summary                   Show only a summary (count) of the objects.
       -h, --help                      Show this message and exit.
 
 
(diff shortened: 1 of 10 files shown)


andy-maier:
Comments:

* The following short option names should also become short long option names:
  - class enumerate CIM_Foo -n interop

Options:
  -d, --deep-inheritance        Include the complete subclass hierarchy of the
                                requested classes in the result set. Default:
                                Do not include subclasses.
  -l, --local-only

KSchopmeyer:
Updated with requested changes.  Made changes to docs/pywbemclicommands.rst.  i also made other changes to pywbemclicommands.rst behond just hte change from -d to -dk, -n to -no in that I 

1 instance create section and added more examples

2.  Changed the section that talked about CIM instance and instance names to cinclued classes since this affects classes.

3. Moved your paragraph on using backslash escap in shells to a footnote so it could be referenced multiple places. NOTE; The footnoe shouws up as section 4.9   since I have not figued out how to put in section header that you do not number.

docs/pywbemclicommands.rst
```diff
@@ -535,42 +529,69 @@ The **instance** group defines commands that act on CIM instances including:
...
   properties are defined as name/value pairs, one property for each instance of
-  the ``--property`` option. Since the WBEM server (and pywbem) requires that
-  each property be typed, pywbemtools uses the CIMClass defined by CLASSNAME
-  retrieved from the WBEM server to define the type required to define the
-  CIMProperty.
+  the ``--property`` option.
 
-  For a single property in the new instance this is simply the `--property`` option
-  with the property name and value:
+  The ``--property`` argument value component defines the value of the
```
andy-maier: What is a "argument value component"? Suggest to change to simply "option". It is clear from the context that it is actually the argument of the option that has the property value.

docs/pywbemclicommands.rst
```diff
@@ -535,42 +529,69 @@ The **instance** group defines commands that act on CIM instances including:
...
+  property. Since the WBEM server (and pywbem) requires that each property be
+  typed, pywbemcli uses the CIMClass defined by CLASSNAME retrieved from the
```
andy-maier: On "pywbemcli uses the CIMClass defined by CLASSNAME retrieved from the WBEM server to define the type required to define the CIMProperty.":
Suggest to simplify to: 
"pywbemcli retrieves the CIM class to determine the property types."

docs/pywbemclicommands.rst
```diff
@@ -535,42 +529,69 @@ The **instance** group defines commands that act on CIM instances including:
...
+  WBEM server to define the type required to define the CIMProperty. The value
+  argument contains just the value itself in the form of a string or numeric
```
andy-maier: What is "The value argument"?
Suggest to change to "The option argument".

docs/pywbemclicommands.rst
```diff
@@ -535,42 +529,69 @@ The **instance** group defines commands that act on CIM instances including:
...
+  value. For numeric values, the creation will fail if the values of the
+  numeric exceeds the range of the CIM type for the property defined in the
+  class (ex. -3 for Uint32).
+
+  See [#fbackslash]_ for information on use of backslashes in formating
+  property argument values:
```
andy-maier: Backslash-escaping the double quotes around string or datetime property values is not needed. The shell will make sure the (unescaped) double quoted string is passed as a single value to the option's variable.

docs/pywbemclicommands.rst
```diff
@@ -535,42 +529,69 @@ The **instance** group defines commands that act on CIM instances including:
...
+  For a single property in the new instance this is simply the `--property``
+  option with the property name and value:
 
   .. code-block:: text
 
-    --property <property-name>=<property-value"
+    --property <property-name>=<property-value
+
+    where quotes around the value are only required if the value includes
+    whitespace.
```
andy-maier: I think instead of giving an (incomplete) description here, we should have a section "Specifying CIM property and parameter values" that is similar to "Specifying INSTANCENAME" and that defines the format precisely, e.g. by referring to DSP0004 stringValue etc.

docs/pywbemclicmdshelp.rst
```diff
@@ -1948,11 +1950,13 @@ The following defines the help output for the `pywbemcli server get-centralinsts
...
                                       only scopig methodology
-      -s, --scoping-class CLASSNAME   Optional. Required only if profiles supports
+      --sc, --scoping-class CLASSNAME
+                                      Optional. Required only if profiles supports
                                       only scopig methodology
-      -S, --scoping-path CLASSLIST    Optional. Required only if profiles supports
+      --sp, --scoping-path CLASSLIST  Optional. Required only if profiles supports
                                       only scopig methodology. Multiples allowed
       -r, --reference-direction [snia|dmtf]
                                       Navigation direction for association.
```
andy-maier: docs/pywbemclicmdshelp.rst is out of date, which is th ereason it shows `-r` on `--reference-direction`.

docs/pywbemclicommands.rst

docs/pywbemclicommands.rst

docs/pywbemclicommands.rst

docs/pywbemclicommands.rst
KSchopmeyer: Good idea. I was thinking about it also.  I added the section and referenced it from both the class and instance invokemethod definitions. NOTE: I put a TODO on this one since the first goal was to replace the short options and I need to do more cleanup on the text in this section.  We have this pr, the sections naming pr and your pywbmbemclicommands.rst pr that all get merged and then we can look at this chapter in more detail

docs/pywbemclicmdshelp.rst
KSchopmeyer: DONE

docs/pywbemclicommands.rst
KSchopmeyer: DONE with 2 TODOs in the documentation to refactor the text.

PR #312, opened 2019-09-03
Fixed help file dependency in make build

andy-maier:
Ready for review and merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -562,7 +562,7 @@ endif
 # regenerate MANIFEST. Otherwise, changes in MANIFEST.in will not be used.
 # Note: Deleting build is a safeguard against picking up partial build products
 # which can lead to incorrect hashbangs in the pywbem scripts in wheel archives.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -562,7 +562,7 @@ endif
 # regenerate MANIFEST. Otherwise, changes in MANIFEST.in will not be used.
 # Note: Deleting build is a safeguard against picking up partial build products
 # which can lead to incorrect hashbangs in the pywbem scripts in wheel archives.
-$(bdist_file) $(sdist_file): _check_version setup.py MANIFEST.in $(dist_dependent_files) $(moftab_files)
+$(bdist_file) $(sdist_file): _check_version setup.py MANIFEST.in $(doc_utility_help_files) $(dist_dependent_files) $(moftab_files)
```
KSchopmeyer: Why does moftab_files show here.  Should we have to make pywbem depencies part of pywbemtools dependencies???

andy-maier: Good catch, and I agree that we should not.
Turns out the variable was not set, so make expands it then to nothing.
I removed it.

PR #320, opened 2019-09-04
Fixes #297: Updated README and README_PYPI

### LINKED ISSUE #297: Update of README and README_Pypi
These readme files need to be updated to match the latest description of pywbemtools from the Introduction chapter of the docs.

andy-maier:
Ready for review and merge.

Note: The issue with missing pywbem_os_setup.sh in the installation section of the docs will need to be fixed separately.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,241 +28,298 @@ pywbemtools: Python tools for communicating with WBEM servers
 .. image:: https://img.shields.io/pypi/pyversions/pywbemtools.svg?color=brightgreen
     :alt: PyPI - Python Version
 
-.. # .. contents:: **Contents:**
-.. #    :local:
 
 Overview
 --------
 
 Pywbemtools is a collection of command line tools that communicate with WBEM
-Servers. The tools are written in pure Python and support Python 2 and Python
+servers. The tools are written in pure Python and support Python 2 and Python
 3.
 
-Pywbemtools provides a command line tool (pywbemcli) that uses the pywbem
-Python WBEM client infrastructure to issue operations to a WBEM server using
-the CIM/WBEM standards defined by the DMTF to perform system management tasks.
+At this point, pywbemtools includes a single command line tool named
+``pywbemcli`` that uses the `pywbem package on Pypi`_ to issue operations to a
+WBEM server using the `CIM/WBEM standards`_ defined by the `DMTF`_ to perform
+system management tasks.
 
-CIM/WBEM infrastructure is used for a wide variety of systems management tasks
-in the industry including DMTF management standards and the SNIA Storage
-Management Initiative Specification(SMI-S).
+CIM/WBEM standards are used for a wide variety of systems management tasks
+in the industry including DMTF management standards and the `SNIA`_
+Storage Management Initiative Specification (`SMI-S`_).
 
-Pywbemcli provides the functionality to :
+Pywbemcli provides access to WBEM servers from the command line.
+It provides functionality to:
 
-* Explore the CIM data of WBEM Servers. It can manage/inspect the CIM model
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -28,241 +28,273 @@ pywbemtools: Python tools for communicating with WBEM servers
...
 
-.. _Installation section: https://pywbemtools.readthedocs.io/en/stable/introduction.html#installation
+The quick way for Linux and OS-X:
 
-Documentation
--------------
+.. code-block:: bash
 
-The latest pywbemtools documentation is available on ReadTheDocs:
+    $ wget -q https://raw.githubusercontent.com/pywbem/pywbem/master/pywbem_os_setup.sh
```
KSchopmeyer: You don't have to get it.  For now we use our own shell cmd and wget the .bat file from pywbem

README_PYPI.rst
```diff
@@ -1,50 +1,92 @@
...
+Requirements:
+
+1. Python 2.7, 3.4 and higher
+
+2. Operating Systems: Linux, OS-X, native Windows, UNIX-like environments on
+   Windows (e.g. Cygwin)
+
+The quick way for Linux and OS-X:
+
+.. code-block:: bash
```
KSchopmeyer: Again, we the .sh is ours because of cygwin and we wget the .bat

README.rst
andy-maier: If a user installs from Pypi, they don't have the github repo checked out. So they need to get the script, just not from pywbem but from pywbemtools.
DONE - Changed to get from pywbemtools.

README_PYPI.rst
andy-maier: See previous comment.

README.rst
KSchopmeyer: I was really trying to get this down to just: 

   $ pip install pywbemtools

Forgot we were doing our installs from the git repo in our tests.

Since it turns out most users at this level are operating from Windows, we need to add the window alternative (at least the native windows but not about cygwin) to this also, which, of course is a mess because of wget.  I assume this means:

'''

install Chocolatey
$ choco install wget
$ wget -q https://raw.githubusercontent.com/pywbem/pywbemtools/master/pywbem_os_setup.bat
$ pywbem_os_setup.sh
pip install pywbemtools

As an alternative could we put the scripts into PyPi?????

andy-maier: The Windows alternative is already described in the installation chapter of the docs. I have made an update to this PR that includes installation on Windows (but wget is stated as a prerequisite).

We cannot put the scripts into the Pypi package because I did not find a way to let them be executed upon install. We can of course just have them there as a script, but that breaks the pip install into pip download, execute the script, pip install, which is not less ugly, particularly since pip download only exists since pip version soandso.

KSchopmeyer: I agree with you comments on the scripts but I makes what should be simple for the user, in particular windows, messy.

Darn.  Then I only have one more comment.  It comes down to "hey windows user, you figure out how to install wget. Not our problem.". The choices are a) leave it to them, add the install of Chocolatey which again leads them off to clipping and pasting a script, 3) a sentence about where they can get wget (install with chocolatey, https://eternallybored.org/misc/wget/, or possibly other sites.

I will go with any of them but from what I have seen, the first users will want to use pywbemtools, not fight installing it.

andy-maier: I can make the following changes to address your comments, but I would rather do them in a separate PR because they affect not just the READMEs but also the install section in the docs:
* On wget: I can add the one line to install wget using choco.
* On need for pywbem_os_setup scripts: For just installation, they are needed on all OS platforms but only with Python 2. I can update the install description accordingly.
  Note: For development, they are needed on all OS platforms with all Python versions, but that is covered in the how to develop section.

PR #325, opened 2019-09-05
Adds a new file pywbemclifeatures.rst

KSchopmeyer:
Moves the whole features section from 4.1 to this section.  I put in
index between chapters 3 and 4

This is an experiment. Converted the footnot on backslash to a term and put in appendix.


andy-maier:
Good idea, I think this is the right direction.

PR #327, opened 2019-09-08
Ks/add class find sort

KSchopmeyer:
Adds the option sort to class find so the results can be sorted by classname or namespace. cannot really test with mock because we only have one namespace.


KSchopmeyer:
Delete committed in error

PR #328, opened 2019-09-08
Add --sort option to class find.

KSchopmeyer:
Allows sorting results by classname or namespace.


andy-maier:
I assume this one was ready to review. Approved.

PR #333, opened 2019-09-10
Fixes #330: Simplify installation on Python 3 re scripts

### LINKED ISSUE #330: Simplify install for Python 3 users re pywbem_os_setup scripts
It turns out that the pywbem_os_setup scripts are needed only with Python 2 when installing pywbemtools.

Update the install description accordingly, in the installation section and the README* files.

Note: For development of pywbemtools, the scripts are needed on all OS platforms with all Python versions, and that is already covered in the development section.

andy-maier:
READY FOR REVIEW AND MERGE.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ Requirements:
 2. Operating Systems: Linux, OS-X, native Windows, UNIX-like environments on
    Windows (e.g. Cygwin)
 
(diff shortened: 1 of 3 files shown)


andy-maier:
Change needed on the prereqs: wget and choco are only needed on Python 2.

Issue #300, opened 2019-08-31
Final review of docs and help texts for fixes and editorial changes


andy-maier:
Done:
* #351 - chapters 1 to 4
* #356 - chapter 5 class commands; some terms in appendix
* #361 - appendix (some terms)
* #359 - chapter 5 instance & qualifier commands
* #362 - chapter 5 server commands
* #364 - chapter 5 connection commands, help, repl
* chapter 7 (mock support) - nothing to do
* chapter 8 (development) - nothing to do
* #376 - chapter 9 (appendix)
* #377 - help texts of all commands

andy-maier:
Needs to wait for merge of all PRs that update docs and help texts.

andy-maier:
Will be addressed by PR #351.

PR #336, opened 2019-09-12
Fixes #335 #338: Modify so general opts work in interactive mode

### LINKED ISSUE #335: Modify so we use general options in interactive mode
Modify startup and command setup so that we use the general options for connections (--server, etc. in interactive mode as well as at startup.  This involves two changes.

i. Process all of the options to the cli method on all calls, not just the initial call.
2. Modify Pywbemcliserver class to do clean modification of parameters as well as init.
3. Modify the code in the cli function that processes commands in interactive mode so that it processes all input parameters to cli and sets them into the current context.

NOTE: The goal is to use general options only for the current command execution, not to make them the norm for succeeding commands in the interactive session.

KSchopmeyer:
Overview:

1. Save saves the current connection with modifications from the general options

```
    pywbemcli> --user fred connection save t1

    Sets user to fred and saves the connection as t1.

```
2. Connection save changed to allow overwrite of the connection if already in the connections file by removing the test for existing name in the file. Since the new general options are not retained after the connection save command, there is no impact on the current connection. # issue #338

3. Uses general options before all commands except connect [select]. Sets the general options values into the current connection but restores any previous connection definition for subsequent commands in the interactive session. Connection select, makes the defined option the current option for the defined interactive session but DOES not apply the modifications from the general options because they would have been applied to the current connection, not the connection we are selecting.  NOTE: we do not notify the user of this issue.

4. Connection select with general options gets the connection from the repo, makes the changes from the general options and also makes this the current option for this interactive session. There will be NO way to select the current connect and change it, only persistent connections.

```
  $  pywbemcli  -n myserver
  pywbemcli> --user blah --pasword blah connection save myserver

 Changes the user and password for the myserver connection.

```
5. When the general options are applied to a connection, the connection is reset so that the first command to the server reinitiates the connecton.

Issues:

Some options cannot be unset.  Thus, you can set value into --user but you cannot set it back to None because it is the existence of the general option that causes it to be set. This applies to --user, --password, --keyfile, etc.

**COMMENT" Since click does not support the concept of options with optional arguments, see click issue:  https://github.com/pallets/click/issues/549, we propose that setting the value of these options to "" is a valid way to let pywbemcli set the values to None.  I propose that this be a separate PR just to keep the changes to this PR limited.

Modified -no-verify per the commit message so it can be set, unset, etc. The new option is `--verify/--no-verify' and there is no short form


andy-maier:
Since PR #306 is now merged, please rebase this on master.

docs/appendix.rst
```diff
@@ -143,6 +143,22 @@ This documentation uses a few special terms to refer to Python types:
       backslash-escaping any double quotes in INSTANCENAME, and putting it into
       double quotes.
 
+   default connection
+      A connection definition in the :ref:`--server general option` that is used
```
andy-maier: I don't understand "in the :ref:`--server general option`" in this context.
Should this be changed to: "in the :ref:`connections file`" ?

docs/appendix.rst
```diff
@@ -143,6 +143,22 @@ This documentation uses a few special terms to refer to Python types:
...
+      by pywbemcli as the :term:`current connection` if pywbemcli is started without
+      any connection definition (no :ref:`--server general option`,
+      :ref:`--mock-server general option`, or :ref:`--name general option`).
+      A connection definition in the :term:`connections file` becomes the
+      current connection on pywbemcli startup if it is defined using the
+      :ref:`connection select command`.
```
andy-maier: I think this sentence does not belong to the description of the default connection, but to the description of the current connection.

docs/appendix.rst
```diff
@@ -143,6 +143,22 @@ This documentation uses a few special terms to refer to Python types:
...
+
+   current connection
+      The connection definition in pywbemcli that is currently active; the
+      target WBEM server for pywbemcli commands.  The current server is defined
```
andy-maier: On "the target WBEM server": It may also be a target mock server.
Can we change this to "the target connection"?

docs/appendix.rst
```diff
@@ -143,6 +143,22 @@ This documentation uses a few special terms to refer to Python types:
...
+      either upon startup (:ref:`--server general option`,
+      :ref:`--mock-server general option`, or :ref:`--name general option` or
+      :term:`default connection`) or with the :ref:`connection select command`.
```
andy-maier: It is **always** defined on startup: If none of these three general options is specified, the default connection is used. That needs to be clarified. The select command **changes** the current connection.

docs/pywbemclicmdshelp.rst
```diff
@@ -75,10 +75,11 @@ The following defines the help output for the `pywbemcli  --help` command
       -p, --password TEXT             Password for the WBEM server. Default:
                                       EnvVar PYWBEMCLI_PASSWORD, or prompted for
                                       if --user specified.
-      -N, --no-verify                 If true, client does not verify the X.509
+      --verify / --no-verify          If --verify, client verifies the X.509
                                       server certificate presented by the WBEM
-                                      server during TLS/SSL handshake. Default:
-                                      EnvVar PYWBEMCLI_NO_VERIFY, or false.
+                                      server during TLS/SSL handshake. If --no-
+                                      verify client bypasses verificationDefault:
```
andy-maier: Missing `. ` between "verification" and "Default".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -136,14 +136,15 @@ def connection_select(context, name, **options):
...
-    Default and current connction are set if the --default option exists;
-    otherwise only the connection in the current interactive session is set .
-    Once define the default connection will be used as the server definition in
+    Selects the connection named NAME from the persistently stored named
+    connectionsto be the current connection if NAME exists in the store . If
+    the NAME argument does not exist, a list of connections from the
+    connections definition file is presented with a prompt for the user to
+    select a connection.
+
+    Default and current connection are set if the --default option exists;
```
andy-maier: Change "--default option exists" to "--default option is used".

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -104,16 +106,16 @@
...
               # defaulted in code
-              envvar=PywbemServer.no_verify_envvar,
-              help='If true, client does not verify the X.509 server '
+              envvar=PywbemServer.verify_envvar,
+              help='If --verify, client verifies the X.509 server '
                    'certificate presented by the WBEM server during TLS/SSL '
-                   'handshake. '
-                   'Default: EnvVar {ev}, or false.'.
-                   format(ev=PywbemServer.no_verify_envvar))
+                   'handshake. If --no-verify client bypasses verification'
```
andy-maier: Add "`. `" (dot and space) at the end of this line, to separate next line.

docs/pywbemclicommands.rst
```diff
@@ -979,14 +979,7 @@ options for pywbemcli. The data includes:
   received from the WBEM server during connection setup. See :ref:`--no-verify general option`.
 * **certfile** - optional server certificate filename. See :ref:`--certfile general option`.
 * **keyfile** - optional client private keyfile filename. See :ref:`--keyfile general option`.
-* **use-pull** - optional parameter that defines whether pull operations are
-  to be required, used if they exist or not used. See :ref:`--use-pull general option`.
-* **pull-max-cnt** - optional count of object per pull operation.  See :ref:`--pull-max-cnt general option`.
```
andy-maier: I think use-pull and pull-max-cnt should be part of the data stored for a connection definition, because they represent how to interact with the WBEM server referenced by the connection definition.

docs/pywbemclicommands.rst
```diff
@@ -1111,13 +1105,19 @@ Connection select command
...
 The ``connection select`` command selects a connection from the connections
-file and marks that connection as the default connection. The default
-connection will be used if it is defined and pywbemcli is started with no
-:ref:`--server general option` or :ref:`--mock-server general option` general options. A connection may be
-selected to be default either by using the NAMEargument or if no argument is
-provided by selecting from a list presented on the console. The following
-example shows changing connection from within the interactive mode of
-pywbemcli:
+file sets that connection as the :ref:`current connection` and optionally the
+:ref:`default connection` (``--default option``).
```
andy-maier: Change "(``--default option``)" to "by using the ``--default`` option" (ideally with a link to the option section)

docs/pywbemclicommands.rst
```diff
@@ -1154,15 +1154,66 @@ pywbemcli:
...
 
 Connection show command
-^^^^^^^^^^^^^^^^^^^^^^^^^
-The ``connection show`` command shows information in the current connection.  See the the ``select``
-above for an example of this command.
+^^^^^^^^^^^^^^^^^^^^^^^
+The ``connection show`` command shows information about the current connection or the
+connection with name NAME.  The connection to be displayed will be
+the :ref:`current connection` if the NAME argument is omitted, the connection named
+with the NAME argument, or selected from a selection list the NAME argument "?"
```
andy-maier: We should not introduce multiple different ways to trigger the prompting. In other commands, the prompting is triggered by the use if the `--interactive` option, and I think that's what we should use also in this case.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -187,9 +188,15 @@ def connection_save(context, name, **options):
     """
     Save the current connection to a new WBEM connection definition.
 
-    Create a new WBEM connection definition in the connections file from the
-    current connection. A connection definition with the name NAME must not yet
-    exist. The NAME argument (the name of the new connection) is required.
+    Save WBEM connection definition in the connections file from the current
+    connection. The NAME argument (the name of the new connection) is required.
```
andy-maier: The first sentence of this paragraph could be streamlined, e.g. to "Save the current connection as a new connection definition in the connections file."

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -187,9 +188,15 @@ def connection_save(context, name, **options):
...
+    If the connection exists, it will be overwritten with the current
+    connection definition.
```
andy-maier: I think it would be good to be more clear as to what is considered an existing connection. I suggest to change "If the connection exists" to "If a connection definition with that name already exists".

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: Not quite.  There may be no default. In that case, pywbemcli is started with no connection defined.  In that case, commands like connection will work but commands like ;class enumerate` will fail because no server is defined.  In that case, the select would create a current connection rather than change an existing one.

Rewrote that one.  DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONEE

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: DONE

docs/pywbemclicommands.rst
KSchopmeyer: We agreed that they wern't the other day.  Since the default is to try pull and then go to the traditional, the only real reason for this option is special cases of testing by forcing a pull, etc. 
The reasons I see for doing this is for the user to test response time with pull vs no pull, etc.
 Even a server that does not sujpport pull will work OK with just the overhead of the pull tested first. 

Of course, on the other hand, if it becomes part of the connection itself, it means that the extra pull request can be completely turned off for a particular server.

In reality, this one should probably exist in both camps.  Apply to a particular server but override with in general but I am not sure that I can really do that.  

At the same time, that then brings up the question of the pull_max_cnt. It is part of the server or part of the general ops.

**DISCUSSION**

docs/pywbemclicommands.rst

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE except for word new since it can overwrite an existing connection

DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

docs/pywbemclicommands.rst
KSchopmeyer: This clearly show 3 options. Current, defined name or showme.  With the interactive option, you
have an inherent conflict since the name and the -i option can both occur so you have to define what overrides what. 

Actually  I would like to get the interactive option with the instance get/references/associators commands (i.e. try to get with just what is presented (the class) and when that fails do the selection) but you can't because dmtf combined class and instance associators/references that the operation with just the class is valid. 

**DISCUSSION**

docs/appendix.rst
andy-maier: DONE

docs/appendix.rst
andy-maier: DONE

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: I don't see this as done.

docs/pywbemclicommands.rst
andy-maier: I am ok with not making it part of the connection definition. CLOSED.

docs/pywbemclicommands.rst
andy-maier: In the existing cases of --interactive we also specify constraints on the argument: These are 6 cases that are all on instance operations that operate on a single instance, and they all have an INSTANCENAME argument, which is required to be the class name instead of an instance name if the option is used. In the case here, we can require that NAME is omitted if --interactive is used.

andy-maier:
With this code level, I get:
```
$ pywbemcli class --help
  File "/Users/maiera/virtualenvs/pywbem27/bin/pywbemcli", line 10, in <module>
    sys.exit(cli())
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/core.py", line 1134, in invoke
    Command.invoke(self, ctx)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/maiera/PycharmProjects/pywbem/pywbemtools/pywbemtools/pywbemcli/pywbemcli.py", line 410, in cli
    connections = ConnectionRepository()
  File "/Users/maiera/PycharmProjects/pywbem/pywbemtools/pywbemtools/pywbemcli/_connection_repository.py", line 69, in __init__
    self._read_connections_file()
  File "/Users/maiera/PycharmProjects/pywbem/pywbemtools/pywbemtools/pywbemcli/_connection_repository.py", line 174, in _read_connections_file
    replace_underscores=True, **svr)
  File "/Users/maiera/PycharmProjects/pywbem/pywbemtools/pywbemtools/pywbemcli/_pywbem_server.py", line 392, in create
    return PywbemServer(**kwargsout)
TypeError: __init__() got an unexpected keyword argument 'pull_max_cnt'
```

andy-maier:
Issue in docs/pywbemcligeneraloptions.rst: It contains the following reference: ":ref:`--output-format`" which should be changed to ":ref:`--output-format general option`".

docs/appendix.rst
```diff
@@ -143,6 +143,25 @@ This documentation uses a few special terms to refer to Python types:
...
       double quotes.
 
+   default connection
+      A connection definition in the :term:`connections file` that is used
+      by pywbemcli as the :term:`current connection` if pywbemcli is started without
+      any connection definition (no :ref:`--server general option`,
+      :ref:`--mock-server general option`, or :ref:`--name general option`) and
+      the default connection is defined.
+      A connection definition in the :term:`connections file` becomes the
+      default connection on pywbemcli startup if it is defined using the
```
andy-maier: Change "defined" to "specified".

docs/appendix.rst
```diff
@@ -143,6 +143,25 @@ This documentation uses a few special terms to refer to Python types:
...
+      :ref:`connection select command` and the ``--default option`` is used.
+
+   current connection
+      The connection definition in pywbemcli that is currently active; it is the
+      target connection for pywbemcli commands.  The current
+      server is created on pywbemcli startup with the following options
```
andy-maier: Change "current server" to "current connection".

docs/appendix.rst
```diff
@@ -143,6 +143,25 @@ This documentation uses a few special terms to refer to Python types:
...
+      :ref:`--server general option`,
+      :ref:`--mock-server general option`, or :ref:`--name general option` or
+      if a :term:`default connection`) has been defined. The current sconnection can be
```
andy-maier: Typo "sconnection".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -136,14 +136,15 @@ def connection_select(context, name, **options):
     """
     Select a WBEM connection definition as current or default.
 
-    Selects a connection from the persistently stored named connections if NAME
-    exists in the store to be the current connection. If the NAME argument does
-    not exist, a list of connections from the connections definition file is
-    presented with a prompt for the user to select a connection.
+    Selects the connection named NAME from the persistently stored named
+    connectionsto be the current connection if NAME exists in the store . If
```
andy-maier: Insert space in "connectionsto"

andy-maier: Remove "if NAME exists in the store". If you want to keep it, change it to "if it exists in the connections file", but I think we should describe the normal case here. The exceptional case is described in the next sentence.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -136,14 +136,15 @@ def connection_select(context, name, **options):
...
+    the NAME argument does not exist, a list of connections from the
```
andy-maier: Change "the NAME argument does not exist" to "the NAME argument is not specified".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -136,14 +136,15 @@ def connection_select(context, name, **options):
...
+    connections definition file is presented with a prompt for the user to
```
andy-maier: Shorten "connections definition file" to "connections file".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -136,14 +136,15 @@ def connection_select(context, name, **options):
...
+    select a connection.
```
andy-maier: Change "select a connection" to "select a connection definition".

docs/pywbemcligeneraloptions.rst
```diff
@@ -74,6 +74,13 @@ interactive mode and as a table:
 command, general the options redefinition's are not retained between command
 executions.
 
+The general options may be used in the interactive mode to override
+the default general options or specific general options for the current command.
+
+Thus, behavior of pywbemcli (ex. :ref:`--output-format`) or the parameters of the
```
andy-maier: Change invalid reference ":ref:`--output-format`" to ":ref:`--output-format general option`".

docs/appendix.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

docs/pywbemcligeneraloptions.rst
KSchopmeyer: DONE

docs/appendix.rst
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: DONE

docs/pywbemclicommands.rst
andy-maier: I opened issue https://github.com/pywbem/pywbemtools/issues/349 for this, so we can move ahead with this PR for now.

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: we can fix this later as part of issue #300.

andy-maier:
Approved

PR #341, opened 2019-09-13
Fixes #316: Docs: Fixed examples in section 5 and made more consistent

### LINKED ISSUE #316: Issues with the use of --mock-server and --name options in command examples
Section 4 shows command examples. These examples have these issues:
* Use of `--mock-server` option with arguments that are not files. Suggest to use files that exist in pywbemcli.
* Use of `--name` option without explaining that this refers to a named connection definition, or how the connection definition is established. Suggest not to use `--name` in examples that can also use `--mock-server` instead.

andy-maier:
Ready for  review and merge.


andy-maier:
If I think about it, the long names are more expressive to a first time reader, so I'll change that back.

Issue #339, opened 2019-09-13
Remove the --verify option on "connection save"


andy-maier:
For a saved connection definition, prompting for confirmation seems a bit over-protective. It is expected that users will list the connections anyway so they recognize any errors, and when save can overwrite existing connections (see issue #338) , any errors can be corrected easily.

andy-maier:
**Discussion point:** Should we leave the --verify option in, since by default it does not prompt.

My view on that is that a verification is not worthwhile on this command, and I suggest that we clean it up nevertheless. But let's discuss.

andy-maier:
Addressed by PR #352.

PR #352, opened 2019-09-16
Fix issue #339 Remove verify from connection save and connection delete

KSchopmeyer:
Removed the option completely from connection save and connection
delete commands.

Also modified the tests that test this option to just execute the save
and modified the help test for these commands.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_connection.py
+++ b/pywbemtools/pywbemcli/_cmd_connection.py
@@ -28,8 +28,7 @@ from pywbem import Error
 
 from .pywbemcli import cli
 from ._common import CMD_OPTS_TXT, pick_one_from_list, format_table, \
-    verify_operation, hide_empty_columns
-from ._common_options import add_options, verify_option
+    hide_empty_columns
 from ._pywbem_server import PywbemServer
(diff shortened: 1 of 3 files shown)


pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -483,20 +480,11 @@ def cmd_connection_delete(context, name, options):
 
     name = select_connection(name, context, connections)
 
-    context.spinner.stop()
```
andy-maier: Is `context.spinner.stop()` not needed anymore?

andy-maier: Are you sure that `context.spinner.stop()` can be removed? In connection save, you kept it.

KSchopmeyer: Yes because just above  the get_current_connection stops the spinner. The save does not call  select_ connection so I stop the spinner before the method ends.

KSchopmeyer: Put a comment in just to avoid confusion.

Also ran makebuild again and put update in new push

Issue #363, opened 2019-09-18
"server get-centralinsts" lists instance names as one string separated by colon


andy-maier:
Not sure whether the example in the description of the "server get-centralinsts" command in chapter 5 is correct, but if it is then the issue is that the instance names of the central instances are one large string where the instance names are separated by colon.

This should be changed to separate them by blank, and to break them into multiple lines of the row so that each loine shows one instance name (like it is done on other occasions where a table row can have multiple lines).

Here is the example from the docs which shows two instance names:

```
$ pywbemcli server centralinsts --org DMTF --profile "Computer System"
Advertised Central Instances:
+---------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Profile                         | Central Instances                                                                                                                                                                                                                       |
|---------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| DMTF:Computer System:1.0.0      | //leonard/test/TestProvider:Test_StorageSystem.Name="StorageSystemInstance1",CreationClassName="Test_StorageSystem"://leonard/test/TestProvider:Test_StorageSystem.Name="StorageSystemInstance2",CreationClassName="Test_StorageSystem" |
+---------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

KSchopmeyer:
Corrected this pr #368

PR #365, opened 2019-09-19
Change 'subcmd' to 'command' and command group in tests and code

KSchopmeyer:
We have decided on the names "command" and "command group" to define the
components of the pywbemcli cmd line.  However they were all named
subcmds in the tests, test methods, test definition tools. Changed all of that to command and command_group just to be consistent.

The word subcmd should not occur any more.

This just changes the text and the key names in the test dictionaries. 

It does not change code.

Also changed the test file names from test_*_subcmd.py to test_*_cmd.py

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_server.py
+++ b/pywbemtools/pywbemcli/_cmd_server.py
@@ -375,7 +375,7 @@ def cmd_server_centralinsts(context, options):
                     scoping_class=options['scoping_class'],
                     scoping_path=options['scoping_path'],
                     reference_direction=options['reference_direction'])
-                row.append("/n".join([str(p) for p in ci]))
+                row.append(":".join([str(p) for p in ci]))
             # mark current inst as failed and continue
             except Exception as ex:  # pylint: disable=broad-except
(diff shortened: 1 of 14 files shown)


pywbemtools/pywbemcli/_cmd_server.py
```diff
@@ -375,7 +375,7 @@ def cmd_server_centralinsts(context, options):
                     scoping_class=options['scoping_class'],
                     scoping_path=options['scoping_path'],
                     reference_direction=options['reference_direction'])
-                row.append(":".join([str(p) for p in ci]))
+                row.append("/n".join([str(p) for p in ci]))
```
andy-maier: This change fixes issue #363. Either say that in commit message title and description, or move to a separate PR.

KSchopmeyer: My BooBoo.  It was not supposed to.  I accidently leaked one pr into another. I am taking this one out and making it separate.

KSchopmeyer: REMOVED.

PR #369, opened 2019-09-20
Fixes issue # 360 server connection vs connection show

KSchopmeyer:
**This is one proposal to just remove this command** since it is a duplicate
of the connection show command.

**DISCUSSION**  Should we just remove the command for this release.

This is just a proposal at this point.  Some day we will have a command
that wants to show a lot of information about the server but that will
be more than just the connection information.

Suggestions welcome


andy-maier:
I suggest to remove the command for now, and have approved this PR.

PR #370, opened 2019-09-20
Fixes #357: Docs: Added error handling section

### LINKED ISSUE #357: Documentation defining good vs error responses
I made a mistake in putting statements in specific command about good vs. exception responses.  Things if successful returns ... or exception.  I think we should make the statement once about the commands that if unsuccessful they return exceptions received from the server or from pywbem interpretation of the response data and let it go at that.  Then remove anything from individual commands about unsuccesful responses except in those cases that differ from this general policy.

andy-maier:
Ready for review and merge.

### DIFF
--- a/docs/pywbemclicmdlineinterface.rst
+++ b/docs/pywbemclicmdlineinterface.rst
@@ -269,3 +269,67 @@ following examples, an underscore ``_`` is shown as the cursor:
 The pywbemcli shell supports history across multiple invocations of the shell
 using <UP-ARROW>, <DOWN-ARROW>.
 The pywbemcli history is stored in ``~/.pywbemcli_history``.
+
+
+.. _`Error handling`:
+
+Error handling
+--------------
+
+Pywbemcli terminates with one of the following program exit codes:
+
+* **0 - Success**: The pywbemcli command has succeeded.
+
+* **1 - Error**: In such cases, pywbemcli aborts the requested operation and
+  displays one or more human readable error messages on standard error.
+
+  If this happens for a command entered in interactive mode, the pywbemcli shell
+  is not terminated; only the command that failed is terminated.
+
+  Examples for errors reported that way:
+
+  * Local system issues, e.g. pywbemcli history file or connections file cannot
+    be written to.
+
+  * WBEM server access issues, e.g. pywbemcli cannot connect to or authenticate
+    with the WBEM server. This includes CIM errors about failed authentication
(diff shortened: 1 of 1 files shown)


docs/pywbemclicmdlineinterface.rst
```diff
@@ -269,3 +269,60 @@ following examples, an underscore ``_`` is shown as the cursor:
...
+    error. This will mostly be caused by CIM errors returned by the server,
+    but can also be caused by the pywbemcli code itself.
+
+* **1 - Python traceback**: In such cases, pywbemcli terminates during its
+  processing, and displays the Python stack traceback on standard error.
+
+  If this happens for a command entered in interactive mode, the pywbemcli shell
+  also terminates with a program exit code of 1.
+
+  These Python tracebacks should never happen and are always considered a
```
KSchopmeyer: Not completely true.  For an error in a mock-server python script we output the traceback to help the user find their problem.

docs/pywbemclicmdlineinterface.rst
```diff
@@ -269,3 +269,60 @@ following examples, an underscore ``_`` is shown as the cursor:
...
+  `pywbemtools issue tracker <https://github.com/pywbem/pywbemtools/issues>`_`.
+
+* **2 - User error**: In such cases, pywbemcli terminates without even
+  attempting to perform the requested operation, and displays one or more human
+  readable error messages on standard error.
+
+  If this happens for a command entered in interactive mode, the pywbemcli shell
+  is not terminated; only the command that failed is terminated.
+
+  An example for a user error is a missing required argument, or the use of an
```
KSchopmeyer: or an invalid value on an option

docs/pywbemclicmdlineinterface.rst
andy-maier: I agree. I looked at how it does that and it is an error message with multiple lines that include the traceback. So I'll clarify both the item about errors and the traceback statement you commented on.

BTW, I have opened issue #373 on the way this error is displayed, but I think we can defer that to 0.6.0.

docs/pywbemclicmdlineinterface.rst
andy-maier: DONE

docs/pywbemclicmdlineinterface.rst
andy-maier: DONE

PR #372, opened 2019-09-20
Fixes issue # 358 Fix instance count command

KSchopmeyer:
Modifications to instance count

1. Adds namespace to table output
2. Expanded instance count command to allow multiple namespaces. default is to search all namespaces
3. Clarify that the count is for instances of the class named in the
table, not counting subclasses.
4. Fix issue with actual count. There was a class name comparison that
was not case insensitive.  Since in the mock tested the classname uses different case in the class declaration and instance declarations, the bad test failed.

The fix extended over a number of files.

1. Make common option for multiple namespaces (class find and instance
count)
2. Add note to mock_assoc_model.mof about the case difference in class
name.  Better than fixing it as it will catch errors.
3. Change the same option in class find to use the common option.

### DIFF
--- a/docs/pywbemclicommands.rst
+++ b/docs/pywbemclicommands.rst
@@ -439,15 +439,25 @@ Instance count command
 ^^^^^^^^^^^^^^^^^^^^^^
 
 The ``instance count`` command counts the CIM instances of some or all classes
-in the namespace specified with the ``-namespace``/``-n`` command option, or
-otherwise in the default namespace of the connection.
+in the namespaces specified with the ``-namespace``/``-n`` command option, or
+all namespaces in the server.
 
-The count of instances is displayed for each class name.
-**TBD: Does it include instances of subclasses?**
+This command displays the count of instances of each CIM class whose class name
+matches the specified wildcard expression (CLASSNAME-GLOB) in all CIM
+namespaces of the WBEM server, or in the specified namespaces (--namespace
+option).  This differs from instance enumerate, etc. in that it counts the
+instances specifically for the classname of each instance returned (the
+creation classname), not including subclasses.
 
-If the ``CLASSNAME-GLOB`` argument is specified, only the classes are counted
-that match the specified :term:`Unix-style path name pattern`.
-Otherwise, all classes are counted.
+If the ``CLASSNAME-GLOB`` argument is specified, only instances of classes that
+match the specified :term:`Unix-style path name pattern` are counted. If the
+``CLASSNAME-GLOB`` argument is not specified all instances of all classes in
+the target namespaces are counted.
+
+Results for class that have no instances are not displayed.
+
+This command can take a long time to execute since it potentially enumerates
+all instance names for all classes in all namespaces.
 
 Valid output formats are :term:`Table output formats`.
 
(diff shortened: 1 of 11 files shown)


pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -542,29 +542,34 @@ def instance_query(context, query, **options):
...
+    wildcarding are supported: `*` to match zero or more characters, and `?` to
+    match a single character. To avoid shell expansion of wildcards, the
+    CLASSNAME-GLOB argument should be put in quotes.
 
-    The CLASSNAME-GLOB argument is a wildcard expression that is matched on
-    the class names case insensitively. The special characters known from file
-    nme wildcarding are supported: `*` to match zero or more characters, and
-    `?` to match a single character. In order to not have the shell expand
-    the wildcards, the CLASSNAME-GLOB argument should be put in quotes.
+    If CLASSNAME-GLOB is not specified, the all classes in the specified
```
andy-maier: typo "the" -> "then"

docs/pywbemclicommands.rst
```diff
@@ -439,15 +439,25 @@ Instance count command
...
-**TBD: Does it include instances of subclasses?**
+If the ``CLASSNAME-GLOB`` argument is specified, only instances of classes that
+match the specified :term:`Unix-style path name pattern` are counted. If the
+``CLASSNAME-GLOB`` argument is not specified all instances of all classes in
+the target namespaces are counted.
 
-If the ``CLASSNAME-GLOB`` argument is specified, only the classes are counted
-that match the specified :term:`Unix-style path name pattern`.
-Otherwise, all classes are counted.
+Results for class that have no instances are not displayed.
```
andy-maier: grammar "class" -> "classes"

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -542,29 +542,34 @@ def instance_query(context, query, **options):
...
-    (--namespace option).
+    namespaces of the WBEM server, or in the specified namespaces
+    (--namespace option).  This differs from instance enumerate, etc. in that
+    it counts the instances specifically for the classname of each instance
+    returned, not including subclasses.
+
+    The CLASSNAME-GLOB argument is a wildcard expression that is matched on the
+    class names case insensitively. The special characters known from file name
+    wildcarding are supported: `*` to match zero or more characters, and `?` to
+    match a single character. To avoid shell expansion of wildcards, the
```
andy-maier: It also supports "[]" ranges. In the docs description, that detail was now moved to the pattern term description, but in the help text I think we should mention it.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -542,29 +542,34 @@ def instance_query(context, query, **options):
...
+    CLASSNAME-GLOB argument should be put in quotes.
 
-    The CLASSNAME-GLOB argument is a wildcard expression that is matched on
-    the class names case insensitively. The special characters known from file
-    nme wildcarding are supported: `*` to match zero or more characters, and
-    `?` to match a single character. In order to not have the shell expand
-    the wildcards, the CLASSNAME-GLOB argument should be put in quotes.
+    If CLASSNAME-GLOB is not specified, the all classes in the specified
+    namespaces are counted (GLOB "*").
```
andy-maier: Instead of the somewhat cryptic "(GLOB "*")", I suggest either removing it or saying "(same as when specifying CLASSNAME-GLOB as "*")".

pywbemtools/pywbemcli/_cmd_instance.py

docs/pywbemclicommands.rst

pywbemtools/pywbemcli/_cmd_instance.py

pywbemtools/pywbemcli/_cmd_instance.py

Issue #349, opened 2019-09-16
Consistency of prompting for user choice


andy-maier:
Currently, there are different ways how a user is prompted for a choice:
- 6 cases where operations on a single instance prompt for the instance to operate on. Triggered by option --interactive. Specified by passing a class name in the INSTANCENAME argument.
- 1 case where an operation on a connection definition prompts for the connection definition to operate on. Triggered by passing `?` for the NAME argument.

**COMMENT/KS** and two cases 'connection delete' and 'connect select' where the prompt is used if there is no NAME argument.  Note that in the connections group, the show is an outlier because it wants to take both the connections file and the current connection into account where select and delete only touch the connections file.

Can we define an approach that is more consistent than that?

KSchopmeyer:
These cases are very different, they are not a general auto-completion mechanism.

1. INSTANCENAME - Goal was to simplify inputting INSTANCENAME by allowing select process to pick the name but requires  the CLASSNAME as a starting point. The selection process shows the instances are returns a valid one.

2. connection select/delete.  Allows alternative of selection if there is no NAME argument.. 'connection select' prompts but 'connection select blah' processes the name blah. The goal  is to act on a name in the connections file so the alternative of providing the NAME or nothing which activates the selection is easy.  In fact if TAB completion could be used in general we could simply have specified that (name or TAB autocomplete).  I cannot see any reason that adding an option here would help. This would be completly consistent with shell auto completion if we could use tab and if we could get to the list outside of pywbemcli.

3. The outlier to all this is "connection show".  The only reason for the ? option is because we really have 3 choices here we want the user to make since we want to be able to show 1) the current connection (connection show), 2) a named NAME argument connection(connection show blah) or 3) the list of existing connections(currently connection show ?. 

NOTE: I separate click autocomplete from bash autocomplete. There is a bash autocomplete for the static components (i.e options and choice type arguments. But I am not sure we could extend it to doing bash auto complete for connections in the connections file and I would really guess NO for getting instances of a class

IDEAS:

i mispoke in previous conversationabout the --interactive.  We could actually drop the interactive option and if only the class was provided (no keys) do the interaction because there are no singleton classes in CIM today or even to be sure create an instance name with no key and test it on the server.  Thus, the --interactive would disappear.  If the user supplied a class with no keys we would first try it as an instance name and if that failed, we would do the prompt for instance names.

[...]

PR #374, opened 2019-09-21
Fixes issue #349 consistency of prompting

KSchopmeyer:
1. Modified code to use ? as a wild card for  the key component of
wbem uri to kick off the selection process.  This modifies all the
following instance commands: get, delete, modify, references, associatiors,
invokemethod.

2. Modified the instance tests and the doc on the instance get, delete,
associations, and references, etc. to reflect this new way of specifying
the INSTANCENAME.

3. I think that the connection show is correct now in that it uses
the current connection if name not provided and uses ? if the user wants
the prompted list.

4. Modified the documentaton for instance get, delete, modify,
references, associators, invokemethod.  I shortened the test in each
command in favor of pointing them to the general section on the
INSTANCENAME in features.

This turns out to be a  really good solution, much better than the
--interactive option.

NOTE: I fixed one ref definition error in pywbemclicmdline.rst

### DIFF
--- a/docs/pywbemclicommands.rst
+++ b/docs/pywbemclicommands.rst
@@ -387,13 +387,8 @@ Instance associators command
 The ``instance associators`` command lists the CIM instances that are associated
 with the specified source instance.
 
-The instance name (INSTANCENAME argument) can be specified in two ways as
-defined in :ref:`Interactively selecting INSTANCENAME`:
-
-* By specifying an untyped WBEM_URI.
-
-* By specifying the WBEM_URI with the wild card "?" in place of the keys
-  component of the WBEM_URI, (i.e. <classname>.?).
+The specification of the instance name (INSTANCENAME argument) is documented
+in the section :ref:`Specifying the INSTANCENAME command argument`.
 
 If the ``--names-only``/``--no`` command option is set, only the instance paths
 are displayed. Otherwise, the instances are displayed.
@@ -517,13 +512,8 @@ Instance delete command
 
 The ``instance delete`` command deletes a CIM instance.
 
-The instance name (INSTANCENAME argument) can be specified in two ways as
-defined in :ref:`Interactively selecting INSTANCENAME`:
-
-* By specifying an untyped WBEM_URI.
-
-* By specifying the WBEM_URI with the wild card "?" in place of the keys
-  component of the WBEM_URI, (i.e. <classname>.?).
+The specification of the instance name (INSTANCENAME argument) is documented
+in the section :ref:`Specifying the INSTANCENAME command argument`.
 
 The following example deletes an instance by specifying its instance name.
 Note the extra backslash (see :term:`backslash-escaped`) that is required to
@@ -578,13 +568,8 @@ Instance get command
 
 The ``instance get`` command gets a CIM instance.
 
-The instance name (INSTANCENAME argument) can be specified in two ways as
-defined in :ref:`Interactively selecting INSTANCENAME`:
-
-* By specifying an untyped WBEM_URI.
-
-* By specifying the WBEM_URI with the wild card "?" in place of the keys
-  component of the WBEM_URI, (i.e. <classname>.?).
+The specification of the instance name (INSTANCENAME argument) is documented
+in the section :ref:`Specifying the INSTANCENAME command argument`.
 
 The command displays the instance using :term:`CIM object output formats`
 or :term:`Table output formats`.
@@ -598,7 +583,7 @@ This example gets an instance by instance name:
        name = "Saara";
     };
 
-or using the keys wild card:
+or using the keys wildcard:
 
 .. code-block:: text
 
@@ -624,13 +609,8 @@ Instance invokemethod command
 The ``instance invokemethod`` command invokes a CIM method on the specified
 instance and displays the return value and any output parameters.
 
-The instance name (INSTANCENAME argument) can be specified in two ways as
-defined in :ref:`Interactively selecting INSTANCENAME`:
-
-* By specifying an untyped WBEM_URI.
-
-* By specifying the WBEM_URI with the wild card "?" in place of the keys
-  component of the WBEM_URI, (i.e. <classname>.?).
+The specification of the instance name (INSTANCENAME argument) is documented
+in the section :ref:`Specifying the INSTANCENAME command argument`.
 
 Input parameters for the method can be specified with the ``--parameter``/``-p``
 command option, which can be specified multiple times.
@@ -649,7 +629,7 @@ Example:
     ReturnValue=0
     arrBool=true, false
 
-Or using the wild card to create a selection list for the instance names
+Or using the wildcard to create a selection list for the instance names
 
 .. code-block:: text
 
@@ -675,13 +655,8 @@ Instance modify command
 The ``instance modify`` command modifies the properties of an existing CIM
 instance.
 
-The instance name (INSTANCENAME argument) can be specified in two ways as
-defined in :ref:`Interactively selecting INSTANCENAME`:
-
-* By specifying an untyped WBEM_URI.
-
-* By specifying the WBEM_URI with the wild card "?" in place of the keys
-  component of the WBEM_URI, (i.e. <classname>.?).
+The specification of the instance name (INSTANCENAME argument) is documented
+in the section :ref:`Specifying the INSTANCENAME command argument`.
 
 The new property values are specified by possibly multiple ``--property``/``-p``
 command options.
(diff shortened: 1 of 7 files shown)


docs/pywbemclicommands.rst
```diff
@@ -387,22 +387,13 @@ Instance associators command
...
-  ``-namespace``/``-n`` command option, or otherwise the default namespace
-  of the connection.
-  For details, see :ref:`Interactively selecting INSTANCENAME`.
+The instance name (INSTANCENAME argument) can be specified in two ways as
+defined in :ref:`Interactively selecting INSTANCENAME`:
+
+* By specifying an untyped WBEM_URI.
+
+* By specifying the WBEM_URI with the wild card "?" in place of the keys
+  component of the WBEM_URI, (i.e. <classname>.?).
```
andy-maier: Use "CLASSNAME" instead of "<classname>", for consistency.

docs/pywbemclicommands.rst
```diff
@@ -526,22 +517,13 @@ Instance delete command
...
 
-* By specifying the ``--interactive`` command option and a class name in the
-  ``INSTANCENAME`` argument. The instances of the specified class are displayed
-  and the user is prompted for an index number to select an instance. The
-  namespace of the instance is the namespace specified with the
-  ``-namespace``/``-n`` command option, or otherwise the default namespace
-  of the connection.
-  For details, see :ref:`Interactively selecting INSTANCENAME`.
+* By specifying the WBEM_URI with the wild card "?" in place of the keys
+  component of the WBEM_URI, (i.e. <classname>.?).
```
andy-maier: Use "CLASSNAME" instead of "<classname>", for consistency.

docs/pywbemclicommands.rst
```diff
@@ -596,22 +578,13 @@ Instance get command
...
```
andy-maier: Use "CLASSNAME" instead of "<classname>", for consistency.

docs/pywbemclicommands.rst
```diff
@@ -636,22 +624,13 @@ Instance invokemethod command
...
```
andy-maier: Use "CLASSNAME" instead of "<classname>", for consistency.

docs/pywbemclicommands.rst
```diff
@@ -681,22 +675,13 @@ Instance modify command
...
```
andy-maier: Use "CLASSNAME" instead of "<classname>", for consistency.

docs/pywbemclicommands.rst
```diff
@@ -387,22 +387,13 @@ Instance associators command
...
-  ``INSTANCENAME`` argument. The instances of the specified class are displayed
-  and the user is prompted for an index number to select an instance. The
-  namespace of the instance is the namespace specified with the
-  ``-namespace``/``-n`` command option, or otherwise the default namespace
-  of the connection.
-  For details, see :ref:`Interactively selecting INSTANCENAME`.
+The instance name (INSTANCENAME argument) can be specified in two ways as
+defined in :ref:`Interactively selecting INSTANCENAME`:
+
+* By specifying an untyped WBEM_URI.
```
andy-maier: Use "WBEM URI" instead of "WBEM_URI".
Occurs multiple times.

docs/pywbemclicommands.rst
```diff
@@ -387,22 +387,13 @@ Instance associators command
...
-
-* By specifying the ``--interactive`` command option and a class name in the
-  ``INSTANCENAME`` argument. The instances of the specified class are displayed
-  and the user is prompted for an index number to select an instance. The
-  namespace of the instance is the namespace specified with the
-  ``-namespace``/``-n`` command option, or otherwise the default namespace
-  of the connection.
-  For details, see :ref:`Interactively selecting INSTANCENAME`.
+The instance name (INSTANCENAME argument) can be specified in two ways as
+defined in :ref:`Interactively selecting INSTANCENAME`:
```
andy-maier: On referencing "Interactively selecting INSTANCENAME" for both ways:
We have two sections detailing how to specify INSTANCENAME:
* "Specifying the INSTANCENAME command argument" - for first way
* "Interactively selecting INSTANCENAME" - for second way
We should either combine these section into one and reference that one, or reference the correct section for each of the two ways.

docs/pywbemclifeatures.rst
```diff
@@ -289,20 +289,44 @@ The following are examples of scalar property definitions:
 Interactively selecting INSTANCENAME
 ------------------------------------
 
-The INSTANCENAME argument has a certain complexity, particularly for
-associations and for classes with multiple keys.
+The INSTANCENAME argument can be complex, particularly for associations and
+classes with multiple keys.
+
+To simplify using INSTANCENAME on the command line, pywbemcli provides a wild
+card character "?" that can be used with commands in place of the INSTANCENAME
```
andy-maier: Suggest removing "with commands".

andy-maier: Suggest replacing "INSTANCENAME" with "instance name"

docs/pywbemclifeatures.rst
```diff
@@ -289,20 +289,44 @@ The following are examples of scalar property definitions:
...
+keys on commands that have an INSTANCENAME argument. This allows the user to
```
andy-maier: Since we are scoping the sentence already to INSTANCENAME as per its first words, I think we can remove " on commands that have an INSTANCENAME argument".

docs/pywbemclifeatures.rst
```diff
@@ -289,20 +289,44 @@ The following are examples of scalar property definitions:
...
+specify only the class name (as the INSTANCENAME argument value) with the wild
+card for the keys. Pywbemcli retrieves all instance names of that class from
```
andy-maier: The sentence "This allows the user ..." basically says what was introduced in the previous sentence can actually be used. I assume you mean to convey that the syntax is class name followed by the wild card for keys, but to me that is not what it says, plus it is not correct because the dot would be missing in that interpretation.
I suggest replacing this sentence with an ABNF syntax definition.

docs/pywbemclifeatures.rst
```diff
@@ -289,20 +289,44 @@ The following are examples of scalar property definitions:
...
+  of the connection. Any host name in the WBEM URI will be ignored.
+  For details, see :ref:`Specifying the INSTANCENAME command argument`.
+
+* By specifying the ``--interactive`` command option and a class name in the
+  ``INSTANCENAME`` argument. The instances of the specified class are displayed
+  and the user is prompted for an index number to select an instance. The
+  namespace of the instance is the namespace specified with the
+  ``-namespace``/``-n`` command option, or otherwise the default namespace
+  of the connection.
+  For details, see :ref:`Interactively selecting INSTANCENAME`.
```
andy-maier: The paragraphs above, from "Thus an INSTANCENAME" to here do not make sense in this section "Interactively selecting INSTANCENAME". They would make more sense (without the section references) when combining the two sections into one.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -160,26 +154,25 @@ def instance_group():
...
+    in the --namespace option, or otherwise the default namespace of the
+    connection. Any host name in the WBEM URI will be ignored.
 
-    * By specifying the --interactive option and a class name in the
-      INSTANCENAME argument. The instances of the specified class are displayed
-      and the user is prompted for an index number to select an instance.
-      The namespace in which the instances are looked up is the namespace
-      specified in the --namespace option, or otherwise the default namespace
-      of the connection.
+    2. By specifying a class name with the characters ".?" as the keys in the
```
andy-maier: On "By specifying a class name with the characters ".?" as the keys in the INSTANCENAME argument": I don't find this an intuitive description of the syntax. Also, since we introduced the term "wildcard keys" i the docs, it makes sense to also use it in the help text.

How about: "By specifying a class name with wildcard keys in the INSTANCENAME argument, i.e. "CLASSNAME.?""

Occurs multiple times.

docs/pywbemclicommands.rst

docs/pywbemclifeatures.rst
KSchopmeyer: DONE

KSchopmeyer: How about the INSTANCENAME argument since that is what it is universally defined as.

DONE

docs/pywbemclifeatures.rst
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_instance.py

docs/pywbemclifeatures.rst
KSchopmeyer: Redid to add ABNF

docs/pywbemclifeatures.rst
KSchopmeyer: Completely redid both the descriptions in pywbemclicommands and the featues section.  Made the section on that argument very short in each command spec and put it all together into a single section in features.  Lets see if that reads better.

docs/pywbemclicommands.rst
KSchopmeyer: Did not quite understand this one.  Hope I fixed it correctly.

andy-maier: The `<classname>` was not showing up in the view mode of my comment. If you go into edit mode, you can see it. Sorry for not checking that.
You guessed correctly in what you fixed.

docs/pywbemclifeatures.rst
andy-maier: I'm on with "INSTANCENAME argument".

docs/pywbemclifeatures.rst
andy-maier: Reviewed and agreed.

Issue #350, opened 2019-09-16
Allow unsetting of general options in interactive mode


KSchopmeyer:
Today we can set general options like --certfile in interactive mode so that they modify the current connection but we cannot unset them, i.e. return them to the default value (The value used when the options is not included on the command line) because that capability does not exist with the click options.   In interactive mode we only change those options that are specifically defined with the command (--certfile abc.pem) and do not change options that are not specified..

Click does not include the concept of an optional value on  options that have a value component (ex. string) so you cannot do something like --certfile  to tell pywbemcli that the option itself exists but the value is None.

However, we can extend the string options so that the value "" is equivalent to None.  In other words, inputting --certfile "" defines the option bur sets the value to None rather than to "" internally.

I think we should do this but I don't think it is a requirement for Version 0.5.0.  I propose it for version 0.6.0

andy-maier:
Agreed to do move this beyond 0.5.0.

andy-maier:
PR #382 addresses this for the --verbose option.

KSchopmeyer:
I boolean flag options (--verify, --verbose) were corrected in version 0.5.0.  This leaves the following options that requires some means of resetting them.

--log
--timestats
--use-pull
--certfile
--ca-certs
--password
--user

These are string options so in general setting the string to "" should be the way to return them to default value (i.e. the same as if they were not specified)

KSchopmeyer:
We agreed to move this one to 7.0

KSchopmeyer:
General Options that will allow resetting to a default value in interactive mode when "" input as value. Thise are all string options:

* --user
* --password
* --certfile
* --keyfile
* --default-namespace (Resets to root/cim/v2)

andy-maier:
Given that there is no good solution for the remaining three options, I suggest that we declare this one completely fixed.

**DISCUSSION**

PR #382, opened 2019-09-23
Allow turning off --verbose in interactive mode 

KSchopmeyer:
Allow turning off --verbose in interactive mode by changing thelong definition name to --verbose/--noverbose and the default for verbose to None on the cli function call.

This PR also adds a display when verbose is set that is turning
output to be worthwhile. It displays the current command and its
parameters when --verbose is set just before the command is called. Note
that this only displays that command, not the command group name


andy-maier:
This addresses issue #350 for the verbose option.

Issue #375, opened 2019-09-21
Mock repo reset after each command in interactive mode


KSchopmeyer:
We clear the wbem server for each new command, not just when we are modifying the inut parameters.  This means that if you delete an instance in interactive mode, it pops up again for the next interactive command because clearing the server also clears the repository.

Must clean up the clearing of the server and repo even in the case where we are modifying some parameters.

andy-maier:
I think that one major purpose of the interactive mode is to be able to run multiple commands against the same mock repository. For a real WBEM server it does not matter if the connection is re-established on each command, but for a mock server it does matter.
Bottom line: I think it would be good if we could address this in 0.5.0

PR #388, opened 2019-09-24
Fixes issue #375 - resetting mock repo between interactive commands

KSchopmeyer:
Fix issue in pywbemcli.py where we were dropping information in setup of interactive commands so that each command reinitialized the mock
repository.

This also enables a test of multiple instance commands through stdin
with that verifies fixing the reinitialization problem.

Finally, we added a display in verbose mode of each command and it
params before the execution of the command is started. This is a
diagnostic.


andy-maier:
There must be something wrong in this PR. Please double check what it does. For example, it removes the --no-verbose option that was introduced in PR #382.

PR #389, opened 2019-09-24
Disabled brew update and upgrade of Python in Travis on OS-X

andy-maier:


andy-maier:
It ends up not saving any time, because now the long upgrade time is spent in the first package installed by pywbem_os_setup.sh.
Abandoning the experiment.

andy-maier:
Closing this. What is needed instead, is the fix in https://github.com/pywbem/pywbem/pull/1799.

Issue #397, opened 2019-09-26
"connection show" shows the password


andy-maier:
Example:
```
$ pywbemcli connection show testconn
name: testconn (current)
  server: http://localhost
  default-namespace: root/cimv2
  user: usr1
  password: blah
  timeout: 30
  verify: True
  certfile: None
  keyfile: None
  mock-server: 
  ca-certs: ['/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt', '/etc/ssl/certs', '/etc/ssl/certificates']
```

That has the risk that someone who demonstrates pywbemcli unintentionally reveals the passsword of a server.

**Proposal:**

Change the default to not show the password, but allow showing it when explicitly requested, e.g. as follows:
* Add a boolean command option `--show-password` to "connection show".
* If the option is not set, show the password value as 8 asterisks.

**Discussion point:** Which release?

KSchopmeyer:
I suggest this release.   This one is straight forward and with the option, It would be ok to put it on the connection list also.

andy-maier:
Addressed by PR #398.

PR #398, opened 2019-09-26
Fixes issue #397, mask password on connection show display

KSchopmeyer:
Adds option to conneciton show to mask password unless option
--show-password defined.

Add tests for this option.

Note: No short form for this option makes them work harder to see the
passwords.


KSchopmeyer:
Fixed issue we discussed.  pw mask is firxed length.

PR #405, opened 2019-09-28
Fix issue # 394 - Aborts session in interactive mode

KSchopmeyer:
We agreed that we limit the issues that cause abort of the session to
just errors in creating mof or py entities in the repository.

Removed the aborts on all errors except mock-server MOF or python script
errors in the files themselves. Even nonexistent or misnamed files do
not cause an abort.

Adds tests for both the Abort from interactive and the cmds that do not
abort in interactive mode.

### DIFF
--- a/pywbemtools/pywbemcli/pywbemcli.py
+++ b/pywbemtools/pywbemcli/pywbemcli.py
@@ -326,7 +326,6 @@ def cli(ctx, server, svr_name, default_namespace, user, password, timeout,
             if local_svr_name:
                 pywbem_server = connections[local_svr_name]
                 # Test for invalid options with the --name option
-                # The following options are part of each PywbemServer object
                 for option in conditional_options:
                     if option[0]:
                         raise click.ClickException(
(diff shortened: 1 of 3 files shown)


pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -413,32 +402,30 @@ def create_server_instance(svr_name):
...
-                                                   '"--mock-server" python '
-                                                   'script %r. '
-                                                   'Exception: %r\n'
-                                                   'Traceback\n%s' %
-                                                   (file_path, ex, tb))
+                            exc_traceback)
+                        raise click.ClickException(
+                            'Exception failure of "--mock-server" python '
+                            'script {0!r}. Exception: %r\n'
+                            'Traceback\n{1}'.format(file_path, ex, tb))
```
andy-maier: `tb` is a list of lines, so to display them as one would expect, it is not sufficient to remote the `repr()` but in addition, the lines need to be joined with newline.

andy-maier: I can fix that in an upcoming PR that improves error messages.

andy-maier:
I am merging this one because the one little issue can be addressed in an upcoming PR I have in the pipe, that improves error messages.

PR #406, opened 2019-09-28
Fixes Ks/#396 creates bad conn file

KSchopmeyer:
Fixed issue and added test to bypass issue in the future.

Added test in test_connection_cmds.py for the sequence that caused the error.


KSchopmeyer:
There is a little horror show in the test added since if fails in windows trying to put the file path into the stdin strings.  loses all of the file path separators but only on windows.  I will file bug on that one but for now we just do not run that one test on windows.

Issue #419, opened 2019-09-30
Bad Exception statement in pywbemcli.py


KSchopmeyer:
A statement in pywbemcli.py mixes the % formatter and the .fromat formatter.

PR #420, opened 2019-09-30
Fixes issue #419. Mixed old an new formatters

KSchopmeyer:
One ClickException statement mixed the old and new formatters using the
{} from the new but the % template/variable list separator from the new
formatter.

Marked for rollback but I would NOT push specal release for this since it is really  code to support our mocking of things like prompts and the script input has an error.

### DIFF
--- a/tests/unit/py_err_processatstartup.py
+++ b/tests/unit/py_err_processatstartup.py
@@ -1,0 +1,14 @@
+# !PROCESS!AT!STARTUP!
+"""
+This is an error test of the code that processes a script at startup. This
+python code specifically includes a syntax error to be caught by pywbemcli.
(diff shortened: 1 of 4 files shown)


tests/unit/py_err_processatstartup.py
```diff
@@ -0,0 +1,14 @@
+# !PROCESS!AT!STARTUP!
+"""
+This is an error test of the code that processes a script at startup. This
+python code specifically includes a syntax error to be caught by pywbemcli.
+Note that this file is marked to process at startup which is really a private
+extension for the test environment.
+"""
+
+
+def mock?prompt(msg):  # noqa: E999
```
andy-maier: Is the question mark intentional? I hope not.

PR #445, opened 2019-10-17
Fixes #255: Reorg doc for multiple tools

### LINKED ISSUE #255: Docs: Move sections on pywbemcli command into one section
Sections 2, 3, 4, and 5 are all about pywbemcli.

If we add a second command, we probably would like to have those sections be subsections of a single H1 section named "Pywbemcli command".

We might want to do that already now, even though we have only one command. This would suggest to increase the number of heading levels shown in the TOC from 2 to 3.

KSchopmeyer:
I did this one early so we can easily work with what should be the structure of the documentation for the next release.

Reorganize the documentation to allow for multiple tools.  This pr moves
all of the pywbemcli specific documentation to a new subdirectory and
creates an index (toctree) for that documentation in the subdirectory (docs/pywbemcli/index) so that the pywbemcli documentation all becomes section 2 of the documentation.  All of the tool specific documentationis now in section 2 with a subsection for possible for each tool. Note that we did
not change the definitions of the headings (i.e. the type of underline for each heading) in the
documentation for pywbemcli.  It starts its headings at the heading level below the level from which it was called.

Added a new sphinx tool, fulltoc that shows the full toc.

However, despite the maxdepth parameter in each toctree, the sidebar shows more levels of toc than the 2 levels defined with this parameter as sublevels are selected from the sidebar.


KSchopmeyer:
I agree that the menu for section 2 should appear in the left window.  At this point I am not sure what the issue is but using the toctree attributes 'includehidden' and 'hidden' do not have any effect except to turn on/off the menu for section 2 within the section 2 window itself.  My original assumption was that the includehidden as an attribute for the master TOC (docs/index) should show the TOC for the pywbemcli sections in the subdirecctory but that does not work.   I am going to have to dig further to find out what is happening.

By the way, the python documentation all uses exactly this structure with a subdirectory for each major section of the documentation so something works here.  We have just not found the magic button yet.

KSchopmeyer:
Now shows complete overview on sidebar and details of section 2.  This is completely different TOC processor (FullToc, a sphinx contribution).

KSchopmeyer:
**DISCUSSION:** 

1. Section 2 only shows the top level now in the overall menu ( not 2.1, etc.)

**ERROR:  FIXED**
I think the mock chapter should be in pywbemcli at least for now.

dev-requirements.txt
```diff
@@ -37,6 +37,7 @@ pytest-cov>=2.4.0,<2.6
 Sphinx>=1.7.6
 sphinx-git>=10.1.1
 GitPython>=2.1.1;
+sphinxcontrib-fulltoc>=1.2.0
```
andy-maier: This new dependency also needs to be added to `minimum-constraints.txt` as:
```
sphinxcontrib-fulltoc==1.2.0
```

KSchopmeyer: DONE

Issue #443, opened 2019-10-11
Order the list of commands in the --help output


KSchopmeyer:
Extend pywbemcli so that we can order the display of the subcommands part of the --help display

KSchopmeyer:
This capability is not part of Click and Click does not intend to implement it.  There are two ways to
implement this as Click extension:  Both involve slightly extending the Group class to modify the list_command.  Note that this is clearly the type of thing that Click wants the developer to use because it specifically allows definition of the implementation class  with the decorator (ex. @click.group(cls=NaturalOrderGroup).  There is a click issue https://github.com/pallets/click/issues/513 that discusses this and notes that the click developers feel it is a user extension and that they do not want to make it part of Click.

1. Add a specific parameter that defines order for command class (Ex. HelpPrority) and modify the get_help to do the reordering when displayed.

2. Create a subclass to the click.command class that uses OrderedDictionary rather than dictionary and let the order of definition of the commands in each _cmd.xxx.py module define the display order.

Either one is only a few lines of code. The difference is that one means adding a new parameter to every option definition, the other means physically reordering the code for each option definitions. Either is more work than the actual code change to create the capability.

DISCUSSION: Any preference for one vs the other.

KSchopmeyer:
Set this to 0.6.0 optional.  Note that the change is minor but it is the first use of extensions of the Click Classes so allows us to create the basic structure.

At this point the proposed change will allow ordering of commands within a normal group.  However, the top level group (i.e. pywbemcli) must be handled in a special manner since there is no real internal ordering, just the order of the files that define each pywbemcli group. We must do that different than the other groups.

KSchopmeyer:
Submitted pr # 450 with proposal to use order of commands in source files (_cmds_xxx.py) ad order in list and to keep alphebatical ordering of commands for top level help except that non-server oriented commands (connection, help, repl go to bottom.

NOTE: The ordering for the top level was a SWAG so we can do whatever we really want to.

andy-maier:
FIxed in PR #450.

PR #450, opened 2019-11-03
Fixes issue #443, ordering of lists of commands in help

KSchopmeyer:
Changes the ordering of the lists of commands in help output from the
alphabetic to the order that the commands in each group were defined for all levels except top level.

Top level commands are ordered in the help display  to put connection, help, repl at bottom

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -34,6 +34,12 @@ Released: not yet
 
 **Enhancements:**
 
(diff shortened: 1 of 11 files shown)


KSchopmeyer:
NOTE: This is my proposed solution.  The issue # 443 is still open dor discussion of multiple possible solutions but this is my recommendation.

PR #454, opened 2019-11-29
Add autosuggest prompt toolkit extension

KSchopmeyer:
Depends on pr #457, Fix tests for pywbem 0.15.0

This adds a component of prompt-toolkit, the auto-suggest module.
This module gathers information from the file history to augment the
suggestions for parameters in the repl mode.

I have found it useful in completion of the data parts of options when the same classnames, etc. are used repeatedly.  However, I am not sure how it will interact with any other auto-complete we add.

Note that it is really only about a 3 line change, attaching the auto_suggest part of prompt to the repl.

TODO: If we think it is useful I need to add it to the doc.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -47,6 +47,12 @@ Released: not yet
   original form using, WBEMURI is error prone with quote marks.
   (see issue #390)
 
+* Add prompt-toolkit auto-suggest.  This extends the command completion 
+  capability in the repl mode (interactive mode) to make suggestions on
+  command line input based on the history file.  Usually auto-suggest completion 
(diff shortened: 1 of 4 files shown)


KSchopmeyer:
The latest version has docs added, and also a config variable so the capability can be disabled.

NOTE: Looks like we are having problems with the AppVeyor build

KSchopmeyer:
NOTE: IWe do not mention auto-suggestion in the help files but then we do not mention the auto-complete there either.  I am trying to keep the help files as simple as possible since it is not easy to read them when they are several terminal pages long.

PR #455, opened 2019-12-01
Fixes #390: Simpler expression for INSTANCENAME

### LINKED ISSUE #390: Less error prone form of INSTANCENAME
The INSTANCENAME as WBEM URI is difficult to enter an error prone.  The use of  wildcard helps but is interactive.

One of our goals is to make pywbemcli as easily usable as possible, in particular make the command line input both intuitive and as free of specialized formats, characters, etc. as possible. Through experience with cimcli it became obvious that entering the CIMInstanceName parameter manually in a form acceptable to the WBEM URI standards was messy, partly because of the requirements for the double-quotes which must be escaped from most command line environments. The instance name is messy, involving multiple key properties and having to enter each with escaped double quotes for all non numerics caused users, especially casual users headaches in getting everything correct. This was compounded by the fact that a malformed instance name input on the command line simply generates an error that it was malformed.

There are several ways to make this easier including:

1. The wildcard which we already have which uses the classname with a wildcard "?* for the keybindings component and pywbemcli presents a list of possible known instance names.

2. Eliminate the need for the quotes as much as possible. We can make a form that is easier to input by eliminating the need for the quote marks in most cases.  This would use a parameter similar to the --property in instance create but it would be --key where each use would be a single key.  pywbemcli would assemble the CIMInstanceName from the individual keys including addin sayg quotes, etc.

3. Lead the user through the process step by step where pywbemcli would determine the key properties and present each key property with its type and ask the user to input a string representing the value

KSchopmeyer:
NOTE: This pr based on pr # 457, (fix tests as we move to pywbem 0.15.0)

Add option -k --key to allow keys of the instance name to be defined
separately and then aggragated by pywbemcli.

This allows us to handle the messy parts of creating CIMInstanceName
keybindings without having to involve the user in quote marks, etc.

Originally INSTANCENAME would have to be expressed as a WBEM_URL
complete with correctly placed double quotes.

CIM_Foo.InstanceId="blah"

This allows entering the INSTANCENAME as

CIM_Foo -k InstanceID=blah

### DIFF
--- a/docs/pywbemcli/features.rst
+++ b/docs/pywbemcli/features.rst
@@ -230,6 +230,12 @@ The instance name (INSTANCENAME argument) can be specified in two ways:
   the ``-namespace``/``-n`` command option, or the default namespace of the
   connection.
 
+* By specifying the WBEM URI without keybindings and using the --key option
+  to specify the keybindings ad defined in section
+  :ref:`Defining INSTANCENAME command argument with --key option`. The
+  advantage of this technique is that it eliminates the use of the double
+  quote surrounding the key values.
+
 
 .. _`The INSTANCENAME command argument as a WBEM URI`:
 
@@ -361,3 +367,19 @@ interactive selection, as shown in the following example:
        InstanceID = "CIM_Foo1";
        IntegerProp = 1;
     };
+
+
+.. _`Defining INSTANCENAME command argument with --key option`:
+
+`Defining INSTANCENAME command argument with --key option`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The INSTANCENAME may be specified by a combination of the namespace/classname
+as an argument with the --key option to define keybindings. Each --key option
+definition defines a single keybinding in the form name=value.   In general,
+the value component does not require the double quote that is required with
+WBEM_URL unless there are space characters in a string value.
+
+  Example:
+
+    CIM_Foo --key InstanceId=inst1
--- a/pywbemtools/pywbemcli/_cmd_instance.py
+++ b/pywbemtools/pywbemcli/_cmd_instance.py
@@ -22,6 +22,7 @@ NOTE: Commands are ordered in help display by their order in this file.
 
 from __future__ import absolute_import, print_function
 
+import re
 import click
 
 from pywbem import Error, CIMError, CIM_ERR_NOT_FOUND
@@ -30,7 +31,8 @@ from .pywbemcli import cli
 from ._common import display_cim_objects, parse_wbemuri_str, \
     pick_instance, resolve_propertylist, create_ciminstance, \
     filter_namelist, CMD_OPTS_TXT, format_table, verify_operation, \
-    process_invokemethod, raise_pywbem_error_exception
+    process_invokemethod, raise_pywbem_error_exception, \
+    create_ciminstancename, warning_msg
 
 from ._common_options import add_options, propertylist_option, \
     names_only_option, include_classorigin_instance_option, namespace_option, \
@@ -112,6 +114,14 @@ property_modify_option = [              # pylint: disable=invalid-name
                       'comma-separated list; embedded instances are not '
                       'supported. '
                       'Default: No properties modified.')]
+
+keybinding_key_option = [              # pylint: disable=invalid-name
+    click.option('-k', '--key', type=str, metavar='KEYNAME=VALUE',
+                 required=False, multiple=True,
+                 help='Value for a key in keybinding of CIM instance name. '
+                      'May be specified multiple times. '
+                      'Allows defining keys without the issues of quotes. '
+                      'Default: No keybindings provided.')]
 
 filter_query_language_option = [              # pylint: disable=invalid-name
     click.option('--fql', '--filter-query-language', 'filter_query_language',
@@ -198,6 +208,7 @@ def instance_enumerate(context, classname, **options):
 @add_options(include_qualifiers_get_option)
 @add_options(include_classorigin_instance_option)
 @add_options(propertylist_option)
+@add_options(keybinding_key_option)
 @add_options(namespace_option)
 @click.pass_obj
 def instance_get(context, instancename, **options):
@@ -232,6 +243,7 @@ def instance_get(context, instancename, **options):
 
 @instance_group.command('delete', options_metavar=CMD_OPTS_TXT)
 @click.argument('instancename', type=str, metavar='INSTANCENAME', required=True)
+@add_options(keybinding_key_option)
 @add_options(namespace_option)
 @click.pass_obj
 def instance_delete(context, instancename, **options):
@@ -301,6 +313,7 @@ def instance_create(context, classname, **options):
               'be modified. '
               'Default: Do not reduce the properties to be modified.')
 @add_options(verify_option)
+@add_options(keybinding_key_option)
 @add_options(namespace_option)
 @click.pass_obj
 def instance_modify(context, instancename, **options):
@@ -358,6 +371,7 @@ def instance_modify(context, instancename, **options):
 @add_options(include_classorigin_instance_option)
 @add_options(propertylist_option)
 @add_options(names_only_option)
+@add_options(keybinding_key_option)
 @add_options(namespace_option)
 @add_options(summary_option)
 @add_options(filter_query_option)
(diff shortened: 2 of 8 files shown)


docs/pywbemcli/features.rst
```diff
@@ -361,3 +367,19 @@ interactive selection, as shown in the following example:
...
+.. _`Defining INSTANCENAME command argument with --key option`:
+
+`Defining INSTANCENAME command argument with --key option`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The INSTANCENAME may be specified by a combination of the namespace/classname
+as an argument with the --key option to define keybindings. Each --key option
+definition defines a single keybinding in the form name=value.   In general,
+the value component does not require the double quote that is required with
+WBEM_URL unless there are space characters in a string value.
```
andy-maier: Change "WBEM_URL" to "WBEM URI"

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -113,6 +115,14 @@
                       'supported. '
                       'Default: No properties modified.')]
 
+keybinding_key_option = [              # pylint: disable=invalid-name
+    click.option('-k', '--key', type=str, metavar='KEYNAME=VALUE',
+                 required=False, multiple=True,
+                 help='Value for a key in keybinding of CIMInstanceName. '
```
andy-maier: I suggest not to talk about pywbem CIM object classes (CIMInstanceName", but to say "CIM instance name" instead.

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -113,6 +115,14 @@
...
+                      'May be specified multiple times. '
+                      'Allows defining keys without the issues of quotes. '
+                      'Default: No initial properties provided.')]
```
andy-maier: On "No initial properties provided": These are neither initial nor general properties. Suggest to change this to "No keybindings provided".

pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -939,7 +1041,7 @@ def cmd_instance_count(context, classname, options):
         except CIMError as ce:
             # allow processing to continue if no interop namespace
             if ce.status_code == CIM_ERR_NOT_FOUND:
-                click.echo('WARNING: {}'.format(ce))
+                warning_msg('{}'.format(ce))
```
andy-maier: This change is unrelated to the purpose of this PR. Please move to a separate PR.
Occurs multiple times in this file.

pywbemtools/pywbemcli/_common.py
```diff
@@ -456,6 +456,51 @@ def create_cimproperty(cim_type, is_array, name, value_str):
     return CIMProperty(name, cim_value, cim_type)
 
 
+def create_ciminstancename(cim_class, kv_keys):
+    """
+    Create a cim instance from the input parameters.
```
andy-maier: It creates a CIM instance name.

pywbemtools/pywbemcli/_common.py
```diff
@@ -456,6 +456,51 @@ def create_cimproperty(cim_type, is_array, name, value_str):
...
+
+      Parameters:
+
+        cim_class_: (:Class:`pywbem.CIMClass`)
+            The class from which the CIMInstance is to be created
+
+        properties ():
+            A tuple of name/value pairs representing the properties and their
+            values that are to be constructed for the instance. Required
```
andy-maier: Adjust the parameter description to the actual parameter.

pywbemtools/pywbemcli/_common.py
```diff
@@ -456,6 +456,51 @@ def create_cimproperty(cim_type, is_array, name, value_str):
...
 
 
+def create_ciminstancename(cim_class, kv_keys):
+    """
+    Create a cim instance from the input parameters.
+
+      Parameters:
+
+        cim_class_: (:Class:`pywbem.CIMClass`)
+            The class from which the CIMInstance is to be created
```
andy-maier: `CIMInstance` -> `CIMInstanceName`

pywbemtools/pywbemcli/_common.py
```diff
@@ -456,6 +456,51 @@ def create_cimproperty(cim_type, is_array, name, value_str):
...
+            values that are to be constructed for the instance. Required
+
+      Returns: CIMInstanceName with namespace = None and host = None
+
+      Exceptions:
+        click.ClickException if Property name not found in class or if mismatch
+          of property type in class vs value component of kv pair
+    """
+    # TODO make usable if no class can be acquired because GetClass not
+    # implemented in server.
```
andy-maier: This TODO is unrelated to the code in this function. The CIMCLass object is already passed in, so if there is an issue with GetClass not being implemented, it would be wherever that GetClass is executed, and that's where this TODO should be.

pywbemtools/pywbemcli/_common.py
```diff
@@ -456,6 +456,51 @@ def create_cimproperty(cim_type, is_array, name, value_str):
...
+        name, value_str = parse_kv_pair(kv_property)
+        try:
+            cl_prop = cim_class.properties[name]
+        except KeyError:
+            raise click.ClickException('Property name "{}" not in class "{}".'
+                                       .format(name, cim_class.classname))
+
+        try:
+            if value_str.startswith('"') and value_str.endswith('"'):
+                value_str = value_str[1:-1]
```
andy-maier: How about `value_str = value_str.strip('"')`?

tests/unit/test_common.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
andy-maier: The coding statement is only needed if the file contains plan Unicode characters. We should not do that however, but always escape them as `\uXXXX` or `\UXXXXXXXX`.

tests/unit/test_common.py
```diff
@@ -1484,6 +1485,189 @@ def test_invalid_propname(self):
...
+                classname=u'CIM_Foo',
+                keybindings=NocaseDict(P1='Fred', P2='John', P3=123456)
+            ),
+        ),
+        None, None, True, ),
+    (
+        "Verify simple key creation with unicode char",
+        dict(
+            cls_kwargs=cls1,
+            kv_args=[u'P1=Fredäö'],
```
andy-maier: Please escape the Unicode characters.

tests/unit/test_common.py
```diff
@@ -1484,6 +1485,189 @@ def test_invalid_propname(self):
...
+            namespace=None,
+            host=None,
+            exp_attrs=dict(
+                classname=u'CIM_Foo',
+                keybindings=NocaseDict(P1=u'Fredäö')
```
andy-maier: Please escape the Unicode characters.

docs/pywbemcli/features.rst
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_instance.py

pywbemtools/pywbemcli/_cmd_instance.py

pywbemtools/pywbemcli/_cmd_instance.py

pywbemtools/pywbemcli/_common.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_common.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_common.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_common.py
KSchopmeyer: Moved to the common funciton in _cmd_instance.py and extended the idea somewhat. Still a todo and marked future

**DONE**

pywbemtools/pywbemcli/_common.py
KSchopmeyer: 1. That would actually have to be more complex because it would have to account for escaped quotes inside the string.  This simply removes the lead and end if they both exist which is what we want to do because otherwise the create_cimvalue gets confused and adds more quotes.

2. No problem moving outside try block.  **DONE**

tests/unit/test_common.py
KSchopmeyer: DONE

tests/unit/test_common.py
KSchopmeyer: DONE

tests/unit/test_common.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_common.py
andy-maier: `strip()` does remove only from start and end.

PR #456, opened 2019-12-02
Fixes #447: qualifier filters on class enumerate/find and instance count

### LINKED ISSUE #447: Find means to show classes filtered by characteristics
In workin with the command line it is often useful to be able to filter classes/classname returned by particular characteristics like associations only all non-association classes.  

The questions are;

1. What would be good criteria for filters
2. How could we incorporte filters into the cmd line

**Proposal one:**
Add an option that would defined a filter that would filter the classes returned by particular class qualifiers.  Thus, it would be able to ask for classes that were assocaitions or not associations, etc.  This does not mean that we would want to use all possible qualifiers as filters so I am not sure today if the idea of something that would filter by filtername/value that allows all qualifiers is the correct solution but it is one possible solution.

Thus the filter could be something like

--filter association=Ture --version >2.45.0

Note that only a few of the qualifier declarations are really used at the class level including: Abstract, Association, Description, Version, UmpPackagePath, Aggradation, Composition, Indication, Deprecated, Experimental, 

Note also that adding this to something like class enumerate causes other issues with the use of class enumerate as a direct access to the CIM/XML enumerate/enumeratenames requests in that if this fiter was to be used, it would mean that the --no opetion could not be used directly since the full class would have to be retrieved to process the filter.

KSchopmeyer:
See commit for details

Added the qualifier filters to:
1. class enumerate
2. class find
3. instance count

Added the following filters:
3 filters (associations, indications, experimental)

Adds a new model with CIMInication and experimental and added a number of tests for each of the modified commands.


pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -67,6 +69,22 @@
                       'the returned class(es). '
                       'Default: Include superclass properties and methods.')]
 
+associations_option = [              # pylint: disable=invalid-name
+    click.option('--association/--no-association',
+                 default=None,
+                 help='Returns only that are either association classes '
+                      '(--association) or are not association classes'
+                      '(--not-association). If the option is not defined '
```
andy-maier: Fix the name of the negative option in the help text: `--not-association` -> `--no-association`.

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -67,6 +69,22 @@
...
+                      'no filtering occurs')]
+
+indications_option = [              # pylint: disable=invalid-name
+    click.option('--indication/--no-indication',
+                 default=None,
+                 help='Returns only that are either indication classes '
+                      '(--indication) or are not indication classes'
+                      '(--not-indication). If the option is not defined '
```
andy-maier: Fix the name of the negative option in the help text: `--not-indication` -> `--no-indication`.

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -67,6 +69,22 @@
                       'the returned class(es). '
                       'Default: Include superclass properties and methods.')]
 
+associations_option = [              # pylint: disable=invalid-name
```
andy-maier: I suggest to name the variable `association_filter_option` instead, for more clarity on what it is.

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -67,6 +69,22 @@
...
+    click.option('--association/--no-association',
+                 default=None,
+                 help='Returns only that are either association classes '
+                      '(--association) or are not association classes'
+                      '(--not-association). If the option is not defined '
+                      'no filtering occurs')]
+
+indications_option = [              # pylint: disable=invalid-name
```
andy-maier: I suggest to name the variable `indication_filter_option` instead, for more clarity on what it is.

tests/unit/test_class_cmds.py
```diff
@@ -525,6 +523,21 @@ class TST_MemberOfFamilyCollection {
...
+      'test': 'innows'},
+     SIMPLE_ASSOC_MOCK_FILE, OK],
+
+    ['Verify class command enumerate with --association filter.',
+     ['enumerate', '--no-association', '--no'],
+     {'stdout': ['//FakedUrl/root/cimv2:TST_FamilyCollection',
+                 '//FakedUrl/root/cimv2:TST_Person'],
+      'test': 'in'},
+     SIMPLE_ASSOC_MOCK_FILE, OK],
+
```
andy-maier: Add test cases for the indication filters.

tests/unit/test_class_cmds.py
```diff
@@ -525,6 +523,21 @@ class TST_MemberOfFamilyCollection {
                  '<METHOD NAME="DeleteNothing" PROPAGATED="false"'],
       'test': 'regex'},
      SIMPLE_MOCK_FILE, OK],
+
+    ['Verify class command enumerate with --association filter.',
+     ['enumerate', '--association', '--no'],
```
andy-maier: Use long option names for clarity: `--no` -> `--names-only`

tests/unit/test_class_cmds.py
```diff
@@ -525,6 +523,21 @@ class TST_MemberOfFamilyCollection {
...
+     {'stdout': ['//FakedUrl/root/cimv2:TST_Lineage',
+                 '//FakedUrl/root/cimv2:TST_MemberOfFamilyCollection'],
+      'test': 'innows'},
+     SIMPLE_ASSOC_MOCK_FILE, OK],
+
+    ['Verify class command enumerate with --association filter.',
+     ['enumerate', '--no-association', '--no'],
```
andy-maier: Use long option names for clarity: `--no` -> `--names-only`

andy-maier:
1. I did also a code review, see my comments.

2. Generally, I like the approach and think we should continue down this path.

3. There seems to be an inconsistency on how `--names-only` shows the result:

    ```
    $ pywbemcli -n assoc class enumerate --no --association
    //FakedUrl/root/cimv2:TST_Lineage
    //FakedUrl/root/cimv2:TST_MemberOfFamilyCollection

    $ pywbemcli -n assoc class enumerate --no --no-association
    //FakedUrl/root/cimv2:TST_FamilyCollection
    //FakedUrl/root/cimv2:TST_Person

    $ pywbemcli -n assoc class enumerate --no
    TST_FamilyCollection
    TST_Lineage
    TST_MemberOfFamilyCollection
    TST_Person
    ```

4. On the remaining TODOs: I suggest to scope this PR to only the operations we discussed in issue #447:
    * class enumerate (with and without --names-only), with filters: --association, --no-association and --indication, --no-indication
    * instance count, with filters: --association, --no-association and --indication, --no-indication

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

tests/unit/test_class_cmds.py
KSchopmeyer: DONE

tests/unit/test_class_cmds.py
KSchopmeyer: DONE

tests/unit/test_class_cmds.py
KSchopmeyer: DONE

PR #461, opened 2019-12-11
Fixes #123: cmd return consistency for commands that return nothing

### LINKED ISSUE #123: Some cmds that return no data on good completion return a statement rather than nothing
Thus, for example, instance delete returns "deleted <instancname>

I propose that those commands that do not return anything should return nothing normally on succusful completion and return a statement (ex. the deleted ...) only in verbose mode.

Thus users who want to use the resulting data in in some way are not distracted by worthless data.

This applies to 

1. create
2. modify
3. delete

Note that we need to be sure that even the enumerates, etc. return nothing like empty return statements. Therefore, any of the commands that can return nothing need the verbose option that says "returned nothing"

KSchopmeyer:
Per the issue, this modifies those commands the return nothing on successful completion to return a message indicating success when --verbose general option is define but return nothing when --verbose is not defined. These return a message with the action (Deleted/Modified, class/instance, and the name of the object.

This applies to class/instance delete, instance modify and the class/instance enumerate, associators, references commands.  These return the fixed message "Return empty' when there are no object returned and --verbose is set

### DIFF
--- a/pywbemtools/pywbemcli/_common.py
+++ b/pywbemtools/pywbemcli/_common.py
@@ -874,6 +874,10 @@ def display_cim_objects(context, cim_objects, output_format=None, summary=False,
         display_cim_objects_summary(context, cim_objects)
         return
 
+    if not cim_objects and context.verbose:
+        click.echo("No objects returned")
+        return
+
     if sort:
         cim_objects = sort_cimobjects(cim_objects)
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -71,6 +71,12 @@ Released: not yet
   any ongoing activities, by setting the env var PYWBEM_SPINNER to 'false',
   '0', or the empty string. This is useful when debugging.  See issue #465.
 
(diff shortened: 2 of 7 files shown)


pywbemtools/pywbemcli/_common.py
```diff
@@ -874,6 +874,10 @@ def display_cim_objects(context, cim_objects, output_format=None, summary=False,
         display_cim_objects_summary(context, cim_objects)
         return
 
+    if not cim_objects and context.verbose:
+        click.echo("Return empty")
```
andy-maier: "Return empty" is a bit on the operational side. How about "No such objects"?

KSchopmeyer:
Updated to fix comments.

pywbemtools/pywbemcli/_common.py
andy-maier: Was DONE by changing to "No objects returned"

PR #466, opened 2019-12-14
Fixes #465: Added support for disabling the Click spinner

### LINKED ISSUE #465: Ability to turn off the spinner
The Click spinner is quite annoying when debugging.

Provide the ability to turn it off.

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -67,6 +67,10 @@ Released: not yet
   '--association' returns classes that are associations and '--no-association'
   returns only classes that are not associations.  See issue # 447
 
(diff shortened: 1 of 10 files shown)


andy-maier:
I think I can accept the argument about complexity.
I'm in favor of option 4: A documented environment variable PYWBEMCLI_NO_SPINNER that when set turns it off.

andy-maier:
We agreed to have an (document) environment variable that turns the spinner off if set.

andy-maier:
The latest update now controls the spinner only via the env var PYWBEMCLI_SPINNER. It is documented only in the development section.
Please re-review.

PR #467, opened 2019-12-14
Added support for pywbem using requests

andy-maier:
This PR now supports both, pywbem versions that use the requests package and pywbem versions that don't. Therefore, we could go forward with it independent of the corresponding pywbem PR pywbem/pywbem#2023.

Ready for review and test (see pywbem/pywbem#2023 for some hints).

### DIFF
--- a/pywbemtools/pywbemcli/pywbemcli.py
+++ b/pywbemtools/pywbemcli/pywbemcli.py
@@ -29,9 +29,8 @@ import click_repl
 from prompt_toolkit.history import FileHistory
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 
-
 import pywbem
-from pywbem import DEFAULT_CA_CERT_PATHS, LOGGER_SIMPLE_NAMES, \
+from pywbem import LOGGER_SIMPLE_NAMES, \
     LOG_DESTINATIONS, DEFAULT_LOG_DESTINATION, LOG_DETAIL_LEVELS, \
     DEFAULT_LOG_DETAIL_LEVEL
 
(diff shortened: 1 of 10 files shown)


KSchopmeyer:
NOTE: This brings up an interesting discussion.  Are we going to force pywbemtools to a pywbem 1.0.0 only release or try to keep it  compatible with both the 0.x and 1.x versions in the same code
**DISCUSSION**

andy-maier:
This question is one of the downsides of major new versions that are incompatible :-)

I think though, that in this case the incompatibility is small enough that both versions of pywbem could be supported in one version of pywbemtools.

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -59,6 +57,26 @@
 # enable -h as additional help option
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
+if getattr(pywbem, 'PYWBEM_USES_REQUESTS', False):
```
KSchopmeyer: I propose comment here about PYWBEM_USES_REQUESTS, and why this code.

andy-maier: DONE - added:
```
# Pywbem 1.0.0 changed to using the 'requests' package, which caused
# differences in how certificate checking is done. Pywbem introduced
# the pywbem.PYWBEM_USES_REQUESTS flag to indicate that. Since
# pywbemtools needs to work with pywbem versions before and after
# 1.0.0, we are checking the flag to generate the correct help message
# for each case.
```

andy-maier:
I have decided for a common description for the `--ca-certs` general option, independent of the installed pywbem version, with the following text:
```
  --ca-certs CACERTS              Certificates to be used for validating the
                                  certificate presented by the WBEM server
                                  during TLS/SSL handshake: FILE: Use the
                                  certs in the specified PEM file; DIR: Use
                                  the certs in the PEM files in the specified
                                  directory; "certifi" (pywbem 1.0 or later):
                                  Use the certs provided by the certifi Python
                                  package; Default: EnvVar PYWBEMCLI_CA_CERTS,
                                  or "certifi" (pywbem 1.0 or later), or the
                                  certs in the PEM files in the first existing
                                  directory from from a list of system
                                  directories (pywbem before 1.0).
 ```

PR #471, opened 2019-12-21
Add tests for some functionality that applies to many commands

KSchopmeyer:
Note that this also corrects two errors found during this testing.

See commit for details


pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -833,15 +835,18 @@ def cmd_instance_modify(context, instancename, options):
 
     If successful, this operation returns nothing.
     """
+    context.spinner_stop()
```
andy-maier: Is there a reason for this spinner_stop() call or is this a remainder of a debug session?

KSchopmeyer: Arrgh hangover from debug session.  Actually should be applied for every case of the if verbose pattern to stop spinner before outputting the message.

KSchopmeyer: Note that this did point out an issue where I was not consisently turning off spinner before those verbose outputs.  **Fixed**

KSchopmeyer:
Issues fixed, passes tests.

PR #477, opened 2020-01-09
Fixes #476: Promoted development status from Alpha to Beta

### LINKED ISSUE #476: Promote development status to Beta
Currently it is Alpha

andy-maier:
Already agreed as per issue #476, no further review needed.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -45,6 +45,9 @@ Released: not yet
 
 * Added documentation for the --version general option.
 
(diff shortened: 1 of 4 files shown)


andy-maier:
This PR fails on Python 3.4. Fixing that requires pywbem 0.16.0 to be released and required as a minimum version. A quick fix would be to pin PyYAML to <5.3 for Python 3.4

PR #486, opened 2020-02-21
pick_from_list bypass user request if only one choice

KSchopmeyer:
Modify the pick_from_list() that normally issues a list of choices for the
user to pick one without issuing the pick list when there is only a
single choice and return that choice.  This bypasses the user having
to pick from a choice of one when there is only one choice.

Adds test for this and a change to the documentation to document this
behavior.

COMMENTS:

This is a change of behavior so not sure it is what we should do. 

 The alternatives are:

1.  Not to do this since it is unexpected unexpected behavior for the user. The fact that the user would not get a selection list in this case is a surprise that changes the normal behavior.
2.  Just do enter rather than the number if there is only one. Again, this is unexpected behavior.
3.  Put out message indicating that we picked one without issuing the list if there is only one and we pick it for them.

NOTE: Andy asked if their was a default on the pick list. No there is not.  There is no easy way to pick defaults for things like instances unless we were to do it from history (i.e. mark as default the last one picked or something like  that). In the case of the connection list, we do the default with the connection table itself by defining one connection that will be the default if the user starts pywbemcli with no connection specified.

Any suggestions are welcome.

### DIFF
--- a/docs/pywbemcli/cmdshelp.rst
+++ b/docs/pywbemcli/cmdshelp.rst
@@ -190,14 +190,14 @@ Help text for ``pywbemcli class`` (see :ref:`class command group`):
       -h, --help  Show this message and exit.
 
     Commands:
-      associators   List the classes associated with a class.
-      get           Get a class.
       enumerate     List top classes or subclasses of a class in a namespace.
-      tree          Show the subclass or superclass hierarchy for a class.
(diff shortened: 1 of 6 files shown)


andy-maier:
This one was not marked "review needed". I reviewed it and have approved. Is it ready to be merged?

PR #487, opened 2020-02-24
Add common  class NocaseList for use with ks/shrub-add-nocaselist

KSchopmeyer:
Required before ks/add_shrub_display pr

Adds a new common funct and it tests that allows simple tests for no case test of string in list of strings. This is primarily for use with the shrub display changes but is probably more general that that so we put it into _common.py

### DIFF
--- a/pywbemtools/pywbemcli/_common.py
+++ b/pywbemtools/pywbemcli/_common.py
@@ -1006,13 +1006,9 @@ def format_keys(obj, max_width):
 
 class NoCaseList(object):
     """
-    This class is a dervivative of lists the simplifies working with
-    some of the CIM object attributes that are case insensitive such as
-    class names, roles, etc.
-
-    # TODO: Future; Possibly optimize by setting strs to lower in constructor
-    Right now the usage is primarily for small lists to this implementation
-    does not cause excessive overhead.
+    This class simplifies working with lists of strings that are to be tested
+    case-insenstive. This simplifies working with  the CIM object attributes
+    that are case insensitive such as class names, roles, etc.
 
     NOTE: This code does not handle None either for items in the list or for
     the str parameter in __contains__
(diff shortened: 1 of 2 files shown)


pywbemtools/pywbemcli/_common.py
```diff
@@ -1004,6 +1004,63 @@ def get_wbemurikeys(obj):
...
+
+    def __contains__(self, str):
+        """
+        Implement 'in'
+        """
+
+        assert str is not None
+        return str.lower() in (n.lower() for n in self.str_list)
+
+    def add(self, strs):
```
andy-maier: In a Python list, this method is named "append". Suggest to use that name here, too.

pywbemtools/pywbemcli/_common.py
```diff
@@ -1004,6 +1004,63 @@ def get_wbemurikeys(obj):
     return wbem_uri_keys
 
 
+class NoCaseList(object):
+    """
+    This class is a dervivative of lists the simplifies working with
```
andy-maier: Typo: dervivative

pywbemtools/pywbemcli/_common.py
```diff
@@ -1004,6 +1004,63 @@ def get_wbemurikeys(obj):
...
+    some of the CIM object attributes that are case insensitive such as
+    class names, roles, etc.
```
andy-maier: I find the term "derivative of lists" confusing. The class is not derived from the Python `list` type, nor is its behavior compatible to a Python list.

pywbemtools/pywbemcli/_common.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_common.py
KSchopmeyer: Removed that confusing statement completly.  Just says "This class simplifies working with lists of strings that are to be tested  case-insenstive." DONE

PR #506, opened 2020-03-10
Fixes #354: Command-specific default output format and improved output format validation

### LINKED ISSUE #354: Issues with specifying and defaulting output format
Note: Updated to reflect the discussion below.

Some commands ignore the specified output format without further notice, and change it to something else. An example is the `class find` command which honors table output formats (and seemingly also the txt format) but ignores all other output formats, using the txt format instead.

This is bad for the user, because they do not get any feedback that the output format they specified is not supported at all.

I suggest to change this as follows:
1. Every command has a reasonable default output format.
2. The specified output format is never changed to something else, but used as specified.
3. If the specified output format is not supported, the command fails, saying that.

KSchopmeyer:
Change default for output format to None and add test for each command  that uses output_format to validate that it can handle the cmd line output-format value  provided.

### DIFF
--- a/docs/pywbemcli/cmdshelp.rst
+++ b/docs/pywbemcli/cmdshelp.rst
@@ -146,12 +146,13 @@ Help text for ``pywbemcli``:
                                       PYWBEMCLI_DEFAULT_NAMESPACE, or root/cimv2.
 
       -o, --output-format FORMAT      Output format for the command result. The
-                                      specified format may be overriden since not
-                                      all formats apply to all result data types.
-                                      FORMAT is a table format
+                                      default and allowed output formats are
+                                      command specific. The default output_format
+                                      is None so that each command selects its own
+                                      default format. FORMAT is either a table
+                                      format:
                                       [table|plain|simple|grid|psql|rst|html] or
-                                      object format [mof|xml|repr|txt]. Default:
-                                      simple.
+                                      CIM object format: [mof|xml|repr|txt].
 
       -l, --log COMP[=DEST[:DETAIL]],...
                                       Enable logging of the WBEM operations,
--- a/docs/pywbemcli/generaloptions.rst
+++ b/docs/pywbemcli/generaloptions.rst
@@ -803,8 +803,9 @@ pull operations are not supported in the WBEM server:
 Output formats
 ^^^^^^^^^^^^^^
 
-Pywbemcli supports various output formats for the command result. The output
-format can be selected with the ``--output-format``/``-o`` option.
+Pywbemcli supports multiple output formats to present command results. The output
+format can be selected with the ``--output-format``/``-o`` option. The allowed
+output formats are different for the various command groups and commands.
 
 The output formats fall into three groups:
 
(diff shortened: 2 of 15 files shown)


docs/pywbemcli/cmdshelp.rst
```diff
@@ -51,6 +51,7 @@ Help text for ``pywbemcli``:
                                       --mock-server options, since each defines a
                                       WBEM server. Default: EnvVar PYWBEMCLI_NAME,
                                       or none.
+
```
andy-maier: Please use Click 7.,0 in the python env where you create this file. We should not mix different things into one PR.
The transition to the changes Click 7.1 introduces (e.g. to this file) should be done in a different PR.

docs/pywbemcli/cmdshelp.rst
```diff
@@ -190,14 +209,14 @@ Help text for ``pywbemcli class`` (see :ref:`class command group`):
...
-      tree          Show the subclass or superclass hierarchy for a class.
-      references    List the classes referencing a class.
+      get           Get a class.
+      delete        Delete a class.
       invokemethod  Invoke a method on a class.
+      references    List the classes referencing a class.
+      associators   List the classes associated with a class.
       find          List the classes with matching class names on the server.
-      delete        Delete a class.
+      tree          Show the subclass or superclass hierarchy for a class.
```
andy-maier: This reordering of commands should not be there. Please make sure you are using the latest master branch as a basis for this file.

docs/pywbemcli/cmdshelp.rst
KSchopmeyer: The master is now usig click 7,1.1 and bypassing 7.1.

docs/pywbemcli/cmdshelp.rst
KSchopmeyer: DONE

docs/pywbemcli/cmdshelp.rst
andy-maier: No. Your python env is set up with Click 7.1.1. The requirements file only requires Click>=7.0, excluding 7.1, so both 7.0 and 7.1.1 are allowed. What I am suggesting is that we manually install Click 7.0 in our python envs for the time being, so that the changes driven by Click 7.1.1 are introduced separately.

docs/pywbemcli/cmdshelp.rst
andy-maier: The file is still showing the reordered commands. If you want to make sure it is re-created, delete it and then run make build or make builddoc.

docs/pywbemcli/cmdshelp.rst
andy-maier: I have PR #511 under way that provides the basis to use Click 7.1.1.It is easier for this PR to wait for or rebase on that PR.

docs/pywbemcli/cmdshelp.rst
andy-maier: The order of commands depends on whether py2 or py3 is used to generate the file. Issue #510 has been opened for that. I suggest that for now, we use py2 to generate this file in any PRs.

docs/pywbemcli/cmdshelp.rst
KSchopmeyer: Now I understand.

andy-maier:
Karl, I have resolved this PR and will merge it.

PR #507, opened 2020-03-11
Fixes #505: Added support for debugging via --pdb option

### LINKED ISSUE #505: Improve ability to debug pywbemcli commands
Debugging issues like those reported in #504 is pretty hard especially since the test driver invokes pywbemcli in a sub-shell. Some support is needed for debugging that can be controlled from the testcases.

andy-maier:
See commit message.

### DIFF
--- a/pywbemtools/pywbemcli/pywbemcli.py
+++ b/pywbemtools/pywbemcli/pywbemcli.py
@@ -223,13 +223,21 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 @click.option('-v', '--verbose/--no-verbose',
               default=None,
               help='Display extra information about the processing.')
+@click.option('--pdb', is_flag=True,
+              # defaulted in code
+              envvar=PywbemServer.pdb_envvar,
+              help='Pause execution in the built-in pdb debugger just before '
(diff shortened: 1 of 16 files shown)


pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -223,13 +223,20 @@
 @click.option('-v', '--verbose/--no-verbose',
               default=None,
               help='Display extra information about the processing.')
+@click.option('--pdb', is_flag=True,
+              # defaulted in code
+              envvar=PywbemServer.pdb_envvar,
+              help='Break in the debugger before executing the command. '
```
KSchopmeyer: Change the help to "the debugger will pause execution just before executing the command"

andy-maier:
pdb is the name of the built-in debugger in the standard python executable, so I think we should use it. We may want to have a link to the python pdb docs in our docs, though. I agree with updating the option help text, see my response above.

pywbemtools/pywbemcli/pywbemcli.py
andy-maier: DONE, using this help text: "Pause execution in the built-in pdb debugger just before executing the command within pywbemcli."

PR #519, opened 2020-03-14
Fixes #518: Added support for adjusting help output width to terminal width

### LINKED ISSUE #518: Width of help output is constant and ignores terminal width
Invoking the `--help` option at any command level including globally creates output that is alway 78 characters wide, independent of the current width of the terminal. Click claims that it adjusts the output to the terminal width, but as of click 7.1.1 that is not the case.

The width can be set explicitly via Context attribute `terminal_width`.

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -143,6 +143,9 @@ Released: not yet
 * Test: Added support for testing on Python 3.8 in Travis, Appveyor and Tox.
   (See issue #494)
 
+* Added support for adjusting the width of any help output to the terminal
+  width. (See issue #518)
+
 **Cleanup:**
 
 * Test: Enabled Python warning suppression for PendingDeprecationWarning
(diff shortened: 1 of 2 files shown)


andy-maier:
Waiting for PR #520 to be merged.

andy-maier:
Click does not honor the terminal width at all, if context.terminal_width is None, not even initially. I think this is a bug in Click, but I did not investigate further after I saw how convoluted the logic there is.

The way we specify the terminal width only works initially, so unless something changes in Click, dynamic changes of terminal width will not be honored.

My suggestion is to go with this PR as it is, because it improves the situations for most cases. We can leave a dynamic reaction to the terminal width for the future.

KSchopmeyer:
I think we could make the more dynamic change work (how many hours have I spent digging around in click) but not a big issue.  A agree and approve this pr.

andy-maier:
Update: It turns out that with the help texts now depending on the terminal width, that we need to be careful when re-creating the docs/pywbemcli/cmdshelp.rst file. Experiments show that the terminal width that needs to be used to re-create that file in the same formatting as currently, is 78.

PR #523, opened 2020-03-15
Fixes #510: cmds order display py dependent

### LINKED ISSUE #510: Help shows commands in different order on py2 vs py3
When comparing the generated pywbemcli/cmdshelp.rst file (on master) when generated on Python 2.7 and when generated on Python 3.7 they are different and the difference is the order of the subcommands listed in the command help.

The implementation is supposed to honor the order of the commands in the source file, so something in that implementation must be dependent on the python version.

KSchopmeyer:
Fixes the issue where the order of commands listed in the --help for command groups was different in python 2 than in python 3.

See commit for details.


andy-maier:
This PR depends on PR #450 which has not been rolled back. I suggest not to roll these two back.

PR #531, opened 2020-03-19
Fixes #530 - Issue with use_pull general option

### LINKED ISSUE #530: Error in handling of use_pull option
The option is mapped internally incorrectly. This is clearly a bug since in impacts ability to use 'either' option with servers that do no have pull.

KSchopmeyer:
Corrects issue(bug) where invalid internal representation of the option was
passed internally.

Marked for rollback since it is really a bug and there is one server in the snia testbed that fails.

Note that this change currently does not have a corresponding test.  I created pywbem issue # 2125 to extend pywbem mocker and tested a subset of that code with pywbemcli to confirm fix.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -74,6 +74,9 @@ Released: not yet
   NocaseDict, _to_unicode, _ensure_unicode, _format from pywbemcli. (See
   issue #489)
 
+* Corrected issue with use-pull general option that causes issues with using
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
We have an issue with the cmdshelp.rst file changing and that issue caused me to accidently commit cmdshelp.rst to this PR. It has now been removed.

PR #533, opened 2020-03-20
Fixes #529,#534: use-pull internally only partly implemented

### LINKED ISSUE #529: Handling of --use-pull incomplete
This variable is not included in the "connection show" and the tests are marked skipped

KSchopmeyer:
use-pull was never incorporated into the reports, lists, etc. and was left as a "transient" option in that it was not really attached to any particular connection.

1. Added to connection list and connection show which meant making it part of the PywbemServer class with its own property rather than just passed in the context in an interactive session.

2. made persistent and part of the persistent file per issue #534 (DISCUSSION is really what is discussed in that issue)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -192,6 +192,11 @@ Released: not yet
 * Removed a circumvention for a pywbem bug related to colons in WBEM URIs
   that was fixed in pywbem 0.13.0. (See issue #131)
 
+* Added the general option `--use-pull` to the the PywbemServer() class so that
+  it is persisted in the connection file and to the display of connection
+  information (`connection show` and `connection list`). This means that
+  `--use-pull` can now be set for a particular server permanently.(See issues
+  #529 and #534).
 
 **Known issues:**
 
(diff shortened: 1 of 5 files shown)


KSchopmeyer:
Karl will resubmit with cmdshelp.rst removed.  We noted that this is now terminal width sensitive.

andy-maier:
Also, please wait until PR #531 has been merged and rebase on master.

docs/changes.rst
```diff
@@ -187,6 +187,9 @@ Released: not yet
 * Removed a circumvention for a pywbem bug related to colons in WBEM URIs
   that was fixed in pywbem 0.13.0. (See issue #131)
 
+* Corrected issue with use-pull general option that causes issues with using
+  the 'either' option with servers that do not have pull. (See issue #530)
```
andy-maier: Add statement that this adds a "use_pull" argument to PywbemServer() (it is incompatible to add that in the middle).

KSchopmeyer: DONE

KSchopmeyer: NOTE: removed duplicated comment about issue #530 (it was in bugs and enhancements) and added one for this pr.

andy-maier:
I have updated the PR to apply my two suggestions from the previous comment. Up for discussion.

PR #543, opened 2020-03-23
Fixes issue # 537 Hide the text output from a number of tests.

KSchopmeyer:
Since changing the use of tools like simplified_test_function to allow
capturing text output (i.e. using the capsys fixture is signicantly
difficult, we found a simple mechanism, simply mock the click.echo so
that the output is never generated.

Changed the tests in common that output text so that the test output is
significantly cleaner now.

While this works for real unit tests, it does not work for functional tests (i.e calls to pywbemcli)

### DIFF
--- a/tests/unit/test_common.py
+++ b/tests/unit/test_common.py
@@ -662,43 +662,6 @@ def test_split_str(testcase, input_str, delimiter, exp_rtn):
     assert act_result == exp_rtn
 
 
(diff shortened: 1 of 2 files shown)


KSchopmeyer:
Note that the first commit produced a test with thousands of extra lines in some of the pywbemcli files.  I resubmitted and it appears clean now

PR #560, opened 2020-03-30
Fixes #446: Clean up the Help Usage: statement to reflect General and command options

### LINKED ISSUE #446: Change help to show command options after command arguments
I recommend changing the help for commands to show command options after command arguments. When you make an error in the command options the error message is more informative.

See the example below:
```
pywbemcli> instance enumerate --pl StatisticTime,TotalIOs, KBytesTransferred CIM_BlockStorageStatisticalData
Usage: pywbemcli  instance enumerate [COMMAND-OPTIONS] CLASSNAME
Try "pywbemcli  instance enumerate -h" for help.

Error: Got unexpected extra argument (CIM_BlockStorageStatisticalData)
```

```
pywbemcli> instance enumerate CIM_BlockStorageStatisticalData --pl StatisticTime,TotalIOs, KBytesTransferred
Usage: pywbemcli  instance enumerate [COMMAND-OPTIONS] CLASSNAME
Try "pywbemcli  instance enumerate -h" for help.

Error: Got unexpected extra argument (KBytesTransferred)
```

The usage help should read:
```
Usage: pywbemcli  instance enumerate CLASSNAME [COMMAND-OPTIONS]
```

KSchopmeyer:
Added one more separate commit to change CMD-OPTIONS to COMMAND-OPTIONS and the help options section to Command Options.  

The end result is:

>pywbemcli --help

Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS] [COMMAND-OPTIONS]

pywbemcli class --help

Usage: pywbemcli [GENERAL-OPTIONS] class COMMAND [ARGS] [COMMAND-OPTIONS]

>pywbemcli class find --help

Usage: pywbemcli [GENERAL-OPTIONS] class find CLASSNAME-GLOB [COMMAND-OPTIONS]

> pywbem repl --help

Usage: pywbemcli [GENERAL-OPTIONS] repl

Note that this is probably pushing the limits we want to go with modifying click in that we override two methods in click itself to do this.  See _click_extensions.py


andy-maier:
Please rebase on master, that should also trigger the builds properly on all envs.

docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -175,7 +175,7 @@ user):
 
 .. code-block:: text
 
-    pywbemcli> [GENERAL-OPTIONS] [COMMAND-GROUP] COMMAND [COMMAND-OPTIONS] [ARGS]
+    pywbemcli> [GENERAL-OPTIONS] [COMMAND-GROUP] COMMAND [ARGS] [CMD-OPTIONS]
```
andy-maier: `COMMAND-GROUP` is mandatory, so the square brackets around it should be removed.

docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -232,7 +232,7 @@ list of the supported command groups and commands without command group.
 
     $ pywbemcli
     pywbemcli> --help
-    Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS]...
+    Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS] [CMD-OPTIONS]
```
andy-maier: Insert `COMMAND-GROUP` before `COMMAND`.

docs/pywbemcli/cmdshelp.rst
```diff
@@ -14,7 +14,7 @@ Help text for ``pywbemcli``:
 
 ::
 
-    Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS]...
+    Usage: pywbemcli [GENERAL-OPTIONS] COMMAND [ARGS] [CMD-OPTIONS]
```
andy-maier: Insert `COMMAND-GROUP` before `COMMAND`.

docs/pywbemcli/cmdshelp.rst
```diff
@@ -867,7 +867,7 @@ Help text for ``pywbemcli help`` (see :ref:`help command`):
 
 ::
 
-    Usage: pywbemcli help [OPTIONS]
+    Usage: pywbemcli [OPTIONS] help
```
andy-maier: Can `OPTIONS` be renamed to `GENERAL-OPTIONS` ?

docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -232,7 +232,7 @@ list of the supported command groups and commands without command group.
...
     . . .
 
     Options:
```
andy-maier: We currently have a section heading "Options:" in the help text that is ambiguous as to whether it lists the `GENERAL-OPTIONS` or the `CMD-OPTIONS`. If technically possible, I think it would be great if that section heading could be changed to "General options:" for the top level help, and to "Command options:" for the lower level help.
PS: I'm just adding this here because there is an "Options:" section shown.
 Suggestion:
* Merge `[COMMAND-GROUP] COMMAND` in usage to just `COMMAND` which is consistent with the section title "Commands:".
* Change title of "Options:" section to "General options:".

docs/pywbemcli/cmdlineinterface.rst
andy-maier: In the actual help, we no longer have `[COMMAND-GROUP]`in the usage lines, and i think that is the right direction. It should be removed in this file as well (this file is the only one that still contains that). Occurs multiple times. Some words need to be added that explain that `COMMAND` is sometimes directly a command and sometimes a command group that has sub-commands.

KSchopmeyer: Fixed.  Removed [COMMAND=GROUP] and defined what command means.

docs/pywbemcli/cmdshelp.rst
andy-maier: The change addresses the second comment but not the first. That's actually ok because we don't show it in other cases that have only the --help option, either. DONE.

docs/pywbemcli/cmdlineinterface.rst
andy-maier: **Discussion point:** How do we go about this?

andy-maier: I have described that discussion point in more detail in issue #446.

andy-maier:
This PR should wait for resolution of the discussion point in issue #446.

KSchopmeyer:
I added the code to fix your reopen of the issue #446.  The title to the options section now reflects whether the options are General Options or Cmd Options. Note that the cmd line pywbemcli class -h is probably confusing in that there is only one options (--help) but I listed it as a Cmd Option.

Also, I would have preferred to change CMD-OPTION to COMMAND-OPTION but that longer text causes a number of usage lines to overflow 80 characters.

andy-maier:
I like it!

On CMD-OPTION vs COMMAND-OPTION: I think COMMAND-OPTION as a meta variable and "Command Options:" as section title is much more intuitive, also since in the text we talk about "command options" and it is more intuitively clear that they belong to the COMMAND.

I would value these benefits higher than the 80 char limitation. Let me know what you think about that.

KSchopmeyer:
I would have liked to keep in the 80 col limit for the usage line (the change to COMMAND-OPITONS means that about 9 usage statements overflow instead of the original 2) but it was not a big issue.
The change has been made in a separate commit to use COMMAND-OPTIONS as the tag and the title of Command Options: for the help options section

KSchopmeyer:
Fixed issue with mispelling COMMANDD id docs.

PR #564, opened 2020-03-31
[WIP] Fixes #182: Capture output with simplified_test_function

### LINKED ISSUE #182: No way to capture output with simplified_test_function
While this is not an issue with pywbem it is with pywbemtools because so much of the function results are actually stdout.  In pytest the fixture capsys must be defined on the test function to enable stdout capture and this does fit into the fixed structure of the simplified_test_function definition.

andy-maier:
Very experimental at this point.

### DIFF
--- a/tests/unit/pytest_extensions.py
+++ b/tests/unit/pytest_extensions.py
@@ -11,6 +11,7 @@ import functools
 from collections import namedtuple
 import six
 import pytest
+import wrapt
 if six.PY3:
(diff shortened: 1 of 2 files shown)


andy-maier:
I have given up on the approach to extend the simplified_test_function decorator to support additional arguments. PR #625 addresses issue #182 by removing the use of the simplified_test_function decorator on the test function in question.

PR #576, opened 2020-04-06
Fixes #572: Table output for connection show

### LINKED ISSUE #572: Connection show output does not match pywbemcli known output-format types
Connection show is a piece of work I did initially just for me in testing but survived.  It is a sort of text form:

name: mock1 (current)
  server: None
  default-namespace: root/cimv2
  user: None
  password: None
  timeout: 30
  verify: True
  certfile: None
  keyfile: None

That is neither table, nor CIM object but not really raw text. 

I propose to at least make a table option for this.

DISCUSSION: Should the table output be the only output or just when table is selected???

KSchopmeyer:
Adds the table output as an option when table output is selected.

This means that the old format is still the default.

DISCUSSION: Should we drop the old format and go with just the table format? (Andy: Yes).


andy-maier:
I think we should drop the old format

andy-maier:
Hi Karl. I decided to go ahead and update this PR. I squashed your commits into one, rebased to master, resolved the conflicts, and added a commit that removes the old format.
Please review.

If you decide to update the PR, it is probably best if you remove your local branch and check it out freshly.

KSchopmeyer:
Downloaded and reviewed.  Added one minor comment to changes.rst as a new commit and APPROVE

PR #600, opened 2020-05-06
Fixes #594: Simplify output and implement TEXT output group

### LINKED ISSUE #594: Simplify the output of single value server commands
The `server brand` and `server interop` commands show only a single value but add quite some words around that. This could be simplified to showing only the output value.

Examples (from quickstart docs, not verified with latest code):

```
pywbemcli> server brand

Server Brand:
WBEM Server Brand
-------------------
OpenPegasus
```
```
pywbemcli> server interop

Server Interop Namespace:
Namespace Name
----------------
root/PG_InterOp
```

KSchopmeyer:
Implements a new output format group (TEXT) and modifies two commands to
use this group (server interop, and server brand). This removes the
table option for these commands.

It also implements the text output format as an optional format for
namespaces so that the namespace list appears as a comma-separated
line of text.  In this case the default is table and the optional
choice is text.

NOTE: The pr also updates the changes.rst file to add the section
for version 0.7.0 and moves the few changes we have made into
that new section.


andy-maier:
Fails tests.

docs/changes.rst
```diff
@@ -22,6 +22,30 @@ Released: not yet
 
 **Enhancements:**
 
+* Enhance output formats to allow an additional format group TEXT with
+  a single format ``text``. This format outputs the command result as a
+  text string to the console and is use for simple commands like
+  ``server interop`` that only output one piece of data. (see issue #594)
+
```
andy-maier, line 29: The following four change items have been moved from the 0.6.0 release into the 0.7.0 release. I believe this was by accident. Please double check and undo the move.

docs/pywbemcli/generaloptions.rst
```diff
@@ -927,15 +931,15 @@ table = table|plain|simple|grid|psql|rst|html
 +----------+---------------+----------+----------------+--------------------------------------------+
 |          |  get          | 'mof'    | table          |                                            |
 +----------+---------------+----------+----------------+--------------------------------------------+
-|server    |  brand        | 'simple' | table          |                                            |
+|server    |  brand        | 'text  ' | text           | Alternate is table format                  |
```
andy-maier, line 934: Remove trailing blanks in `'text. '`

docs/changes.rst
KSchopmeyer: Actually two of them (from issue #594 and 598 are part of v 0.7.0 and I accidently put them into the section for 0.6.0 when I added them.  The other two were my mistake when I was moving the two that I had placed wrong.

**DONE**

docs/pywbemcli/generaloptions.rst
KSchopmeyer: DONE

PR #613, opened 2020-06-05
Fixes #504: Changes to pywbemcli required for pywbem 1.0.0

### LINKED ISSUE #504: Support for pywbem 1.0.0
The master branch of pywbem (i.e. the upcoming pywbem 1.0.0) has changed some behavior so that some of the pywbemtools unit tests fail with it.

Running `make test` on the master branch of pywbemtools reports 50 failures when using pywbem from master (on MacOS with Python 2.7). This succeeds when using pywbem 0.16.0 from Pypi.

Among the failures are:
* ModelError: Namespace class could not be determined (tried ['CIM_Namespace', '__Namespace'])
* Unexpected WBEM URI (port is always shown in pywbem 1.0)
* pywbem 1.0 has more strict error checking when processing mock files
* Invalid port number in server URL is handled differently.
...

KSchopmeyer:
Issue #504, Changes to pywbemcli required for pywbem 1.0.0compatibility.

NOTE: This currently tests against the latest release

A number of changes were required to make pywbemcli compatible with both
pywbem pre version 1.0 and the incompatible differences in pywbem 1.0.0.

These are issues in the tests and not in pywbemcli itself. 

Changed the tests so that we create variables PYWBEM_0 and PYWBEM_1 in
a common file and use these variables to modify tests.

Modified a number of tests to account for the fact that the returned
url for many tests now includes the port in addition to the host.

Modified the invokemethod test code to create the test provider
Since the pywbem 1 invokemethod server responder is completely different
that the pywbem 0 version, two different test driver python files are
for the invoke method tests are defined.

1. Changed name of simple_mock_invokemethod.py to
simple_mock_invokemethod_pywbem_0.py

2. Created simple_mock_invokemethod_pywbem_1.py as the user-defined
provider for InvokeMethod tests.

3. Added if to each test file that tests invokemethod to use the correct
file.

4. Created new test for testing url invalid port for pywbem 1.


andy-maier:
Since a beta version pf pywbem 1.0.0 has not been released yet, the CI tests use pywbem 0.17.2 for the test with latest package versions. I verified that the tests succeed (on py38) when using the current master branch of pywbem.

tests/unit/test_class_cmds.py
```diff
@@ -1541,72 +1532,3 @@ class command that can be tested.  This includes:
         """
         self.command_test(desc, self.command_group, inputs, exp_response,
                           mock, condition, verbose=False)
-
```
andy-maier, line 1544: Why have these tests been removed?

KSchopmeyer: Sorry, forgot to add comment.  Turns out that they were the original tests I wrote when starting the project and added nothing to the overall testing except to make it harder to isolate to a single test.  All that functionality is incorporated into the set of test cases.   I had meant to drop them long ago and forgot.  Should have made it a separate PR.

andy-maier: ok with that.

andy-maier:
Karl, I accepted your explanation for the removal of the two test cases, but the 3 test failures against pywbem master still need to be resolved, so I am putting this PR back to update needed.

KSchopmeyer:
The issues were actually in pywbem and pywbem pr # 2529 fixes those issues.  The pywbem_server test issue was a test that passed with the old network software but failed with requests because we were using a bogus certificate in the test.   Modified test and created another with the bogus cert where the error is caught.

andy-maier:
After Karl's comment in the Slack channel, I retried after double checking that the pywbem package was installed in edit mode and that the pywbem working directory contains the master branch, and that the master branch was pulled, and voila, the error was gone.
Considering this resolved.
Approved and merged.

PR #625, opened 2020-06-14
Fixes #182: Added output checking in test_print_insts_as_table()

### LINKED ISSUE #182: No way to capture output with simplified_test_function

andy-maier:

### DIFF
--- a/pywbemtools/pywbemcli/_cimvalueformatter.py
+++ b/pywbemtools/pywbemcli/_cimvalueformatter.py
@@ -184,7 +184,7 @@ def _scalar_value_tomof(value, type, indent, maxline, line_pos=0, end_space=0,
         val = six.text_type(value)
         return _mofstr(val, indent, maxline, line_pos, end_space, avoid_splits)
     elif type == 'reference':
-        val = value.to_wbem_uri()
+        val = value.to_wbem_uri(format='standard')
         return _mofstr(val, indent, maxline, line_pos, end_space, avoid_splits)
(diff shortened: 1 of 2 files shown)


andy-maier:
The reason this fails is that the order of keys in `CIMInstanceName.to_wbem_uri()` depends on the Python version, because `_scalar_value_tomof()` uses the default format "standard".

Possible options:
* Have `_scalar_value_tomof()` use the format "canonical". This guarantees that the keys are sorted, but it also converts class names and key names to lower case.
* Add a new format "ordered" to `CIMInstanceName.to_wbem_uri()` (in pywbem) that guarantees the key order but does not convert class names and key names to lower case.

andy-maier:
We agreed to go for the second option (new format "sorted").

andy-maier:
It fails only on native Windows:
https://ci.appveyor.com/project/KSchopmeyer/pywbemtools/builds/33540603/job/w6jrugbtscxxax90#L1360
in `click.echo()` in combination with the `capsys` fixture.

andy-maier:
We agreed to skip the failing test on Windows.

andy-maier:
The latest update skips the tests on Windows (due to the Click issue) and the test with keys when pywbem is earlier than 1.0.0b1 (due to the WBEM URI key sorting)

PR #636, opened 2020-06-20
Fixes #634: Added value-mapped string to properties in instance table output

andy-maier:
**Update: For latest status, see the last comments of this PR**

Ready for initial use and review.

The use of valuemap/values is always enabled when displaying instances in a table format and when the WBEM server supports class operations, so there is no option to enable or disable it.
The properties that are integer-typed and that have a valuemapping defined or inherited in the creation class of the instance are displayed with the Values value in parenthesis after the integer value.

Example with the changes from this PR, that added a "gender" property with value-mapped values for male and female:
```
$ pywbemcli -m tests/unit/simple_assoc_mock_model.mof -o table instance enumerate TST_Person
Instances: TST_Person
+------------+------------+------------------+-----------+
| name       | gender     | secondProperty   | counter   |
|------------+------------+------------------+-----------|
| "Gabi"     | 1 (female) |                  |           |
| "Mike"     | 2 (male)   |                  |           |
| "Saara"    | 1 (female) |                  |           |
| "Sofi"     | 1 (female) |                  |           |
| "Gabisub"  | 1 (female) | "four"           | 4         |
| "Mikesub"  | 2 (male)   | "one"            | 1         |
| "Saarasub" | 1 (female) | "two"            | 2         |
| "Sofisub"  | 1 (female) | "three"          | 3         |
+------------+------------+------------------+-----------+
```

Discussion points:
* Do we need an option to enable or disable the value-mapping output?
* If the WBEM server does not support class operations, do we expect a warning or just silently not showing value-mapping output?
* Do we also need it for string-typed properties?
Mike has provided his views on these discussion points further down.

Limitation: This PR limits the support to scalar elements. Array elements are displayed as normal, i.e. without the value-mapped value in parenthesis.

PR #648, opened 2020-06-25
[WIP] Fixes #634: Added value-mapped value to properties in instance table output

### LINKED ISSUE #634: Need an output option for displaying Value/Valuemap information
I would like an option on output for displaying VALUE information instead of, or in addition to, VALUEMAP information. SNIA uses the Dedicated property on ComputerSystem. But you need to have a cheat sheet handy to tell you Dedicated="24" is a NAS Head, Dedicated="25" is a Self-contained NAS and Decidated="16" is a File Server. Similarly, you need a cheat sheet to tell you that HealthState="5" means "OK".
For demo's I can get the class to show the cheat sheet, but Dedicated has over 40 values. 24 and 25 are pretty much in the middle. Manually coordinating what a number corresponds to is not all that easy.

andy-maier:
TODO:
* Add support on `pywbem.ValueMapping` for array elements. See issue https://github.com/pywbem/pywbem/issues/2304


andy-maier:
Abandoned this PR, because the support for array-typed properties could be added to PR #636.

PR #649, opened 2020-06-25
Fixes #637: MOF compiler-display-compile-errors

### LINKED ISSUE #637: MOF compile error in mock support no longer displays compiler error
The MOF compiler in pywbem changed to no longer display the error message by default. So far, pywbemcli relied on the error being displayed by default. That needs to be adjusted in pywbemcli to accomodate the new behavior of the MOF compiler in pywbem.

KSchopmeyer:
Modify code to display compile errors both for pywbem 1 and previous
versions of pywbem.

For pywbem 1 it displays the exceptions.  For earlier versions, it
notes that the error preceeds the exception text.

NOTE: We not catch all Error exceptions to keep a common exception
name for all versions.


pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -316,14 +318,23 @@ def build_repository(conn, server, file_path_list, verbose):
...
+                except Error as er:
                     # Abort the entire pywbemcli command because the
                     # MOF compilation might have caused inconsistencies in the
                     # mock repository.
-                    click.echo(
-                        "Mock MOF file '{}' failed compiling (see above)".
-                        format(file_path),
-                        err=True)
+
+                    if not PYWBEM_VERSION.startswith("0"):
```
andy-maier, line 326: I would prefer a more robust and also more self-explanatory test for the pywbem version being at least 1.0.0:
```
import pywbem
import packaging.version

PYWBEM_VERSION = packaging.version.parse(pywbem.__version__)

. . .
if PYWBEM_VERSION.release >= (1, 0, 0):
```
That allows also checking whether it is a beta version or dev version, if needed:
```
if PYWBEM_VERSION.is_devrelease:  # a dev release
if PYWBEM_VERSION.is_prerelease:  # an alpha or beta or rc release
```

KSchopmeyer: Changed code.  Note that I also changed code in tests to use the parse module.

andy-maier: OK

PR #651, opened 2020-06-28
Fixes #630: Deprecated Python 2.7 and 3.4 support

### LINKED ISSUE #630: Mark Python 2.7 and 3.4 support as deprecated

andy-maier:
See commit message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -637,9 +637,9 @@ safety_$(pymn).done: develop_$(pymn).done Makefile minimum-constraints.txt
 test: develop_$(pymn).done $(doc_utility_help_files)
 	@echo "makefile: Running unit and function tests"
 ifeq ($(PLATFORM),Windows_native)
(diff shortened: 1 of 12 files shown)


andy-maier:
If we have the default to not show them, they will never be shown.

Look at pip, they show the warning about the Python release by default and provide an option `--no-python-version-warning` to turn them off. I modeled it after that approach, plus the env var.

There is nothing wrong if the warning is shown every time, IMO. It even goes to stderr, so if someone ventures to process the stdout somehow, it will not be part of the stdout.

andy-maier:
Needed to circumvent an issue on Windows where the syntax:
```
> set PYWBEMCLI_NO_DEPRECATION_WARNINGS=false & pywbemcli ...
```
caused the envvar to have a trailing blank which causes Click to fail after checking its valid values. I have created https://github.com/pallets/click/issues/1605 along with a PR to fix it in Click, and have updated the Makefile to circumvent the issue by removing the blank between the value and the '&'.

In addition, I inverted the option and envvar and added the complementary option. The default is still to show the warnings by default, but it is now very eay to change the default.

DISCUSSION on default behavior.

KSchopmeyer:
We reviewed and agreed to the current code as the solution. Karl wants to do one more review before approving.

andy-maier:
We agreed that the default should be to display the warnings.

Review points should be the means to disable the warnings.

PR #653, opened 2020-06-29
Refactor tree display code

KSchopmeyer:
Refactor the code in the _displaytree.py module to make it more comprehensible, faster, and correct an outstanding issue where we were not always displaying the tree in the same order and ignoreing possible issues of classname case-independence

This also adds tests for boundary conditions of the "class tree" command

This pr does not have a defined issue since it does not change external behavior.

It was tested against both 0.17 and 1.0.0 b2 and handles the full 400+ class hiearchy tree display in about 4 seconds.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_class.py
+++ b/pywbemtools/pywbemcli/_cmd_class.py
@@ -823,56 +823,89 @@ def cmd_class_find(context, classname_glob, options):
         raise_pywbem_error_exception(er)
 
 
+def get_class_hierarchy(conn, classname, namespace, superclasses=None):
+    """
+    Get the class hierarchy from the server, either the superclasses
+    associated or subclasseswith classname .  If getting subclasses
+    the classname parameter may be None which requests that the complete
+    class hiearchy be retrieved.
+
+      Parameters:
+        conn (:class:`~pywbem.WBEMConnection` or subclass):
+            Current connection to a WBEM server.
+
+        classname(:term:`string):
+            classname if the tree is to be initiated from
+            within the class hiearchy. May be None
+
+        namespace(:term:`string):
+          Namespace to use to acquire classes from WBEM server
+
+        superclasses (:class:`py:bool`):
+            If `True` display the superclasses of classname. If not True
+            display the subclasses. The default is None (display subclasses).
+
+      Returns:
+        tuple of  classes, classnames where
+          classes (list of :class:`~pywbem.CIMClass`): that are either the
+          superclasses or the subclasses of classname.
+
+      Raises:
+        CIM_Error:
+    """
+
+    try:
+        if superclasses:
+            # classname must exist for superclass tree
+            assert classname is not None
+
+            # get the superclasses into a list
+            klass = conn.GetClass(classname, namespace=namespace)
+            classes = []
+            classes.append(klass)
+            while klass.superclass:
+                klass = conn.GetClass(klass.superclass, namespace=namespace)
+                classes.append(klass)
+        else:
+            # get the subclass hierarchy either complete or starting at the
+            # optional CLASSNAME. Gets minimum data from server to define
+            # class, superclass data
+            classes = conn.EnumerateClasses(ClassName=classname,
+                                            namespace=namespace,
+                                            LocalOnly=True,
+                                            IncludeQualifiers=False,
+                                            DeepInheritance=True)
+
+    except Error as er:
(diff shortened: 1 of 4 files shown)


pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -823,58 +823,79 @@ def cmd_class_find(context, classname_glob, options):
...
-    left-justified tree using the asciitree library.
-    The --superclasses option determines if the superclass tree or the
-    subclass tree is displayed.
+    Get the class hiearchy from the server, either the superclasses
+    associated with a class or the subclasses.  If getting subclasses
+    the classname parameter may be None which requests that the complete
+    class hiearchy be retrieved..
+
+      Parameters:
+        conn(TODO):
```
andy-maier, line 834: Change to:
```
    conn(WBEMConnection): Connection to the server.
```

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -823,58 +823,79 @@ def cmd_class_find(context, classname_glob, options):
...
+
+        classname(:term:`string):
+            classname if the tree is to be initiated from
+            within the class hiearchy. May be None
+
+        namespace(:term:`string):
+          Namespace to use to acquire classes from WBEM server
+
```
andy-maier, line 842: Add a description of the `superclasses` parameter.

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -823,58 +823,79 @@ def cmd_class_find(context, classname_glob, options):
...
+        CIM_Error:
     """
 
     # TODO FUTURE: Sort out how we handle output format with tree output.
     try:
-        if options['superclasses']:
+        if superclasses:
             if classname is None:
                 raise click.ClickException('CLASSNAME argument required for '
                                            '--superclasses option')
```
andy-maier, line 854: It seems to me that such a handling of missing command arguments should be done by the caller, and not in this retrieval function.

pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -823,58 +823,79 @@ def cmd_class_find(context, classname_glob, options):
...
+        conn(TODO):
+
+        classname(:term:`string):
+            classname if the tree is to be initiated from
+            within the class hiearchy. May be None
+
+        namespace(:term:`string):
+          Namespace to use to acquire classes from WBEM server
+
+      Returns:
```
andy-maier, line 843: Add description of return value

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: Done

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: Do not understand this comment.  The code is

if superclasses:

It is not there as an assertion but as a test whether the user wanted superclasses or subclasses.

**DISCUSSION**

andy-maier: That was also a good change, but I meant the whole check on lines 858 to 861 (in the changed version):
```
        if superclasses:
            if classname is None:
                raise click.ClickException('CLASSNAME argument required for '
                                           '--superclasses option')
```
That check is already done in cmd_class_tree() which is the only caller.
So the check here could be changed to become an assertion:
```
        if superclasses:
            assert classname is not None
```

andy-maier:
There is one comment left - see above.

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: Whoops, missed that duplication.  Fixed. DONE

pywbemtools/pywbemcli/_cmd_class.py
KSchopmeyer: DONE

Issue #458, opened 2019-12-08
instance associators/references tests with wildcard keybindings fail


KSchopmeyer:
We have two tests that use mock prompt that started failing on some versions of python when we moved to use python 0.15.0 (Note that this may not even be tied to that version but the failures started when we started testing with this verrsion of pywbem).

These fails are marked with this issue number in test_instance_cmds.py

andy-maier:
Not clear when the issue was fixed, but PR #656 removed the TODO for this issue and verifies that the testcase now works (with 0.7.0 dev). Considering this issue as fixed.

PR #656, opened 2020-07-01
Partly fixes #586, #458: Addressed some TODOs and improved command_test()

### LINKED ISSUE #586: Work down TODOs
Current list of TODOs as of 0.6.0 is:

```
************* Module pywbemtools.pywbemcli._association_shrub
W: 44, 0: TODO: Future Could we combine this tree into tree file??? (fixme)
W:513, 0: TODO: ks: Create more general width algorithm (fixme)
W:531, 0: TODO: Remove this before release (fixme)
************* Module pywbemtools.pywbemcli._cmd_class
W:830, 0: TODO FUTURE: Sort out how we handle output format with tree output. (fixme)
************* Module pywbemtools.pywbemcli._cmd_instance
W:685, 0: TODO: Reflow according to target width (fixme)
************* Module pywbemtools.pywbemcli._cmd_server
W:196, 0: TODO: reactivate and implement this in version 0.6.0 (fixme)
************* Module pywbemtools.pywbemcli._common
W:587, 0: TODO: Avoid the CIMClass object as input, because GetClass is not (fixme)
************* Module pywbemtools.pywbemcli._connection_repository
W:294, 0: TODO should "Default failed be part of this message"? (fixme)
************* Module pywbemtools.pywbemcli._context_obj
W:151, 0: TODO there are two creates and also, since all the inputs (fixme)
W:302, 0: TODO: clean up pywbem stats so this is in pywbem, not here (fixme)
W:364, 0: TODO: FUTURE investigate putting this into parent context (fixme)
************* Module pywbemtools.pywbemcli.pywbemcli
W:416, 0: TODO: Future: Create common method with code in build_respository (fixme)
W:511, 0: TODO: refactor this code to avoid deepcopy when not needed. (fixme)
************* Module test_pegasus
W:170, 0: TODO finish this based on the test_ops in the tools directory (fixme)
W:177, 0: TODO create tests for qualifier, server (fixme)
************* Module tests.unit.test_class_cmds
W:870, 0: TODO include class origin. TODO not returning class origin correctly. (fixme)
W:1513, 0: TODO command class delete. Extend this test to use stdin (delete, test) (fixme)
W:1515, 0: TODO: add test for  errors: class invalid, namespace invalid (fixme)
W:1519, 0: TODO the following two test classes should be removed as I believe they (fixme)
************* Module tests.unit.test_common
W:1073, 0: TODO: The mock for the following test is broken. Not sure yet how to (fixme)
W:1122, 0: TODO: This does not really test the abort_msg that is output (fixme)
W:1594, 0: TODO create multiple and test result (fixme)
W:1691, 0: TODO: Convert this test to pytest (fixme)
W:1793, 0: TODO add datetime property (fixme)
W:1840, 0: TODO expand this to test errors for other types (fixme)
W:1874, 0: TODO test no properties, test property with no value (fixme)
W:1894, 0: TODO add datetime property (fixme)
W:2009, 0: TODO: add one class with ref property as key. Is that even allowed? (fixme)
W:2105, 0: TODO key with no value, datetime key (fixme)
W:2432, 0: TODO this is a pytest. param (fixme)
W:2640, 0: TODO: See line 973. We have some test duplication. (fixme)
W:2747, 0: TODO fix simplified_test_function so we can use so we capture output. (fixme)
W:2763, 0: TODO Test compare and failure in compare_obj (fixme)
W:2765, 0: TODO Test compare with errors (fixme)
************* Module tests.unit.test_connection_cmds
W:580, 0: TODO 'PYWBEMCLI_NO_VERIFY=True$', (fixme)
************* Module tests.unit.test_general_options
W:335, 0: TODO: This fails with pywbemcli version 0.5.0 PY2. Temporarily disabled (fixme)
W:924, 0: TODO: test use --name with valid name in interactive mode. (fixme)
W:1142, 0: TODO this test fails on windows because the statement that builds the (fixme)
W:1191, 0: TODO add test for pull operations with pull ops max size variations (fixme)
************* Module tests.unit.test_instance_cmds
W:495, 0: TODO: Add tests for output format xml, repr, txt (fixme)
W:835, 0: TODO: modify this to output log and test for info in return and log (fixme)
W:892, 0: TODO the following uses deep-inheritance because of issue in pywbem_mock (fixme)
W:1599, 0: TODO: For some reason the required quote marks on the instance IDs in (fixme)
W:1903, 0: TODO additional modify error tests required (fixme)
W:2237, 0: TODO add more invalid references tests (fixme)
W:2373, 0: TODO add more associators error tests (fixme)
W:2590, 0: TODO expand the number of invokemethod tests to include all options (fixme)
W:2592, 0: TODO: Create new method that has all param types and options (fixme)
W:2657, 0: TODO: KS. using host causes random issues (fixme)
W:2778, 0: TODO: Fix the following failed tests. Some minor diff in table I have (fixme)
W:2806, 0: TODO test results if we have keys that get hidden.  This will require (fixme)
************* Module tests.unit.test_pywbem_server
W: 29, 0: TODO: Future - Move these tests to pytest (fixme)
W:103, 0: TODO test for errors (fixme)
************* Module tests.unit.test_qualdecl_cmds
W:149, 0: TODO: Add tests for xml, repr, txt formats. (fixme)
************* Module tests.unit.test_server_cmds
W:225, 0: TODO: Remove this test once removal of --sort option is agreed. (fixme)
W:344, 0: TODO add error tests. (fixme)
************* Module tests.unit.test_tableformat
W: 37, 0: TODO ks rewrite this test for pytest.  Note that the capture IO (fixme)
************* Module tests.unit.utils
W: 76, 0: TODO should we consider removing all env variables before each (fixme)
************* Module wbemserver_mock
W:104, 0: TODO: We may expand this for more scoping tests. (fixme)
W:489, 0: TODO this should be driven by the input dictionary (fixme)
W:494, 0: TODO this is simplistic form and only builds one instance of (fixme)
W:504, 0: TODO we are fixing the host name here.  should get from conn (fixme)
```

andy-maier:
See commit messages. The changes have been distributed across multiple commits.

### DIFF
--- a/tests/unit/cli_test_extensions.py
+++ b/tests/unit/cli_test_extensions.py
@@ -55,47 +55,61 @@ class CLITestsBase(object):
             Description of the test
 
           command_grp (:term:`string`):
-            pywbemcli command_grp definedfor this test.  This is the first
-            level of the command(the command group) (ex. class).
+            Pywbemcli command group for this test. This is the first level of
+            the command, e.g. 'class' or 'instance'.
 
-          inputs (:class:`py:dict` or :class:`py:list` of :term:`string` or :term:`string`):
+          inputs (:term:`string` or tuple/list of :term:`string` or :class:`py:dict`):
 
-            If inputs is a list of strings, it contains the command
-            line arguments, without the command name.
+            * If inputs is a string or a tuple/list of strings, it contains the
+              command line arguments, without the command name or command group.
 
-            Each single argument must be its own item in the iterable; combining
-            the arguments into a string does not work.
-            The arguments may be binary strings encoded in UTF-8, or unicode
-            strings.
+              Each single argument must be its own item in the iterable;
+              combining the arguments into a string does not work, and the
+              string is not split into words (anymore).
 
-            If inputs is a dict it can contain the following possible keys:
+              The arguments may be binary strings encoded in UTF-8, or unicode
+              strings.
 
-              * 'args': defines local arguments to append to the command after
-                  the command name. Each single argument must be its own
-                  item in the iterable.
+            * If inputs is a dict it can contain the following optional items:
 
-              * 'general': defines general arguments that will be inserted
-                  into the command line before the command name. Each
-                  single argument must be its own item in the iterable.
+              - 'args': String or tuple/list of strings with local
+                (command-level) options that will be added to the command
+                line after the command name.
 
-              * 'env': Dictionary of environment variables where  key is the
-                variable name as a :term:`string`; dict value is the variable
-                value as a :term:`string` (without any shell escaping needed).
-                If None, no environment variables are set for the test.
+                Each single argument (e.g. option name and option arguments)
+                must be its own item in the iterable.
 
-              * 'stdin': If the key is `stdin`, the value is either a string or
-                list of strings. If `stdin` is a list or tuple of strings, each
-                string defines a line of text that will be included as stdin to
-                the call to pywbemcli. The iterable is mapped to a single string
-                with each line separated by an EOL char. This becomes the stdin
-                parameter for the call to pywbemcli. Each line is processed
-                as a separate repl line by the pybemcli executable.
-                If `stdin` is a single string it is passed  to the stdin
-                parameter for the call to pywbemcli. If `stdin` is defined, the
-                parameters defining a command to be executed are ignored and
-                the full command must be in the `stdin` iterable.
+              - 'general': String or tuple/list of strings with general
+                options that will be added to the command line before the
+                command name.
 
-                If None, no stdin input is defined for the test.
+                Each single argument (e.g. option name and option arguments)
+                must be its own item in the iterable.
+
+              - 'env': Dictionary of environment variables where key is the
+                variable name; dict value is the variable value (without any
+                shell escaping needed).
+
+                If omitted, None, or empty, no environment variables will be
+                set for the test.
+
+              - 'stdin': A string or a tuple/list of strings, that contains the
+                standard input for the command.
+
+                This can be used for commands in interactive mode, and for
+                responding to prompts.
+
+                If present, the command group specified in the `command_grp`
+                parameter is not added to the command line, but the local and
+                global options are added.
+
+                If specified as a tuple/list of strings, these strings are
+                joined into a single string separated by an EOL char. Each line
+                is processed as a separate repl line or prompt input by
+                pybemcli.
+
+                If omitted or None, no standard input will be provided to the
+                pywbemcli command.
 
           exp_response (:class:`py:dict`)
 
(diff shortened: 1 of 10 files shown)


tests/unit/cli_test_extensions.py
```diff
@@ -170,13 +184,17 @@ def command_test(self, desc, command_grp, inputs, exp_response, mock_files,
...
             If True, the test is executed.
             If False, the test is skipped.
             If 'pdb', the test breaks in the debugger.
+            If 'verbose' and verbose is None, verbose is set to True.
 
           verbose (:class:`py:bool`):
-            If `True` the assembled command line will be displayed
+            If `True` the assembled command line and the environment variables
+            set by 'env' will be displayed.
+            If `None`, verbose is set to True if condition=='verbose', otherwise
```
KSchopmeyer, line 196: I must be getting old.  This statement confuses me.

tests/unit/test_instance_cmds.py
```diff
@@ -2939,4 +2952,4 @@ def test_execute_pywbemcli(self, desc, inputs, exp_response, mock,
         Execute pybemcli with the defined input and test output.
         """
         self.command_test(desc, self.command_group, inputs, exp_response,
-                          mock, condition, verbose=False)
+                          mock, condition)
```
KSchopmeyer, line 2955: In removing the verbose=False, we are sort of hiding the fact that this exists.  That makes sense of general code but I think it would be easier for testers if the default was in place as a reminder.  This was done for all tests.

tests/unit/cli_test_extensions.py
andy-maier: I agree it is hard to read, but it is meant as written. None means that the condition argument determines the verbosity.

But maybe that is too complicated, and we should not have a value of None for verbose, and display in verbose mode if verbose == True or condition == 'verbose'. I'll change that.

tests/unit/test_instance_cmds.py
andy-maier: That is the case for any parameter that the caller does not specify. Are you arguing that a caller should never benefit from parameter defaults?

tests/unit/cli_test_extensions.py
KSchopmeyer: OK

tests/unit/test_instance_cmds.py
KSchopmeyer: NO, just noting that I often use that variable and it is easy to remember if it is  there whether True or False.  Old peoples memory thing.

KSchopmeyer: This is only because this is in test code and having the variable there is a visual reminder when you are trying to figure out why a test blew up.

KSchopmeyer:
APPROVED

PR #659, opened 2020-07-02
Fixes #596: Location of connection file

### LINKED ISSUE #596: Location of the pywbem connections file
Today the file is located in the current directory. 

There are several logical alternatives:

1. located in current directory
2. Located in users home directory
3. Search first current directory and then user home directory
4. New optional parameter to identify the source of the connection file.

NOTE: Since this file is not provided as configuration information with install either clone or PYPI there is no reason for an alternative to look for it in the OS base app configuration information directories.

 It is proposed that it be set into the users home directory.

Note that we want to coordinate this so that the repl history file and the connection file are not in completely independent and different locations.

KSchopmeyer:
Allows connection file to be in either users home directory (default)
or directory/filename defined by new parameter --connections-file

The code in ConnectionRepository was refactored to:

1. Eliminate the class attributes for the repository dictionary and
make the whole class based on the class instance. Thus the Connection
Repository is instantiated only once in an interactive session.
This meant:

    a. Added the created ConnectionRepository instance to the context_obj

    b. References the contextObj.connections_repo rather than constructing a new  ConnectionRepository each time a command uses the repository    c. Make the ConnectionRepository lazy load the YAML file  so that a new repository can be created without getting   caught in the issue of the code not existing.  Each  read connection repository read access  method attempts to   load the repository so that the repository is loaded only   the first time it is used.
    d. Changed some variable naming in pywbemcli when the new  general options --connection-file was added.
    d. Specifically test for existence of a connections_file in  a number of the connection command group commands.

2. Add a new general option (--connections-file) that allows the   user to specifically define the connections file.

3. Move the whole decision process with the name of the file into   pywbemcli before the instance of ContextObj is created so that
   it is the ConnectionRepository object that is maintained in   ContextObj.

NOTE:

1.  The tests currently fail on two cases in test_connection_cmd.py where we use stdin to execute multiple statements.  This worked before and now fails to build the new repository in each case.

2. test_connection_cmds.py uses only the --connections-file so that the tests can run with a know file and file state.

### DIFF
--- a/docs/pywbemcli/generaloptions.rst
+++ b/docs/pywbemcli/generaloptions.rst
@@ -640,13 +640,16 @@ command completion.
 .. _`--connections-file general option`:
 
 ``--connections-file`` general option
-""""""""""""""""""""""""""""""""""""""
+"""""""""""""""""""""""""""""""""""""
 
 The ``--connections-file``/``-C`` general option allows the user to select
 a file path for the connections file. It is the file path of a YAML file
 that defines server configurations by name.  The default if this option is
-not specified is to use the configuration file
-``pywbemcli_connection_definitions.yaml`` defined in the user home directory.
+not specified is to use the connections file
+``pywbemcli_connection_definitions.yaml`` in the users home directory.
+
+The users home directory is OS dependent.  The connection file name is
+displayed in the ``connections list`` command
 
 When this option is set, the YAML file defined with the option is used
 as the connections file.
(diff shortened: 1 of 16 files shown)


docs/pywbemcli/generaloptions.rst
```diff
@@ -635,6 +635,29 @@ normally return nothing upon successful execution(ex. instance delete,
...
+.. _`--connections-file general option`:
+
+``--connections-file`` general option
+""""""""""""""""""""""""""""""""""""""
+
+The ``--connections-file``/``-C`` general option allows the user to select
+a file path for the connections file. It is the file path of a YAML file
+that defines server configurations by name.  The default if this option is
+not specified is to use the configuration file
+``pywbemcli_connection_definitions.yaml`` defined in the user home directory.
```
andy-maier, line 649: 1. c/configuration file/connections file/
2. The file is not "defined" in the user home directory, it simply is there. I.e. remove "defined".

KSchopmeyer: DONE

KSchopmeyer:
I made changes per comment and some other things I found yesterday in a separate commit.

PR #665, opened 2020-07-11
Add test for some miscelaneous common errors

KSchopmeyer:
Adds tests for some common errors that cross multiple commands.

1. invalid namespace

2. connection fail

This is to assure that the exceptions caught are integrated into all of
the commands


andy-maier:
Rebased to resolve conflict in changes.rst.

Issue #660, opened 2020-07-05
pywbemcli help for --connections-file


KSchopmeyer:
The general help outputs a help text that defines the default connections-file as the file name "pywbemcli_connection_definitions.yaml" in the users home directory.  What it should really output it the exact path name for that file.  This is trivial for linux  but can be a mess for windows to find the home directory.

However, the current test structure in test_generalopionts.py tests the full text of the help against the pywbemcli -h output.   If we were to insert the actual path of the default file, that messes up the current test.  Maybe it is time to change that test so it matches the others, testing portions of the help rather than the complete help. Then the help output could ignore that variable data.

Right now I set this one to 0.8.0

**DISCUSSION**

andy-maier:
Adding the file path to the help text requires loosening the check for the expected help text, e.g. to check only for a few strings to be in, because the location changes from test system to test system.

KSchopmeyer:
See pr #675

PR #675, opened 2020-07-18
Fixes issue  #660, show default connection file in general help

KSchopmeyer:
Modifies the general help to include the actual help file path in the
help text for the --connections-file general option.

To do this we had to modify the general help text from a single compare
of the complete file to compare of individual components of the help
output.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -26,6 +26,8 @@ Released: not yet
 
 **Enhancements:**
 
+* Modify general help to display the full path of the default connections file.
+  (See issue #660)
 
 **Known issues:**
 
@@ -45,7 +47,7 @@ Released: 2020-07-12
 
 * The default location for the connections file (pywbemcli_connection_definitions.yaml)
   has been moved from the users current directory to the users home directory.
-  A general option "connections_file" allows the user to set other directories
+  A general option (``connections_file``) allows the user to set other directories
   and file names for this file. (See issue #596)
 
 **Deprecations:**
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -22,6 +22,9 @@ Released: not yet
 
 **Enhancements:**
 
+* Modify general help to display the full path of the default help file.
```
andy-maier, line 25: Change "default help file" to "default connections file".

andy-maier, line 25: Change "default help file" to "default connections file".

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -265,7 +265,7 @@ def validate_connections_file(connections_repo):
                    'definitions. The default if this option is not specified '
                    'is the file name "{df}" in the users home directory. '
```
andy-maier, line 266: The variable shown is no longer just the file name, but the full file path. What we still want to get across is that that directory is the home directory, because it will change for each user.

I suggest changing the text to something like:
```
    'is the file "{df}" (in the users home directory). '
```

KSchopmeyer: DONE

docs/changes.rst
KSchopmeyer: DONE

KSchopmeyer:
Requested changes made, rebased and resubmitted with review-needed marked.

Issue #680, opened 2020-07-21
pywbem cli test loses the default connections file


KSchopmeyer:
There is an issue in the test mechanism that backs up and restores the default connection file such that the yaml file can be lost during tests.  The expected behavior is that the file and its backup are saved at the beginning of each test session and restored
at the end of the session but we have noticed the yaml file gone.

KSchopmeyer:
See pr #681

andy-maier:
Addressed by PR #681 for master and PR #685 for 0.7.1.

PR #681, opened 2020-07-21
Fix issue #680 - tests lose default connection file.

KSchopmeyer:
This fixes an issue where the default connection file was not being
restored in all cases when tests were run. It also cleans up the
nameing of the paths to be saved and restored to be more readable.

Note that this is a test change only.  It does not affect our CI tests
but could affect a user who had a data in the default connections file
and ran the pywbemtools test suite.

**DISCUSSION** Should we consider backing this one to 7.1???. It is an issue with the test suite but could lose the default connections file for a user.


andy-maier:
PR #685 rolls back this PR into 0.7.1.

Issue #224, opened 2019-08-04
Repeated unhandled exception in event loop, Python 2


KSchopmeyer:
In the REPL mode we get repeated unhandled exceptions in the  prompt_toolkit event loop.

As noted in issue #192, we can fix this by restricting the use of prompt-toolkit to < 2.0.0 for python 2.7.  That fix was proposed for  #71 

However we still have the longer term issue of using Click with python 2.7 and updated versions of prompt-toolkit so this issue is now documented to document the issues.

Note that the tracebac for the problem occurance is in issue #192 and that the issue is apparently in prompt-toolkit trying to issue the hints as commands and subcommands are entered which causes an unhandled Exception with note that it is expecting HTML,... utf-8 but for which Click and Click Repl are apparently supplying not UTF-8 for some reason.

I also propose that we not make this issue 0.5.0 but at least 0.6.0

Note the only investigation needed is to determine whether the issue in in click, clck-repl, or prompt_toolkit and file a bug.

This appears to be an issue causes when we started to use prompt-toolkit version 2.0. That does not mean that it is prompt-toolkit that is actually responsible for the problem however.

KSchopmeyer:
The following is a typical set of exceptions. This occurs only when repl is enabled.

Unhandled exception in event loop:
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/eventloop/posix.py", line 155, in _run_task
    t()
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/eventloop/context.py", line 116, in new_func
    return func(*a, **kw)
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/application/application.py", line 403, in redraw
    self._redraw()
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/application/application.py", line 471, in _redraw
    self.renderer.render(self, self.layout)
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/renderer.py", line 588, in render
    screen.draw_all_floats()
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/prompt_toolkit/layout/screen.py", line 232, in draw_all_floats
    functions[0][1]()
[...]

PR #686, opened 2020-07-25
fixes issue #224 py2 click-repl exception when command entered

KSchopmeyer:

We did not add any tests for this because I could not work out a way to execute the completer code in test mode.  The logical test is to execute partial commands in interactive mode (ex. "class get "). The completer would then initiate but that does not work when commands are entered through stdin.

The manual test consisted of entering interactive mode and then step through the various options of command group and command followed by a space to initiate the completion logic.   That works manually.

### DIFF
--- a/docs/pywbemcli/cmdshelp.rst
+++ b/docs/pywbemcli/cmdshelp.rst
@@ -786,7 +786,7 @@ Help text for ``pywbemcli connection select`` (see :ref:`connection select comma
           . . .
 
     Command Options:
-      -d, --default  If set, the connection is set to be the default connection  in the connections file in addition to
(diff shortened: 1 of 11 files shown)


KSchopmeyer:
I marked this for rollback but that is up for discussion.  The 0.7.0 code works and the only reason for this fix is so that we can get prompt-toolkit off of our radar.  Since we are going to release 0.8 in near future, and that has fix, I am not sure that the rollback is worth it.

**DISCUSSION** should we roll back

andy-maier:
Since we have the upcoming 0.7.1 open, I decided to roll this one back. See PR #688

Issue #690, opened 2020-07-27
I am getting a Windows error (WinError 995) on a number of pywbemcli commands


FarmerMike252:
I am getting the WinError 995 error on some of my pywbemcli commands. The command produces the expected output but then I get the following error message:

Unhandled exception in event loop:
  File "c:\users\farmermike\appdata\local\programs\python\python38-32\lib\asyncio\proactor_events.py", line 768, in _loop_self_reading
    f.result()  # may raise
  File "c:\users\farmermike\appdata\local\programs\python\python38-32\lib\asyncio\windows_events.py", line 808, in _poll
    value = callback(transferred, key, ov)
  File "c:\users\farmermike\appdata\local\programs\python\python38-32\lib\asyncio\windows_events.py", line 457, in finish_recv
    raise ConnectionResetError(*exc.args)

Exception [WinError 995] The I/O operation has been aborted because of either a thread exit or an application request
Press ENTER to continue...

I press enter and can continue with my script, so this is not a high priority, but it would nice to get rid of the error message.

andy-maier:
This seems to be the error reported here: https://github.com/prompt-toolkit/python-prompt-toolkit/issues/1023#issue-534396318. They have opened a Python bug against 3.8, which changed the type of event processor to async. The prompt-toolkit issue suggests that changing the type of event processor back may address the issue.

The 'blessed' project also encountered this (issue https://github.com/jquast/blessed/issues/121) and they relate it to using prompt-toolkit>=3.0.

I think we should first try to pin prompt-toolkit to <3.0.

andy-maier:
@FarmerMike252 Mike, I have a PR #696 that I think fixes this, but it requires your confirmation. Could you please verify that the error is gone when using pywbemcli from the branch of that PR?

FarmerMike252:
I have run two of my CLI scripts, including the one that had errors in yesterday's SM Lab meeting. Both scripts ran clean. So, I believe the problem is successfully fixed.

FarmerMike252:
Ooops! I didn't realize I was not supposed to close the issue.

andy-maier:
NP, Mike. Thanks for testing it! I'll roll back the PR to 0.7.1 then and we'll discuss releasing it in our call today.

PR #696, opened 2020-07-30
Pinned prompt-toolkit to <3.0 on Python 3.8 on Windows to avoid WinError 995

andy-maier:
See commit message.

Requires test of the interactive mode on Windows with Python 3.8.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,13 +19,22 @@ click-spinner>=0.1.8
 click-repl>=0.1.6
 asciitree>=0.3.3
 tabulate>=0.8.2
-prompt-toolkit>=2.0.1; python_version >= '3.4'
-# See pywbemtools issue # 192. The repl mode fails with unexpected Exception
+
+# prompt-toolkit>=2.0 failed on py27 (issue #192), so it was pinned to <2.0.
+#   Later, the fix for issue #224 allowed to lift that pinning.
+# prompt-toolkit>=3.0 does not support py27.
+# prompt-toolkit>=3.0 may cause WinError 995 on py38 on Windows (issue #690).
 prompt-toolkit>=1.0.15,<3.0.0; python_version == '2.7'
+prompt-toolkit>=2.0.1; python_version >= '3.4' and python_version < '3.8'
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -19,13 +19,22 @@ click-spinner>=0.1.8
...
 tabulate>=0.8.2
-prompt-toolkit>=2.0.1; python_version >= '3.4'
-# See pywbemtools issue # 192. The repl mode fails with unexpected Exception
+
+# prompt-toolkit>=2.0 failed on py27 (issue #192), so it was pinned to <2.0.
+#   Later, the fix for issue #224 allowed to lift that pinning.
+# prompt-toolkit>=3.0 does not support py27.
+# prompt-toolkit>=3.0 may cause WinError 995 on py38 on Windows (issue #690).
 prompt-toolkit>=1.0.15,<3.0.0; python_version == '2.7'
+prompt-toolkit>=2.0.1; python_version >= '3.4' and python_version < '3.8'
```
KSchopmeyer, line 28: Was there a specific reason we set the minimum to 2.0.1 for python 3 and now still keep 1.0.15 for python 2.7?   There was an issue with click_repl (click_repl issue # 34 early in the release of the 2.0 toolkit but that was fixed). Which I think is whyu we specify the minimum as 2.0.1.

KSchopmeyer:
I ran both the test suite and a number of commands manually on windows with no issue.

requirements.txt
andy-maier: First, the minimum versions were not changed by this PR, it only added the pinning to <3.0.0 for py38 on Windows.

The reason the minimum versions are different is that at some point, on py27 it was pinned to <2.0.0 so the minimum version naturally stayed at a 1.x version. Later, the pinning could be lifted to <3.0.0 but there was no real reason to increase the minimum version. That's all on py27.

The minimum version for py3 needs to be a 2.x version because that is required to get py3 support.

andy-maier:
Mike tested it (see issue #690), so we are good to go.

PR #706, opened 2020-08-08
Enforced clean Pylint check

andy-maier:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,6 +225,17 @@ pywbemcli_termwidth := 120
 # PyLint config file
 pylint_rc_file := pylintrc
 
+ifeq ($(PACKAGE_LEVEL),minimum)
+  # Using pywbem 0.x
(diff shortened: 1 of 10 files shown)


andy-maier:
I put review needed on again, after fixing the issues that caused pylint to fail with minimum package levels.

pywbemtools/pywbemcli/_common.py
```diff
@@ -272,6 +272,7 @@ def pick_one_from_list(context, options, title):
         context.spinner_stop()
 
     click.echo(title)
+    index = 0  # Avoid pylint undefined-loop-variable
```
KSchopmeyer, line 275: This pylint should no longer be occuring since I changed to not use the index outside the loop and added max_option yesterday (approved and committed)

andy-maier: DONE - removed it.

PR #711, opened 2020-08-08
Test: Fixed method invocations

andy-maier:
See commit message.

The reason this still fails is pywbem issue https://github.com/pywbem/pywbem/issues/2412.


andy-maier:
Abandoned in favor of PR #712.

PR #712, opened 2020-08-08
Multiple fixes to make tests succeed again with pywbem 1.0 and 0.x

KSchopmeyer:
Fixes a number of issues that appeared at the last minute in testing with pywbem 1.0.0 including:

This PR is built on top of PR #711m the invokemethod changes.

NOTE: This will fail until pr #711 passes.

1. The tests added a new qualifier decl to the simple_mock_model causing the displays of quallifiers to change.

2. There are several cases where tests fail (primarily in instance) becuase the UserWarning  for NocaseDict and ordering is now  possibly included in the output.  Changed these to use innows rather than lines or linesnows

3. For some reason that I have not sorted out, the CIMError for instance enumerate CIM_F00x returns INVSLID_CLASS where it prevously returned NOT_FOUND.   I assume something changed in pywbem_mock.

NOTE: This has been tested with pywbem PR #2414 and now appears to pass all t ests.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -42,6 +42,9 @@ Released: not yet
   some cases for pywbem 1 vs previous versions(inclusion of "/:" prefix).
   (see issue #704)
 
+* Test: Fixed attempt in test_class_cmds.py to invoke a non-static method on a
+  class object. (see issue #707)
+
 **Enhancements:**
 
 * Modify general help to display the full path of the default connections file.
--- a/pywbemtools/pywbemcli/_association_shrub.py
+++ b/pywbemtools/pywbemcli/_association_shrub.py
@@ -663,7 +663,7 @@ class AssociationShrub(object):
 
         ret.append('.\n')
 
-        for key in case_sorted(path.keybindings.iterkeys()):
+        for key in case_sorted(six.iterkeys(path.keybindings)):
             value = path.keybindings[key]
 
             ret.append(key)
--- a/tests/unit/all_types_method_mock_v1.py
+++ b/tests/unit/all_types_method_mock_v1.py
@@ -1,10 +1,8 @@
 """
 mock_pywbem test script that installs a method callback to be executed. This is
 based on the CIM_Foo class in the simple_mock_model.mof test file
-
 """
 
-import six
 from pywbem_mock import MethodProvider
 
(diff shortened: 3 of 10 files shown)


andy-maier:
I'll abandon PR #711 then we can use this single PR to have both commits.

Pywbem PR https://github.com/pywbem/pywbem/pull/2416 adds support for NocaseDict as a valid output parameters type, and I am currently testing this PR here against pywbem with that PR, and against pywbem 0.17.4. If both pass, this PR can be merged.

andy-maier:
One more fix is needed: In _association_shrub.py line 666, there is a use of `path.keybindings.iterkeys()` which is not available on python 3 anymore, since we are using the NocaseDict from nocasedict. CHanged that to use `six.iterkeys(path.keybindings)`, and will push the commit with the fix to this PR.

andy-maier:
The test of the latest version of this PR against pywbem as of PR pywbem/pywbem#2416 succeeded.
However, testing against pywbem stable_0.17 reveals that the new FuzzyStatic method that was added, was not registered the old way. Plus, it does not reject invoking the non-static method Fuzzy on a class. I will push another commit to this PR that fixes both of that.

andy-maier:
Testing against pywbem stable_0.17 revealed that one more fix is needed: The NOT_FOUND error you mentioned above cannot just be changed to INVALID_CLASS, because with pywbem < 1.0.0 NOT_FOUND is returned. The reason for the change was that NOT_FOUND is not the correct error if an instance (!!) operation does not find the class. The fix is to expect NOT_FOUND for pywbem <1.0 and INVALID_CLASS for pywbem >=1.0. I'll post a commit to this PR that fixes that.

andy-maier:
I have added a commit to this PR that has all the fixes to run again against pywbem <1.0.
I verified that the updated PR succeeds tests against pywbem stable_0.17 and against pywbem as of PR pywbem/pywbem#2416.
So from the test perspective, this PR would be ready to merge.

andy-maier:
This PR should wait for the release of pywbem 0.17.5 and 1.0.1, and then increase the minimum pywbem version to 0.17.5.

andy-maier:
pywbem 1.0.1 has been released, and this PR has been rebased on master.
It should now succeed and is ready for review.

KSchopmeyer:
APPROVED

PR #714, opened 2020-08-09
0.7: Multiple fixes to make tests succeed again with pywbem 1.0 and 0.x

andy-maier:
Rolls back PR #712 into 0.7.2, plus contains one more commit that fixes a 0.7 specific issue.


andy-maier:
This PR should wait for the release of pywbem 0.17.5 and 1.0.1, and then increase the minimum pywbem version to 0.17.5.

PR #719, opened 2020-08-10
Property list determines property order in instance table output

andy-maier:
See commit message.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -85,6 +85,10 @@ Released: not yet
   Added the 'mock' package and for Python 2.7, the 'funcsigs' package as new
   dependencies.
 
+* Specifying a property list (--pl option) on instance commands with table
+  output formats now uses the order of properties as specified in the property
+  list in the output table, instead of sorting them. (See issue #702)
+
 **Cleanup**
 
 * Remove unused NocaseList from __common.py
--- a/pywbemtools/pywbemcli/_association_shrub.py
+++ b/pywbemtools/pywbemcli/_association_shrub.py
@@ -49,7 +49,7 @@ from pywbem._nocasedict import NocaseDict
 
 from ._common import output_format_is_table, format_table, shorten_path_str, \
     warning_msg, sort_cimobjects
-from ._utils import _ensure_unicode, _to_unicode
+from ._utils import _ensure_unicode, _to_unicode, get_terminal_width
 
 # Same as in pwbem.cimtypes.py
 if six.PY2:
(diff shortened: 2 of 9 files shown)


andy-maier:
The failure on Appveyor seems to be related to terminal width. For some reason, the table content spills over into a new row, even though the table with is only 57.

We should establish a reliable terminal width for these tests.

KSchopmeyer:
This one is something different.  It is like something is attaching an extra column to the table.  Note that the extra line is actually a column that contains an empty string as value.  Since it is a windows thing (Only occurs in one version of windows do you want to mark it as skip test (Mark the test FAIL and create  an issue and I can track it down)

```

  E       AssertionError: Unexpected number of lines on stdout in test:
  E       Verify instance command references with --pl in sorted order
  E       Expected lines cnt=7:
  E       ------------
  E       Instances:TST_Lineage
  E       +--------------+----------------------------------------+
  E       |InstanceID|parent|
  E       |--------------+----------------------------------------|
  E       |"MikeGabi"|"/root/cimv2:TST_Person.name=\"Mike\""|
  E       |"MikeSofi"|"/root/cimv2:TST_Person.name=\"Mike\""|
  E       +--------------+----------------------------------------+
  E       ------------
  E       Actual lines cnt=9:
  E       ------------
  E       Instances:TST_Lineage
  E       +--------------+---------------------------------------+
  E       |InstanceID|parent|
  E       |--------------+---------------------------------------|
  E       |"MikeGabi"|"/root/cimv2:TST_Person.name=\"Mike\"|
  E       ||"""|
  E       |"MikeSofi"|"/root/cimv2:TST_Person.name=\"Mike\"|
  E       ||"""|
  E       +--------------+---------------------------------------+
  E       ------------
  tests\unit\utils.py:279: AssertionError

andy-maier:
It is not adding a column. The actual output has exactly the expected two columns. It is the cell value that spills over into the next line. I can reproduce this on macOS when making the terminal width narrow.

KSchopmeyer:
In the following:

  E       |"MikeGabi"|"/root/cimv2:TST_Person.name=\"Mike\"|
  E       ||"""|

I understand the last quote on the second line but not the surrounding | | characters.

KSchopmeyer:
I agree that we need to fix the terminal width for the tests.   Completely overlooked that.   Question is, what do we fix it at.  To me 80 char sounds logical since that matches what I assume the terminal width is on the CI tests so that we keep the characteristics of current tests.  I would like to look at this further since I run my terminal windows much wider on linux and still successfully run the full test suite all the time.

Also, this means that windows is seeing something different than linux in terminal width and it seems to be just a char or so more narrow that linux when it does the calculations.

andy-maier:
It may be caused by Windows not honoring the terminal width. Needs investigation.

andy-maier:
I have consolidated the determination of the terminal width in the second commit, and that seems to solve the different behaviors.

Please review.

KSchopmeyer (approved):

PR #728, opened 2020-08-12
Renamed default connections file to .pywbemcli_connections.yaml

andy-maier:


docs/changes.rst
```diff
@@ -84,6 +84,11 @@ Released: not yet
 * Resolved remaining Pylint issues and enforced clean pylint checks.
   (See issue #668)
 
+* Renamed the default connections file in the user's home directory from
```
KSchopmeyer, line 87: Should we have not in incompatibilities since this is an incompatible change? We should have some way of telling people when they first install this version that the connection file name changed. Maybe even warning that we found the old name but they need to change its name.

andy-maier: We discussed this and agreed to add code for migrating the file, displaying a message.

andy-maier: Added support for migrating an old connections file, in a second commit.

PR #731, opened 2020-08-16
Add units to property names for table output of instances

andy-maier:
See commit message.

**Note: This PR requires the master branch of pywbem, until pywben 1.1.0 is released.**

Example (property `counter` has `PUnit("byte * 10^3")`):
```
$ pywbemcli -m tests/unit/simple_assoc_mock_model.mof -o table instance enumerate TST_Person
Instances: TST_Person
+------------+----------------+------------+-----------------------+------------------+
| name       | counter [kB]   | gender     | likes                 | secondProperty   |
|------------+----------------+------------+-----------------------+------------------|
| "Gabi"     |                | 1 (female) | 2 (movies)            |                  |
| "Mike"     |                | 2 (male)   | 1 (books), 2 (movies) |                  |
| "Saara"    |                | 1 (female) | 1 (books)             |                  |
| "Sofi"     |                | 1 (female) |                       |                  |
| "Gabisub"  | 4              | 1 (female) |                       | "four"           |
| "Mikesub"  | 1              | 2 (male)   |                       | "one"            |
| "Saarasub" | 2              | 1 (female) |                       | "two"            |
| "Sofisub"  | 3              | 1 (female) |                       | "three"          |
+------------+----------------+------------+-----------------------+------------------+
```

### DIFF
--- a/pywbemtools/pywbemcli/_display_cimobjects.py
+++ b/pywbemtools/pywbemcli/_display_cimobjects.py
@@ -26,9 +26,11 @@ import re
 from pydicti import odicti
 import six
 import click
+from nocasedict import NocaseDict
 
 from pywbem import CIMInstanceName, CIMInstance, CIMClass, \
-    CIMQualifierDeclaration, CIMClassName, ValueMapping
+    CIMQualifierDeclaration, CIMClassName, ValueMapping, siunit_obj, \
+    CIMError, CIM_ERR_NOT_SUPPORTED
 
 from ._common import format_table, fold_strings, DEFAULT_MAX_CELL_WIDTH, \
     output_format_is_table, sort_cimobjects, format_keys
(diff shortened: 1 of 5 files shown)


pywbemtools/pywbemcli/_display_cimobjects.py
```diff
@@ -478,10 +478,45 @@ def _display_instances_as_table(insts, table_width, table_format,
...
         max_cell_width = table_width
 
-    header_line = []
+    # TODO: Decide whether and how the showing of units should be controlled.
+    show_units = True
+
+    if show_units:
+        # Retrieve the most derived class of these instances. This is necessary
+        # because that is the only class that has all properties that may be
+        # needed.
```
andy-maier, line 487: Second thoughts on the implementation: The algorithm to determine a most derived class is not covering cases where multiple leaf classes are involved. Each of them could add new properties with PUnit/Units qualifiers, so that no single class exists that has all of them.

KSchopmeyer: Agreed

KSchopmeyer: Why not build the  table of instances and supplement a all_properties list with the new properties introduced by each instance if that instance has properties not already in the table.  It is more expensive but we are dealing with thousands of instances for this display.

andy-maier: We need the properties of the classes, not of the instances. But yes, something like that.

pywbemtools/pywbemcli/_display_cimobjects.py
```diff
@@ -478,10 +478,45 @@ def _display_instances_as_table(insts, table_width, table_format,
     else:
         max_cell_width = table_width
 
-    header_line = []
+    # TODO: Decide whether and how the showing of units should be controlled.
+    show_units = True
+
+    if show_units:
```
KSchopmeyer, line 484: Obviously we need to account for servers that do not have class repos but I assume that is for after we commit this

pywbemtools/pywbemcli/_display_cimobjects.py
```diff
@@ -478,10 +478,45 @@ def _display_instances_as_table(insts, table_width, table_format,
...
+        for inst in insts:
+            classname = inst.classname
+            if inst.path is None:
+                # The only operation returning instances without a path is
+                # query execution. For now, we simply don't display units in
+                # that case.
+                # TODO: Pass a namespace for query execution to display units
+                continue
+            namespace = inst.path.namespace
+            class_obj = context.conn.GetClass(
```
KSchopmeyer, line 498: This could get the same class many times.   Should we cache the class to be reused?

pywbemtools/pywbemcli/_display_cimobjects.py
andy-maier: DONE - Changed the implementation to retrieve a cache of all creation classes of the instances, which should normally be a short list, and in the loop through the properties, they are inspected for the qualifiers.

pywbemtools/pywbemcli/_display_cimobjects.py
andy-maier: DONE - added code to detect that and in that case, the showing of units is silently given up.

pywbemtools/pywbemcli/_display_cimobjects.py
andy-maier: DONE - the new implementation that addresses the second thoughts stated above gets each class only once.

andy-maier:
Addressed the comments. Set back to need review.

Issue #730, opened 2020-08-16
pywbemcli connection show <name> where name does not exist, poor exception message


KSchopmeyer:
This command issues the message "No connection definition exists."  when the input command is "pywbemcli connection show BADNAME" which is confusing. It should state that no connection definition exists for the name {}, etc. to help the user fix the problem

Issue #732, opened 2020-08-18
Confusion between adding general params and connection show


KSchopmeyer:
There is confusion within pywbemcli between modifying the general options for server parameters and the show command.

Thus with an existing connection tst1:

```
$ pywbemcli -n tst1
pywbemcli: --user blah --password blah connection show
```
... Shows the original values of --user and --password

However:

```
$ pywbemcli -n tst1
pywbemcli: --user blah --password blah connection select tst1
pywbemcli: connection show
```
... Shows the new values

This causes problems with one of the tests in test_general_options.py and that one uses stdin to process the multiple inputs. We disabled that test with PR #730 until this issues is sorted out.

andy-maier:
I think the whole logic in `connection show` needs to be revisited. See also my comments on PR #733.

KSchopmeyer:
Note line 568 in _cmd_connections.  The connection is reloaded from the connections file even if it is the current connection.

andy-maier:
I tried the two cases described above with PR #743 and the result is (for a mock connection "simple"):

```
$ pywbemcli -n simple
pywbemcli> --user blah --password blah connection show
Connection status:
name               value  (state)
-----------------  ------------------------------------------
name               simple (current)
server
default-namespace  root/cimv2
user               blah
password           ******
timeout            30
use-pull
pull-max-cnt
verify             True
certfile
keyfile
mock-server        tests/unit/simple_mock_model.mof
                   tests/unit/simple_mock_invokemethod_v1s.py
ca-certs
```
... Shows the values of --user and --password that are specified for the show command

However:

```
$ pywbemcli -n simple
pywbemcli> --user blah --password blah connection select simple
"simple" current
pywbemcli> connection show
Connection status:
name               value  (state)
-----------------  ------------------------------------------
name               simple (current)
server
default-namespace  root/cimv2
user
password
timeout            30
use-pull
pull-max-cnt
verify             True
certfile
keyfile
mock-server        tests/unit/simple_mock_model.mof
                   tests/unit/simple_mock_invokemethod_v1s.py
ca-certs
```
... Shows the values of --user and --password from the connections file.

I think this is what one should expect, so I think the PR does not only do a cleanup but also addresses the original issue.

I have linked it with the PR so it will be closed when merging the PR.

PR #733, opened 2020-08-18
Fixes issue #730-fix-error-msg in connection show

KSchopmeyer:
Change error message so it correctly shows the name that was causing the
problem.

Adds test.

We changed a couple of the other tests that use stdin because not
putting a command into the first line of the stdin causes click
to output a help message before the first command is processed.
This eliminates that extra help output from the output.

Marked one test in test_generaloptions.py failed since and filed
issue #730 on the apparent issue.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_connection.py
+++ b/pywbemtools/pywbemcli/_cmd_connection.py
@@ -532,15 +532,16 @@ def cmd_connection_show(context, name, options):
     """
     Show the parameters that make up the current connection information if
     name is None. If Name exists, shows connections in connections file.
+    If name is "?" present list of all connections for selection.
     """
 
     connections = context.connections_repo
 
-    cname = get_current_connection_name(context)
+    curr_name = get_current_connection_name(context)
     # If no name arg, fallback to selection unless there is no connections file
     if not name:
-        name = cname or '?'
-        if not cname and not connections.file_exists():
+        name = curr_name or '?'
+        if not curr_name and not connections.file_exists():
             raise click.ClickException('No current connection and no '
                                        'connections file {}.'
                                        .format(connections.connections_file))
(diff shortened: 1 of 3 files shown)


pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -532,15 +532,16 @@ def cmd_connection_show(context, name, options):
     """
     Show the parameters that make up the current connection information if
     name is None. If Name exists, shows connections in connections file.
+    If name is "?" present list of all connections for selection.
```
andy-maier, line 535: This docstring does not describe what the code does. The logic in the code is so complex that it is hard to describe what it actually does, and the comments in the code are also partly incorrect.

Can you try to get the docstring and the comments in sync with the code?

Also, I have to admit that after 10 minutes of trying to understand the logic in the code I gave up. This logic is really convoluted, or at least I cannot follow it this morning.

I think what would help is to explain the precedence of an inherited current connection in an interactive session, and a current connection defined in the connections file.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -532,15 +532,16 @@ def cmd_connection_show(context, name, options):
...
 
     connections = context.connections_repo
 
-    cname = get_current_connection_name(context)
+    curr_name = get_current_connection_name(context)
     # If no name arg, fallback to selection unless there is no connections file
     if not name:
-        name = cname or '?'
-        if not cname and not connections.file_exists():
+        name = curr_name or '?'
```
andy-maier, line 543: This falls back to selection if no name is specified and there is no current connection. That case is not described in the docstring.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -564,20 +565,21 @@ def cmd_connection_show(context, name, options):
     # and that name is not current, use it. If current name is same as
     # name, use the current version.
```
andy-maier, line 566: If no name was specified and there is no current connection but the connections file exists, we get here but name is still not set. So the comment 'Have a name.'  is incorrect, and the logic likely too.

andy-maier (changes requested):
I gave up trying to understand the code logic. See my comments.

andy-maier:
**DISCUSSION**: Should we let this one go in as it just tries to improve the error message, and fix the logic as part of issue #732 ?

andy-maier:
We agreed to let this PR be merged and address any logic inconsistencies when issue #732 is addressed.

andy-maier (approved):

Issue #736, opened 2020-08-23
Not found behavior of ConnectionRepository


andy-maier:
Currently, having no connections file causes the following to raise ConnectionsFileNotFoundError:

```
cr = ConnectionRepository(connections_file)
cr.keys()  # raises not found
'a' in cr  # raises not found
```

Basically, the detection of the not found condition was pushed down into the _load..() method. However, this violates the principle of single purpose: The primary purpose of the _load..() method is to perform lazy loading of the file.

Also, an initially empty ConnectionRepository object is consistent with no connections file being present. Using the object to retrieve any connections should show an empty repo, rather than raising an exception. As connection definitions are added to the ConnectionRepository object, the file builds up.

Suggestion: Remove the ConnectionsFileNotFoundError exception altogether, and perform error handling due to non-existing user-provided connections file at a higher level (than in ConnectionRepository).

PR #737, opened 2020-08-23
Cleanup and improvements in ConnectionRepository

andy-maier:
See commit message.
This PR fixes issue #736 and does additional cleanup and improvements.


pywbemtools/pywbemcli/_connection_repository.py
```diff
@@ -460,79 +544,84 @@ def add(self, svr_definition):
...
         if self._pywbemcli_servers:
             conn_dict = {name: value.to_dict() for name, value in
                          self._pywbemcli_servers.items()}
 
             # build dictionary for YAML output
-            yaml_dict = {ConnectionRepository.connections_group_name: conn_dict,
-                         ConnectionRepository.default_connection_grp_name:
-                             self.default_connection_name}
+            yaml_dict = {
+                self.connections_group_name: conn_dict,
```
KSchopmeyer, line 622: I would prefer to use the classname here because these are class level variables

pywbemtools/pywbemcli/_connection_repository.py
```diff
@@ -58,296 +59,375 @@
...
         # pylint: disable=useless-super-delegation
         super(ConnectionsFileError, self).__init__(message)
 
 
 class ConnectionsFileWriteError(ConnectionsFileError):
     """
-    Exception to be used for connection file write errors.
+    Exception indicating an error when writing the connections file.
 
+    This includes I/O errors during writing of the file or during renaming of
```
KSchopmeyer, line 80: IO and OS errors

pywbemtools/pywbemcli/_connection_repository.py
andy-maier: Class level variables are officially also available on the object (i.e. self). It makes it shorter, but I'm open either way.

pywbemtools/pywbemcli/_connection_repository.py
andy-maier: DONE

pywbemtools/pywbemcli/_connection_repository.py
andy-maier: DONE After discussing with Karl, I have changed it back to using the class name.

andy-maier:
Added more tests and made the error injection more specific (such as failing open only for write mode).

PR #741, opened 2020-08-25
Ks/#732a refactor pywbemcli

KSchopmeyer:
This is a minor refactor of some of pywbemcli before any further work on issue #732. I separated to make it easy to approve in smalller blocks.

Moves some code around, moves some to a function and hides some pylint issues that are occuring on python 3.6.

### DIFF
--- a/pywbemtools/pywbemcli/pywbemcli.py
+++ b/pywbemtools/pywbemcli/pywbemcli.py
@@ -67,6 +67,9 @@ if TERMWIDTH_ENVVAR:
     except ValueError:
         TERMWIDTH_ENVVAR = None
 
+#
+# Context variables passed to click
+#
 CONTEXT_SETTINGS = dict(
 
     # Enable -h as additional help option:
@@ -75,6 +78,12 @@ CONTEXT_SETTINGS = dict(
     # Default the output width properly:
     terminal_width=TERMWIDTH_ENVVAR or click.get_terminal_size()[0],
 )
+
+###########################################################################
+#
+# Support functions for the the cli(...) function
+#
+###########################################################################
 
 
 def validate_connections_file(connections_repo):
@@ -125,6 +134,51 @@ def execute_startup_script(file_path, verbose):
             raise click.Abort()
 
 
+def _resolve_mock_server(mock_server):
+    """
+    Validate and resolve the mock_server parameter.
+    Validates that all files defined in the server exist and that the
+    types are valid.
+
+    Parameters:
+
+      mock_server (tuple of string):
+        Strings defining all of the files that compromise the mock_server
+        definiton.
+
+    Returns: list of resolved mock server definition components derived
+       from the input parameter mock_server.
+
+    Raises:
+      click.ClickException for any errors.
+    """
+    resolved_mock_server = []
+    assert isinstance(mock_server, tuple)
+    # Normalize paths for different OS's
+    mock_server_path = [os.path.normpath(fn) for fn in mock_server]
+
+    # Test for valid file types and existence for mock_server files.
+    for file_path in mock_server_path:
+        ext = os.path.splitext(file_path)[1]
+        if ext not in ['.py', '.mof']:
+            raise click.ClickException(
+                "Mock file '{}' has invalid suffix '{}' "
+                "- must be '.py' or '.mof'".format(file_path, ext))
+        if not os.path.isfile(file_path):
+            raise click.ClickException(
+                "Mock file '{}' does not exist".format(file_path))
+        resolved_mock_server.append(file_path)
+    return resolved_mock_server
+
+
+############################################################################
+#
+#   cli command (main entry point) and definition of all of the pywbemcli
+#   general options.
+#
+############################################################################
+
+
 # pylint: disable=bad-continuation
 # PywbemcliTopGroup sets order commands listed in help output
 @click.group(invoke_without_command=True, cls=PywbemcliTopGroup,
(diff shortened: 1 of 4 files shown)


pywbemtools/pywbemcli/_context_obj.py
```diff
@@ -281,7 +281,7 @@ def execute_cmd(self, cmd):
             self.spinner_start()
         if self.pdb:
             import pdb  # pylint: disable=import-outside-toplevel
-            pdb.set_trace()
+            pdb.set_trace()  # pylint: disable=no-member
```
andy-maier, line 284: In my 3.8 environment, this Pylint issue does not surface. Its version is the latest for Python 3.8:
```
$ pylint --version
pylint 2.4.4
astroid 2.3.3
Python 3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)]
```

Can you please post the Pylint version from the environment where Pylint reports the the "no-member" issues?

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -124,6 +124,58 @@ def execute_startup_script(file_path, verbose):
...
+        definiton.
+
+    Returns: list of resolved mock server definition components derived
+       from the input parameter mock_server.
+
+    Raises:
+      click.ClickException for any errors.
+    """
+    resolved_mock_server = []
+    assert isinstance(mock_server, tuple)
```
andy-maier, line 159: Description says list, but code checks for tuple. Make consistent. Why not accept Sequence?

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -124,6 +124,58 @@ def execute_startup_script(file_path, verbose):
...
+    # resolve relative and absolute paths
+    mock_server_path = []
+    for fn in mock_server:
+        if fn == os.path.basename(fn):
+            mock_server_path.append(os.path.join(os.getcwd(), fn))
+        else:
+            mock_server_path.append(fn)
```
andy-maier, line 166: So if the path is a relative path it stays relative, but if it is just a base file name we prepend the cwd?

I think that prepending is not needed at all and just complicates things. The specified file paths are supposed to be reachable from the current directory, and there is no need to prepend anything.

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: I am using python 3.6.7 and it always shows up.  The error is that it says set_trace attribute does not exist.

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: FIXED

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: Redid this.  Actually changed it to just run all filenames through the normpath.

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: pylint 2.4.4

pywbemtools/pywbemcli/pywbemcli.py
andy-maier: Another new GitHub feature: Outdated comments no longer seem to show up in the file view even when they are not resolved.

On your use of `os.path.normpath()`: Its description says "This string manipulation may change the meaning of a path that contains symbolic links.". So while I like the idea, it is potentially problematic.

Why do we do anything at all? The user ought to specify a valid path name. If the file path references a file then it works and if not then it fails. That works on all OS types with all variations on how file paths can be specified. The only downside of non-normalized file paths is that the caching mechanism will not consider two file paths as equal that reference the same file in different ways. But that is only a very small problem, IMO.

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -125,6 +134,51 @@ def execute_startup_script(file_path, verbose):
...
+    Returns: list of resolved mock server definition components derived
+       from the input parameter mock_server.
+
+    Raises:
+      click.ClickException for any errors.
+    """
+    resolved_mock_server = []
+    assert isinstance(mock_server, tuple)
+    # Normalize paths for different OS's
+    mock_server_path = [os.path.normpath(fn) for fn in mock_server]
```
andy-maier, line 158: On your use of `os.path.normpath()`: Its description says "This string manipulation may change the meaning of a path that contains symbolic links.". So while I like the idea, it is potentially problematic.

Why do we do anything at all? The user ought to specify a valid path name. If the file path references a file then it works and if not then it fails. That works on all OS types with all variations on how file paths can be specified. The only downside of non-normalized file paths is that the caching mechanism will not consider two file paths as equal that reference the same file in different ways. But that is only a very small problem, IMO.

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/_context_obj.py
andy-maier: Very strange. On Python 3.6.7 on Ubuntu in Travis (with pylint 2.4.4), the issue does not show up: https://travis-ci.org/github/pywbem/pywbemtools/jobs/716348761#L1749

Could you please recreate your Python virtualenv and run `make pylint`? It should install all required dependencies that way.

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: 1. We need to clarify for the user what the path must be.

2. We want to put put the files into the connecitons file with normpath but with forward slashes so the file is portable between systems.

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: I concluded that this is an issue in the particular python environment and it may be because the current python3 installed is 3.6.9 (what occurs on new virtual environment created) but the virtual environment that I normally use reports python 3.6.8.
Looks like ubuntu updated minor version recently.

Removing the pylint disable statements.

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: Did it slightly different.  We normpath when we get from user and again when we upload from yaml. rather than manually playing with the forward and backward slashes.

andy-maier (approved):

PR #742, opened 2020-08-26
Support for caching mock environments of connection definitions

andy-maier:
This PR is good now for serious testing and ready for final review. For details, see the commit message.

Other issues that are handled separately from this PR:
* Issue #747, for updating the docs.
* Issue #748, for handling dependent files.
* Issue #749, for using the released versions of pywbem and nocaselist.

When trying this PR, you need to perform a `make install` to get the repo versions of pywbem and nocaselist that are needed as long as they are not released.

Two of the test mock scripts have been copied to "*_v1s.py" to implement the new setup approach with the `setup()` function, and either of them can be used to try out the caching, for example as follows:

```
$ pywbemcli -m tests/unit/simple_mock_model.mof -m tests/unit/simple_mock_invokemethod_v1s.py connection save simple

$ rm -rf ~/.pywbemcli_mockcache/    # removes the mock cache

$ pywbemcli -v -n simple class tree
Connection repository with 4 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
Mock environment for connection definition 'simple' will be built because it was not cached.
Mock environment for connection definition 'simple' has been written to cache.
root
 +-- CIM_Foo
     +-- CIM_Foo_sub
     |   +-- CIM_Foo_sub_sub
     +-- CIM_Foo_sub2

$ pywbemcli -v -n simple class tree
Connection repository with 4 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
Mock environment for connection definition 'simple' has been loaded from cache.
root
 +-- CIM_Foo
     +-- CIM_Foo_sub
     |   +-- CIM_Foo_sub_sub
     +-- CIM_Foo_sub2
```

**DISCUSSION:**
* Do we agree on the new provider setup interface. -> Yes we do.
* Do we need a Python namespace for each mock script, instead of having all in the same namespace. -> basically yes.


KSchopmeyer:
Questions:

1. Does this mean that the user could implement both the setup method for provider init and the old method with a version flag so that it would operate with python pywbem 1.0.0 and pywbem 1.1.0 or greater. I am not saying they will or that there are that many users of the mock but a question to have in our pockets just in case the question comes up since we are creating a significant incompatibility in a minor version.

2.

andy-maier:
Updated the PR to support both the old and new type of setup for the mock scripts:
* Mock scripts that have no `setup()` function will cause the mock environment not to be cached, and are setup upon import via the global variables CONN, SERVER, VERBOSE. A deprecation warning is issued.
* Mock scripts that have a `setup()` function will not prevent the mock environment from being cached, and are setup via the `setup()` function.

andy-maier:
I implemented support for having each mock script in its own module namespace.

andy-maier:
```
>>> os.path.relpath('c:/abc', 'a:/')
  File "<stdin>", line 1, in <module>
  File "...\Python\Python37\lib\ntpath.py", line 564, in relpath
    path_drive, start_drive))
ValueError: path is on mount 'c:', start on mount 'a:'
```

andy-maier:
Added testcases for building the mock environment.

andy-maier:
I fixed the issue on Windows. It was caused by a different behavior of tell() after opening the file in append mode, where this returns the end of file position on MacOS and Linux, but the begin of file on Windows. This caused the MOF file to be truncated to zero instead of just removing the added line.

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -288,41 +290,240 @@ def PyWbemcliQueryInstances(self, FilterQueryLanguage, FilterQuery,
...
+        # Check that the input files exist. Since we loop through them multiple
+        # times, we check that once.
         for file_path in file_path_list:
             if not os.path.exists(file_path):
-                raise IOError('No such file: {}'.format(file_path))
+                raise mockscripts.MockFileError(
+                    "Mock file does not exist: {}".format(file_path))
+
+        # The connections file is set if a named connection is used, i.e.
+        # when specifying the -n general option. It is unset when the -s or -m
```
KSchopmeyer, line 380: I suggest that we change the word unset to the phrase not set

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -288,41 +290,240 @@ def PyWbemcliQueryInstances(self, FilterQueryLanguage, FilterQuery,
...
+        # general options were specified. In the latter case, no caching happens
```
KSchopmeyer, line 381: Change "In the latter case" to "when -n is not set". Reads better.

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -288,41 +290,240 @@ def PyWbemcliQueryInstances(self, FilterQueryLanguage, FilterQuery,
...
+        # because there is no connection definition context under which that
```
KSchopmeyer, line 382: Suggest change "under which that could be done" to "which is required for caching"

I am trying to get close to clean statement that "caching is only executed for named connections"

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -288,41 +290,240 @@ def PyWbemcliQueryInstances(self, FilterQueryLanguage, FilterQuery,
...
+                if connections_file:
+                    self._dump_mockenv(pickle_file)
+                    with open(md5_file, 'w') as fp:
+                        fp.write(new_md5_value)
+                    if verbose:
+                        click.echo("Mock environment for connection "
+                                   "definition '{}' has been written to "
+                                   "cache.".format(connection_name))
+        else:
+            # If we can use the cache, there was always a connections file
```
KSchopmeyer, line 479: I assume this comment is meant to say "There must be a connections file for caching to occur"

tests/unit/simple_mock_invokemethod_v1.py
```diff
@@ -1,23 +1,26 @@
 """
-mock_pywbem test script that installs a a method provider for the class
-CIM_Foo
-"""
+Test mock script that installs a test method provider for CIM method
```
KSchopmeyer, line 2: Suggestion on naming of the python files for this and the following.  Change the last prefix to aimplw_moxk_invokwmwrho_v1_old and ..._new since "-s" does not have a lot of meaning.

tests/unit/simple_mock_invokemethod_v1s.py
```diff
@@ -0,0 +1,87 @@
+"""
```
KSchopmeyer, line 1: See comment above about changing the file name

tests/unit/test_pywbemcli_operations.py
```diff
@@ -0,0 +1,852 @@
...
+    SetupNotSupportedError
+
+from .pytest_extensions import simplified_test_function
+from .utils import captured_output
+
+OK = True
+FAIL = False
+PDB = "pdb"
+
+# Click (as of 7.1.2) raises UnsupportedOperation in click.echo() when
```
KSchopmeyer, line 42: MIght note that this is apparently fixed for click 8.0

KSchopmeyer (commented):
I did a code review and made some comments on documentation.

Here are a couple of more general comments on usability:

1. How do we ever get rid of a cache file if, for example the named connection is deleted.  DISCUSSION

2. If a user sets up a local cache file and then removes it, those caches will remain forever partly because with the cache file using a hash of the connections file, the user does not know which files to eliminate.

DISCUSSION. partly because I do not yet have a good answer.

All this presumes that the windows tell bug goes away. :-)

pywbemtools/pywbemcli/_pywbemcli_operations.py
andy-maier: DONE

pywbemtools/pywbemcli/_pywbemcli_operations.py
andy-maier: DONE - changed to "When no connections file is set,"

pywbemtools/pywbemcli/_pywbemcli_operations.py
andy-maier: DONE

pywbemtools/pywbemcli/_pywbemcli_operations.py
andy-maier: Close but not exactly. If you want to phrase it this way around, then it would say: "There must be a connections file when no rebuild was needed."

DONE - Changed it to: "When no rebuild was needed, there must have been a connections file set."

tests/unit/simple_mock_invokemethod_v1.py
andy-maier: DONE. Renamed as suggested.

tests/unit/test_pywbemcli_operations.py
andy-maier: Who knows. Click 8.0 is not relased yet, and this goes on for quite some time now. The Click issue is referenced in the text and that allows figuring out whether and when the fix is released. NOCHANGE.

andy-maier:
On the discussion points:

* How do we ever get rid of a cache file if, for example the named connection is deleted.
  -> We can delete the cache when the connection is deleted.

* If a user sets up a local cache file and then removes it, those caches will remain forever partly because with the cache file using a hash of the connections file, the user does not know which files to eliminate.
  -> I assume you mean when a user specifies a local connections file? I see these options:
  - No caching when user-defined connections file is used.
  - We change the cache directory name to be the relative directory path of the connections file with special chars translated to minus, plus the connection name.

In addition, we could add a "connection cleanupcache" command.

**DISCUSSION**

andy-maier:
We tend to think that when a user-defined connections file is used, we would not cache. Ready for implementing, final decision when approving the PR.

andy-maier:
I have implemented that the mock env is not cached when user-specified connections files are used.

andy-maier:
I have fixed the error, and have added support for taking into account the dependent files registered for a mock script, based on the new provider dependent registry of the pywbem mock support.

andy-maier:
At this point, the implementation is complete w.r.t. what we had discussed, and all tests have been written.

This PR currently depends on pywbem PR https://github.com/pywbem/pywbem/pull/2443 and pulls it in from its branch, and therefore it should be merged only after that PR is merged (after being updated to use master again).

Please do a final review.

andy-maier:
Changed to use pywbem master branch again, after merging pywbem PR pywbem/pywbem#2443 .

KSchopmeyer (commented):
DISCUSSION: I only asked two questions.  One is simply my memory, the order of parameters on a class declaration.  The second is what is up for discussion.  Shuld the BuildRepositoryMixin be in its own module?  Even if we answer yes, it would not bother me to put that into another issue.

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -288,41 +290,240 @@ def PyWbemcliQueryInstances(self, FilterQueryLanguage, FilterQuery,
         return list(result)
 
 
-class BuildRepositoryMixin(object):
+class BuildMockenvMixin(object):
```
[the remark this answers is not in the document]
andy-maier: We discussed this and agreed to have such structural changes in a separate PR.

pywbemtools/pywbemcli/_pywbemcli_operations.py
```diff
@@ -383,7 +633,7 @@ def __init__(self, *args, **kwargs):
 
 class PYWBEMCLIFakedConnection(pywbem_mock.FakedWBEMConnection,
                                PYWBEMCLIConnectionMixin,
-                               BuildRepositoryMixin):
+                               BuildMockenvMixin):
```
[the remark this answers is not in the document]
andy-maier: Agreed, this should be turned around.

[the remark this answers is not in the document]
andy-maier: DONE

PR #79, opened 2017-10-26
Manual Travis+Appveyor CI runs with full set of environments

andy-maier:
**!! Do not merge this PR !! This PR stays on its branch and is never merged.**

The change in this PR causes Travis and Appveyor to run with the full set of test environments, compared to the other branches which run with a minimal set of test environments.

**Usage:**

* This PR is run automatically on Travis as a daily build.
* To trigger a manual build on both Travis and Appveyor, rebase the branch of this PR to master:
  ```
  git checkout master
  git pull
  git checkout manual-ci-run
  git rebase master
  git push -f
  ```
* For build results, see branch `manual-ci-run` in https://travis-ci.org/pywbem/pywbemtools/branches.

Note: If you make changes to .travis.yml, the automatic daily build on Travis will not pick them up until you rebase this PR on them.

PR #753, opened 2020-09-07
Test: Added testcases for Python tell/truncate

andy-maier:
The real test for this is in PR #79 which has been rebased on this PR.
Update: That test succeeded for all environments, see particularly Appveyor build https://ci.appveyor.com/project/KSchopmeyer/pywbemtools/builds/35071137

**DISCUSSION**: What should we do with this test, keep it in pywbemtools?

### DIFF
--- a/tests/unit/test_python_tell.py
+++ b/tests/unit/test_python_tell.py
@@ -1,0 +1,90 @@
+"""
+Tests for Python tell() behavior.
+
+This test exists because certain uses of tell() behave differently between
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
My tendency would be to keep it. Does not hurt anything and continues to point out the issue.  I assume we remove it when we remove support for python 2.7

andy-maier:
We decided to keep it, and to document why it is there

andy-maier:
Documented why it is there. Karl approved it verbally in the meerting.

PR #767, opened 2020-09-30
Fixes #766: Pywbemcli tests fail, assertion mof_compiler

KSchopmeyer:
The tests actually fail because of a badly defined qualifier declaration
in the model qualifier_filter_model.mof.  The Indication qualifier
declaration is invalid it this test (scope should be class, indication).

Changes in pywbem PR 2456, issue #2455 added the scope validity test to class resolution and caught the error

Fixed.

The secondary issue of an assertion error is an issue in pywbem, not
pywbemcli

NOTE: cmdshelp.rst was also modified because we must have missed that
change in earlier commit.   Build doc made the changes seen.

PR #770, opened 2020-10-01
Fixed small issues in makefile; Detecting repo files not up to date

andy-maier:
See commit message.


andy-maier:
Since this PR and PR #767 depend on each other and both fail because the other is not in, I am merging this PR without further review to break that cycle.

PR #786, opened 2020-10-05
0.7: Fixed incorrect property order in instance table output

andy-maier:
Rolls back PR #785 into 0.7.2.


andy-maier:
Abandoned, since the 0.7 version requires more than that and the testcase that fails does not yet exist there, so the incorrect ordering is just a slight inconvenience on 0.7 that will then be fixed in 0.8.

andy-maier:
Reopening this PR because the tests of pywbemtools 0.7.2 fail against pywbem master (the upcoming 1.1.0)

Issue #822, opened 2020-12-15
pywbemcli fails test with mock package 4.0.3.


KSchopmeyer:
When used with mock version 4.0.3, pywbemcli fails consistently in two different operating  modes:

1. pywbemcli < x.txt   # Entering commands with stdin.
2. interactive mode after a command has been executed (but it does not fail if the exit occurs before any command has been executed)

The following is an example of the failure with stdin but the failure itself is the same for both modes, a variable is not defined apparently because the mock __exit__ is executed without the corresponding  __enter__ being executed.

```
  pywbemcli < "class -h"

where the failure is in the exit mechanisms upon completing the stdin and the actuall error is:

```
    File "/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/mock/mock.py", line 1529, in __exit__
      if self.is_local and self.temp_original is not DEFAULT:
   AttributeError: '_patch' object has no attribute 'is_local' 

The complete exception trace is:

```

(pycli3) kschopmeyer@leonard:~/pywbem/pywbemtools(ks/#817-class-tree-display-details)$ pywbemcli < x.txt
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli or <CTRL-r> to search history, 
Usage: pywbemcli [GENERAL-OPTIONS] class COMMAND [ARGS] [COMMAND-OPTIONS]

  Command group for CIM classes.

  This command group defines commands to inspect classes, invoke methods on
  classes, delete classes.

  Creation and modification of classes is not currently supported.

  In addition to the command-specific options shown in this help text, the
  general options (see 'pywbemcli --help') can also be specified before the
  'class' keyword.

Command Options:
  -h, --help  Show this help message.

Commands:
  enumerate     List top classes or subclasses of a class in a namespace.
  get           Get a class.
  delete        Delete a class.
  invokemethod  Invoke a method on a class.
  references    List the classes referencing a class.
  associators   List the classes associated with a class.
  find          List the classes with matching class names on the server.
  tree          Show the subclass or superclass hierarchy for a class.
Traceback (most recent call last):
  File "/home/kschopmeyer/virtualenvs/pycli3/bin/pywbemcli", line 33, in <module>
    sys.exit(load_entry_point('pywbemtools', 'console_scripts', 'pywbemcli')())
  File "/home/kschopmeyer/virtualenvs/pycli3/lib/python3.6/site-packages/click/core.py", line 829, in __call__
[...]

PR #823, opened 2020-12-16
Mitigates #822: causing test failure because of mock version update

KSchopmeyer:
The update to mock 4.0.3 causes at least one test to fail.  This pr limits the version of the mock package to lless than version 4.0.3


KSchopmeyer:
Note that the tests are failing because of issue with PIP that I documented in issue #824.

andy-maier:
I rebased this on master after the windows failure was fixed.

andy-maier (approved):

PR #839, opened 2021-01-01
Fixes #822: by rewriting warning message code

### LINKED ISSUE #822: pywbemcli fails test with mock package 4.0.3.

KSchopmeyer:
See commit for details.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -28,6 +28,16 @@ Released: not yet
 * Limit mock package to lt 4.0.3 to avoid issue issue that causes test failure.
   (see #822)
 
+* Fix issue caused by mock package version 4.0.3 by creating replacements for
+  warnings.warn and warnings.warn_explicit functions  and removing the use of
+  the patch decorator in pywbemcli.py before the definition of the cli
+  function.  (see issue #822)
+
+* Fixes issue where in pywbemcli the --timeout and --use-pull general
+  options were not always correctly included in the new object context in
+  interactive mode if they were specified on the interactive mode cmd line.
+
+
 **Enhancements:**
 
 * Add new option to class find command (--summary) to display a summary of
--- a/pywbemtools/pywbemcli/_context_obj.py
+++ b/pywbemtools/pywbemcli/_context_obj.py
@@ -54,6 +54,8 @@ class ContextObj(object):  # pylint: disable=useless-object-inheritance
         self._log = log
         self._verbose = verbose
         self._pdb = pdb
+        # warn included only to insure that original cmd line input flag
+        # maintained through interactive session.
         self._warn = warn
         self._connections_repo = connections_repo
 
--- a/pywbemtools/pywbemcli/_utils.py
+++ b/pywbemtools/pywbemcli/_utils.py
@@ -20,6 +20,7 @@ Utility Functions applicable across multiple components of pywbemcli.
 from __future__ import print_function, absolute_import
 
 import os
+import warnings
 import six
 import click
 
@@ -108,12 +109,30 @@ def _eq_name(name1, name2):
     return name1.lower() == name2.lower()
 
 
-def formatwarning(message, category, filename, lineno, line=None):
+def _formatwarning(message, category, filename, lineno, line=None):
     # pylint: disable=unused-argument
     """
-    Replacement for warnings.formatwarning() that is patched in.
+    Replacement for warnings.formatwarning() that is monkey patched in.
     """
     return "{}: {}\n".format(category.__name__, message)
+
+
+def pywbemcliwarn(*args, **kwargs):
+    """
+    Pywbemcli monkey patch for the warnings.warn function; substitutes our
+    own formatter.
+    """
+    warnings.formatwarning = _formatwarning
+    warnings.warn(*args, **kwargs)
+
+
+def pywbemcliwarn_explicit(*args, **kwargs):
+    """
+    Pywbemcli monkey patch for the warnings.warn_explicit function;
+    substitutes our own formatter
+    """
+    warnings.formatwarning = _formatwarning
+    warnings.warn_explicit(*args, **kwargs)
 
 
 def get_terminal_width():
--- a/pywbemtools/pywbemcli/pywbemcli.py
+++ b/pywbemtools/pywbemcli/pywbemcli.py
@@ -25,7 +25,6 @@ import sys
 import traceback
 from copy import deepcopy
 import warnings
-from mock import patch
 import click
 import click_repl
 from prompt_toolkit.history import FileHistory
@@ -46,7 +45,7 @@ from .config import DEFAULT_NAMESPACE, PYWBEMCLI_PROMPT, \
 from ._connection_repository import ConnectionRepository, \
     ConnectionsFileError
 from ._click_extensions import PywbemcliTopGroup
-from ._utils import formatwarning, get_terminal_width, CONNECTIONS_FILENAME, \
+from ._utils import pywbemcliwarn, get_terminal_width, CONNECTIONS_FILENAME, \
     DEFAULT_CONNECTIONS_FILE
 
 
@@ -366,10 +365,10 @@ def set_default_if_empty_str(tst_str, default=None):
                           ev=PywbemServer.log_envvar,
                           default='all'))
 @click.option('-v', '--verbose/--no-verbose',
-              default=None,
+              default=None,  # None separates no option from --no-verbose
               help=u'Display extra information about the processing.')
 @click.option('--warn/--no-warn', is_flag=True,
-              default=False,
+              default=None,  # None separates no option from --no-warning
               help=u'Warnings control: True enables display of all Python '
               u'warnings; False leaves warning control to the PYHONWARNINGS '
               u'env var, which by default displays no warnings. '
@@ -400,7 +399,6 @@ def set_default_if_empty_str(tst_str, default=None):
     help=u'Show the version of this command and the pywbem package.')
 @add_options(help_option)
 @click.pass_context
-@patch('warnings.formatwarning', formatwarning)
 def cli(ctx, server, connection_name, default_namespace, user, password,
         timeout, verify, certfile, keyfile, ca_certs, output_format, use_pull,
         pull_max_cnt, mock_server, verbose=None, connections_file=None,
@@ -696,8 +694,10 @@ def cli(ctx, server, connection_name, default_namespace, user, password,
                     modified_server = True
                 if timeout:
                     pywbem_server.timeout = resolved_timeout
+                    modified_server = True
                 if use_pull:
                     pywbem_server.use_pull = resolved_use_pull
+                    modified_server = True
                 if server:
                     pywbem_server.server = server
                     modified_server = True
@@ -706,6 +706,7 @@ def cli(ctx, server, connection_name, default_namespace, user, password,
                         set_default_if_empty_str(default_namespace,
                                                  DEFAULT_NAMESPACE)
                     modified_server = True
+
                 if modified_server:
                     pywbem_server.reset()
                 else:
@@ -755,6 +756,10 @@ def cli(ctx, server, connection_name, default_namespace, user, password,
         if warn is None:
             warn = ctx.obj.warn
 
+    # Conditionally set the flag to enable warnings
+    if warn:
+        warnings.simplefilter('once')
+
     # Create a command context for each command: An interactive command has
     # its own command context as a child of the command context for the
     # command line.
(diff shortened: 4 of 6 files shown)


andy-maier:
Update needed:
* Please rebase on master to pick up latest fixes
* Resolve pylint issue:
  ```
  pywbemtools/pywbemcli/_utils.py:24:0: C0411: standard import "import warnings" should be placed before "import six" (wrong-import-order)
  ```

pywbemtools/pywbemcli/_utils.py
```diff
@@ -108,14 +109,50 @@ def _eq_name(name1, name2):
...
+    own formatter.
+    """
+    # Import for each call
+    # pylint: disable=import-outside-toplevel, cyclic-import
+    from .pywbemcli import WARN_OPTION
+    if WARN_OPTION:
+        warnings.simplefilter('once')
+    # else: Leave warning control to the PYTHONWARNINGS env var.
+    # Delete the variable so it is imported with next call
+    del WARN_OPTION
```
andy-maier, line 133: This `del` deletes the name `WARN_OPTION` from the local scope of the function, not from the pywbemcli module's namespace. I'm not sure I understand what the purpose is, since it will be gone anyway once the function returns?

tests/unit/test_general_options.py
```diff
@@ -333,15 +336,23 @@ class       Command group for CIM classes.
...
+     None, OK],  # Only tests that the option is accepted
+
+    ['Verify --warn general option.',
+     {'general': ['-s', 'http://blah', '--warn'],
+      'cmdgrp': 'connection',
+      'args': ['show']},
+     {'stderr': ["Pywbemcli support for Python", "is deprecated"],
+      'rc': 0,
+      'test': 'innows'},
+     None, PYTHON2],  # Only tests that the option is accepted
```
andy-maier, line 355: This variant of the testcase really wants to test the deprecation message. That message is issued on Python 2.7 and 3.4, but the testcase is only run on Python 2.7. I think the `PYTHON2` variable should be changed to become `PYTHON_DEPRECATED` or so, and be True for both of these Python versions.

requirements.txt
```diff
@@ -50,7 +50,7 @@ packaging>=17.0
 
 # See issue #822 about issue in mock 4.0.3.
 mock>=3.0.0,<4.0.0; python_version < '3.6'
-mock>=3.0.0,<4.0.3; python_version >= '3.6'
+mock>=3.0.0; python_version >= '3.6'
```
andy-maier, line 53: When combining the two commits, the merge conflict has been resolved. Please squash the two commits into one, it does not really make sense to have them separated.

pywbemtools/pywbemcli/_utils.py
```diff
@@ -108,14 +109,50 @@ def _eq_name(name1, name2):
...
+def pywbemcliwarn(*args, **kwargs):
+    """
+    Pywbemcli replacement for the warnings.warn function that substitutes our
+    own formatter.
+    """
+    # Import for each call
+    # pylint: disable=import-outside-toplevel, cyclic-import
+    from .pywbemcli import WARN_OPTION
+    if WARN_OPTION:
+        warnings.simplefilter('once')
```
andy-maier, line 130: Which problem is solved by moving the `warnings.simplefilter()` call to here? It introduces the need for the global `WARN_OPTION` so let's do that only if there is a good reason for it.

pywbemtools/pywbemcli/_utils.py
```diff
@@ -108,14 +109,50 @@ def _eq_name(name1, name2):
...
+    # else: Leave warning control to the PYTHONWARNINGS env var.
+    # Delete the variable so it is imported with next call
+    del WARN_OPTION
+
+    warnings.formatwarning = _formatwarning
```
andy-maier, line 135: So the solution is to keep it patched ? Just asking.

docs/changes.rst
```diff
@@ -25,8 +25,12 @@ Released: not yet
...
   (see issues #816 and #817)
 
-* Limit mock package to lt 4.0.3 to avoid issue issue that causes test failure.
-  (see #822)
+* Fix issue with failure in mock package version 4.0.3 by creating local
+  warnings.warn ... functions to implement our own functions that call the
+  Python warnings functions and  to remove the use of the patch decorator
+  before the definition of the cli function.  Note this superceeds previous
+  change that limited versions of mock package. (see issue #822)
+
```
andy-maier, line 33: I'm not happy with this change log entry:
* The first sentence is overly complicated ("by creating local warnings.warn ... functions to implement our own functions that call the Python warnings functions").
* This change log replaces the prior change log about #822, so it does not make sense for the reader to reference the now deleted previous change log entry.

pywbemtools/pywbemcli/_context_obj.py
```diff
@@ -402,7 +405,8 @@ def connect_wbem_server(self):
                 log=self.log,
                 use_pull=self.use_pull,
                 timestats=self.timestats,
-                verbose=self.verbose)
+                verbose=self.verbose,
+                warn=self.warn)
```
andy-maier, line 409: First, there is a second call to `create_connection()` in this module. But foremost: The `create_connection()` method does not accept a `warn` parameter.

andy-maier (changes requested):
See comments. Also, please squash the two commits into one.

pywbemtools/pywbemcli/_utils.py
```diff
@@ -108,14 +109,48 @@ def _eq_name(name1, name2):
     return name1.lower() == name2.lower()
 
 
-def formatwarning(message, category, filename, lineno, line=None):
+def _formatwarning(message, category, filename, lineno, line=None):
     # pylint: disable=unused-argument
     """
-    Replacement for warnings.formatwarning() that is patched in.
+    Replacement for warnings.formatwarning() that replaces the standard
+    warning;.formatwarning for pywbemcli.
```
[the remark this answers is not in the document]
KSchopmeyer: DONE.  Agree that original sounds fine also

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -696,8 +691,10 @@ def create_server_instance(connection_name):
                     modified_server = True
                 if timeout:
                     pywbem_server.timeout = resolved_timeout
+                    modified_server = True
```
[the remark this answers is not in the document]
KSchopmeyer: DONE

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: Done.   My error removing the first attempt to fix this by passing warn around.

docs/changes.rst
KSchopmeyer: Fixed.

tests/unit/test_general_options.py
```diff
@@ -34,6 +34,9 @@
 
 TEST_DIR = os.path.dirname(__file__)
 
+PYTHON3 = sys.version_info[0:2] >= (3, 5)
```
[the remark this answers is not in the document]
KSchopmeyer: The price of getting old.  Did not even think of the obvious.   I added tests to determine exactly on which versions one of the warnings is generated/ not generated but make the validation for exactly the same versions as the definition of the warning message.

pywbemtools/pywbemcli/_utils.py
KSchopmeyer: That one was strange but it forces the import to be redone.  The wonders of stackOverflow.  That causes the reimport to get the current value of the import.  Yes, it apparently works but I hated it the minute I used it.  However, once I determined better what the scope of the warnings filters was all this was not needed so removed all of the export WARN_OPTION, etc.

pywbemtools/pywbemcli/_utils.py
KSchopmeyer: I defy you to find out exactly what  the scope of the filters in warnings is from the documentation.  I am still not sure but it appears that it applies across either a context in which warnings is applied or the complete package and the multiple "import warnings" in pywbemcli.py and _utils.py are irrelevant. O learn.

Removed the global WARN_OPTION, and the setting of the WARN_OPTION in _utils.py.

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -369,7 +369,7 @@ def set_default_if_empty_str(tst_str, default=None):
               default=None,
               help=u'Display extra information about the processing.')
 @click.option('--warn/--no-warn', is_flag=True,
-              default=False,
+              default=None,
```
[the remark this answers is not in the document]
KSchopmeyer: Complex question.  If it is included in context object, there should be a None to indicate that there was no input since --warn and --no-warn represent option input and option state for this command The only reason we include it in the context object is because in interactive mode if you change it that should only apply to that single command so need the original in the top level context object (Note that need only be in the top level context object in reality but since click builds multiple levels, it was easy to include it in each context object.  Therefore, we need to know  what the original setting was and fall back to that after each interactive mode command.

tests/unit/test_general_options.py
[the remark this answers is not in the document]
KSchopmeyer: DONE

requirements.txt
KSchopmeyer: DONE

andy-maier (approved):

PR #871, opened 2021-01-10
[WIP] Fixes #761: Add comment field to connection definition

### LINKED ISSUE #761: Add capability to include a comment for connection definitions in connections file
As the number of connection definitons grows in a users connections file, the more difficult it will be to separate them simply on the name (it should be short to be typeable but long to be descriptive).  It would be useful if the user could maintain a comment entity in the file for each connection that would be a string defined by the user.

KSchopmeyer:
**WIP incomplete. See below for discussion items.**

Adds a new field to connection definitions and a new command to the
connection group to edit data into the field.

connection edit allows editing the comment of any defined connection and
allows re-editing the text at will.

The editor  used is built into prompt-toolkit and allows multiline
editing including adding and deleting lines, moving the cursor with
arrow keys or the mouse, and saving the resulting comment.

This enhances the server definition with an additional field, comment, where the user can include comments on the server definition and a command "connection comment" that allows editing a multiline comment.

An editing session might look as follows with instructions on how to control the edit process.  This allows the string to be created and edited.  It is automatically saved in the svr definition at the end of each edit session.:```

```

pywbemcli > connection comment

0: mock1
1: mock2
Input integer between 0 and 1 or Ctrl-C to exit selection: 1
Edit/create comment connection: "mock2".
Enter creates new line, arrow keys and mouse move cursor,
[Meta+Enter] or [Esc] followed by [Enter] accept edit. Ctrl-C aborts the edit.
Comment: Hi There. This is more comment
       : Added more comment. Line 3.
       : Now added line 4 again

```

`**`LIMITATIONS/DISCUSSION**

the following are a number of the unresolved issues.

1. In retrospect I would prefer to call this notes rather than comment.  a) you could have multiple notes, b) This implies that things could be added.
2. There is no general option to manipulate a comment (i.e. you cannot add a comment on the command line).  That gets messy because it would probably only be used to completely replace a comment or to append to one and therefore is limited compared to the multiline editor.
3. There is no max width to the lines of edit text at this point and the connection list does not reformat other than width in which the user edits the text. You can make the comments as wide as you want and that is how they show up in the display.  It would be logical to set a maximum of some sort and to clean up the relation between NLs  by the editor and a max width.
4. The comment only shows up in the connection list --full option. To me it would be logical to include it in the connection list default (brief mode) since if used, this information could be important in selecting/understanding connections..
5. No tests written because of the number of outstanding decisions to be made.
6. We have only two modes of selecting the connection to be edited "connection comment": presents pick list as shown in example above and 2) "connection comment NAME" that edits the  NAME connection.  There is no selection mode to select the current connection at this point.  This is because:
  a. Wanted to be sure this worked first.
  b. This is consistend with delete where you have choice of select on NAME but no option to delete current connection.
7. Currently the comment is tied to an already saved server definition. 

```
WBEM server connections(full): (#: default, *: current)
file: test_connections.yaml
name     server    namespace    user      timeout  use-pull    pull-max-cnt    verify    certfile    keyfile    mock-server                       comment
-------  --------  -----------  ------  ---------  ----------  --------------  --------  ----------  ---------  --------------------------------  -------------------------------
*#mock1            root/cimv2                  30                              True                             tests/unit/simple_mock_model.mof  Hi There.
                                                                                                                                                  Line 2
                                                                                                                                                  line 3,
                                                                                                                                                  line 4 added.
mock2              root/cimv2                  30                              True                             tests/unit/simple_mock_model.mof  Hi There. This is more comment
                                                                                                                                                  Added more comment. Line 3.
                                                                                                                                                  Now added line 4 again


       :

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -53,6 +53,10 @@ Released: not yet
 * Migrated from Travis and Appveyor to GitHub Actions. This required several
   changes in package dependencies for development.
 
+* Add comment field to connection definition and command "connection comment"
+  to edit this field for any defined connection definition. (see
+  issue #761)
+
 **Known issues:**
 
 * See `list of open issues`_.
--- a/docs/pywbemcli/commands.rst
+++ b/docs/pywbemcli/commands.rst
@@ -1317,6 +1317,9 @@ The attributes of each connection definition in the :term:`connections file` are
 
 The commands in this group are:
 
+* :ref:`Connection comment command` - Create/modify a comment for a WBEM connection definition.
+
+* :ref:`Connection comment command` - Edit comment in WBEM connection definition.
 * :ref:`Connection delete command` - Delete a WBEM connection definition.
 * :ref:`Connection export command` - Export the current connection.
 * :ref:`Connection list command` - List the WBEM connection definitions.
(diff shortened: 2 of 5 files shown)


KSchopmeyer:
Will not do this one

PR #878, opened 2021-01-22
Added namespace command group

andy-maier:
Ready for review and test.

Now uses pywbem from master again.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -16,6 +16,10 @@ Released: not yet
 
 **Deprecations:**
 
+* Deprecated the 'server namespaces' and 'server interop' commands. Use the
+  new commands 'namespace list' and 'namespace interop', respectively.
+  (issue #877)
+
 **Bug fixes:**
 
 * Fixed a ValueError on Windows that was raised when the connections file was
@@ -74,6 +78,11 @@ Released: not yet
 
 * Enabled the tests for Python 3.4 on Windows again - this required
   some changes in the Makefile and constraints files.
+
+* Added a 'namespace' command group that allows listing, creating and deleting
+  CIM namespaces, and showing the Interop namespace. The 'server namespaces'
+  and 'server interop' commands that provide a subset of that functionality
+  have been deprecated. (issue #877)
 
 **Known issues:**
 
--- a/docs/pywbemcli/cmdshelp.rst
+++ b/docs/pywbemcli/cmdshelp.rst
@@ -141,6 +141,7 @@ Help text for ``pywbemcli``:
     Commands:
       class       Command group for CIM classes.
       instance    Command group for CIM instances.
+      namespace   Command group for CIM namespaces.
       profile     Command group for WBEM management profiles.
       qualifier   Command group for CIM qualifier declarations.
       server      Command group for WBEM servers.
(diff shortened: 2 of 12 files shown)


docs/changes.rst
```diff
@@ -70,6 +74,11 @@ Released: not yet
 * The verbose option ('-v' / '--verbose') now also displays the objects that
   are compiled into a mock environment when setting it up.
 
+* Added a 'namespace' command group that allows liting, creating and deleting
```
KSchopmeyer, line 77: spelling (liting)

docs/pywbemcli/cmdshelp.rst
```diff
@@ -1547,6 +1548,152 @@ Help text for ``pywbemcli instance shrub`` (see :ref:`instance shrub command`):
...
+Help text for ``pywbemcli namespace`` (see :ref:`namespace command group`):
+
+
+::
+
+    Usage: pywbemcli [GENERAL-OPTIONS] namespace COMMAND [ARGS] [COMMAND-OPTIONS]
+
+      Command group for CIM namespaces.
+
+      This command group defines commands to create, delete and list namespaces in a WBEM server.
```
KSchopmeyer, line 1567: I note that we use the phrase CIM namespace once and thereafter it is always namespace.  Should we note 1. what is CIM namspace 2) namespace refers to CIM namespace.?

KSchopmeyer:
I think this one is OK.   Tested as far as I could against OpenPegasus. I want to run another test against pegasus in a docker image since there I can test create and delete of namespaces.  Will try to do that tonight.  

Note that most of the existing test scripts now fail any namespace command and return the Model error because they do not have an interop namespace but that is a separate issue.

pywbemtools/pywbemcli/_cmd_namespace.py
```diff
@@ -0,0 +1,210 @@
...
+def namespace_group():
+    """
+    Command group for CIM namespaces.
+
+    This command group defines commands to create, delete and list namespaces
+    in a WBEM server.
+
+    In addition to the command-specific options shown in this help text, the
+    general options (see 'pywbemcli --help') can also be specified before the
+    'class' keyword.
```
KSchopmeyer, line 57: The word 'class' should be 'namespace' since this is the namespace group

pywbemtools/pywbemcli/_cmd_namespace.py
```diff
@@ -0,0 +1,210 @@
...
+def namespace_create(context, namespace):
+    """
+    Create a namespace on the server.
+
+    Leading and trailing slash (``/``) characters specified in the NAMESPACE
+    argument will be stripped.
+
+    The namespace must not yet exist on the server.
+
+    The Interop namespace must exist on the server and cannot be created using
```
KSchopmeyer, line 93: Maybe this deserves a bit more exlplanation, why it cannot be created. ex. "since that namespace is required to implement client requests to manage namespaces.

KSchopmeyer (commented):
I am generally comfortable with this one after playing with it and reviewing.  Made some editorial comments and marked test needed since I would like to run create/delete test against OpenPegasus in a container.

tests/unit/mock_interop.mof
```diff
@@ -0,0 +1,123 @@
+// This is a simple mof model that creates the qualifier declarations,
+// classes, and instances for a very simplistic model to be used in the
+// pywbemcli mock test environment.
```
KSchopmeyer, line 3: Lets add the following  to the comments. This creates a minimal interop namespace including the class for CIM_Namespace and CIMObjectManager so that commands like namespace list will confirm that the required interop namespace exists.

KSchopmeyer:
Tested a variety of the commands against docker image of OpenPegasus with no problems including creating, populating the namespace and deleting the namespace.

Where we failed is in deleting the namespace with data in it.

1. Can delete the classes but it is one by one. Note that the -f option works so it is simply a case of deleting each class with the --force option.
2. Cannot delete the qualifier declarations.  There is no command to do that.

docs/changes.rst
andy-maier: DONE

docs/pywbemcli/cmdshelp.rst
andy-maier: We are doing that also for other CIM objects. I don't think we need that level of precision in the docs.
What do you think?

pywbemtools/pywbemcli/_cmd_namespace.py
andy-maier: DONE

pywbemtools/pywbemcli/_cmd_namespace.py
andy-maier: DONE

tests/unit/mock_interop.mof
andy-maier: DONE, but with different text:
```
// This is a simple mof model that creates the qualifier declarations,
// classes, and instances for a very simplistic model of the Interop namespace
// to be used in the pywbemcli mock test environment.
// The resulting Interop namespace includes the classes for CIM_Namespace and
// CIM_ObjectManager and an instance of CIM_ObjectManager so that the needs
// of the WBEM server functionality of pywbem are satisfied. The instances
// of CIM_Namespace (including the one for the Interop namespace itself)
// will be created by the namespace provider.
```

andy-maier:
On the discussion items:
- The need for 'qualifier delete' has been documented as issue #884.
- The need for deleting a namespace that is not empty has been documented as issue #885.
- The need to compile MOF has been documented as issue #886.

andy-maier:
At this point, there is only the comment open on explaining the use of CIM object terms without the "CIM" term.

docs/pywbemcli/cmdshelp.rst
KSchopmeyer: OK with me.

KSchopmeyer (approved):

PR #893, opened 2021-02-03
Added end2end tests using Pegasus container image

andy-maier:
See commit message.
Ready for review.
It now has a first end2end test function that just checks that the server is OpenPegasus and provides the expected namespaces and profiles.

**DISCUSSION (RESOLVED):**
* This PR adds the capability to have end2end test cases, but does not define any actual end2end test cases yet. I suggest that the end2end test cases are added in subsequent PRs.
* We have an infrastructure for defining end2end server environments in the pywbem project, but not in pywbemtools. The end2end tests here currently implement a very small scale version of that by means of defining a single environment variable for a single target server. Good enough?
* Right now, the container is created just once in the test workflow. Should we create it new for each end2end testcase, e.g. in a fixture? (takes about 7 sec)

### DIFF
--- a/tests/end2endtest/test_server.py
+++ b/tests/end2endtest/test_server.py
@@ -1,0 +1,81 @@
+# Copyright 2021 IBM Corp. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Basic server tests.
+"""
+
+from __future__ import absolute_import, print_function
+
+import re
+
+# pylint: disable=unused-import
+from .utils import server_url  # noqa: F401
+# pylint: enable=unused-import
+from ..unit.utils import execute_pywbemcli
+
+
+def test_server_is_pegasus(server_url):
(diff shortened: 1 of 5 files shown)


andy-maier:
We discussed and agreed on 1+2 being ok, and on 3, to go with the fixtures(s)

tests/end2endtest/test_server.py
```diff
@@ -0,0 +1,76 @@
...
+# pylint: enable=unused-import
+from ..unit.utils import execute_pywbemcli
+
+
+def test_server_is_pegasus(server_url):
+    # pylint: disable=redefined-outer-name
+    """
+    Test that this WBEM server is OpenPegasus and that it provides the
+    expected Interop namespace, and the namespaces and profiles we want to test
+    against.
```
KSchopmeyer, line 34: Add this functionality is provided by the OpenPegasus  make build, make repository, make test repository including the building of repositories, providers, and namespaces. It is also based on OpenPegasus use of the DMTF schema version 2.4.1

KSchopmeyer (commented):
One comment and approve

tests/end2endtest/test_server.py
andy-maier: CIM Schema version 2.4.1 is certainly not correct. I checked all namespaces and the latest Version qualifier says 2.39.0.

For now, I have added the following paragraph:

```
    This container-packaged version of OpenPegasus is provided by the
    OpenPegasus project on GitHub (https://github.com/OpenPegasus/OpenPegasus).
    It includes OpenPegasus 14.2 (version string 2.14.2), the CIM Schema
    version 2.39.0, and a number of providers including a namespace provider.
```

KSchopmeyer: Karl has mixup in repo naming.  Clarify.

andy-maier: DONE - adjusted proposed text to use Pegasus version 2.14.2, and CIM Schema 2.41.0.

Issue #883, opened 2021-01-26
New "Exception ignored" and warning messages


KSchopmeyer:
Effective today I am gettingthe following when leaving interactive mode

Exception ignored in: <socket.socket fd=6, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('127.0.0.1', 33414), raddr=('127.0.0.1', 5988)>
ResourceWarning: unclosed <socket.socket fd=6, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('127.0.0.1', 33414), raddr=('127.0.0.1', 5988)>

pywbemcli, version 0.9.0.dev1
pywbem, version 1.2.0.dev1

Note sure why and not sure if this is just aberation but wanted to document for now.

This happens every time now when using OpenPegasus.

This is generally happening in any connection from pywbemcli on a ubuntu 1.04 python 3.6.8 client to either an openpegasus server on the same system or a local openpegasus container.

1. I happens only when using interactive mode.
2. It happens when communicating either with the local openpegasus or with openpegasus in a container.
3. It happens when you close the connection.
4. It does NOT happen when your are not in interactive mode
5. I saw it for the first time the day before this issues was filed.
6. It does not happen when pywbem_mock is the server

An example both interactive and cmd line modes
``` 

 ppp -n ophttp
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli or <CTRL-r> to search history, 
pywbemcli> class get blah
Error: CIMError: 6 (CIM_ERR_NOT_FOUND): CIM_ERR_NOT_FOUND: blah
pywbemcli>                                                                      
Exception ignored in: <socket.socket fd=6, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('127.0.0.1', 36892), raddr=('127.0.0.1', 5988)>
ResourceWarning: unclosed <socket.socket fd=6, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('127.0.0.1', 36892), raddr=('127.0.0.1', 5988)>

(pycli3) kschopmeyer@leonard:~/pywbem/pywbemtools(ks/#588-support-wbemsvrstats)$ ppp -n ophttp class get blah
Error: CIMError: 6 (CIM_ERR_NOT_FOUND): CIM_ERR_NOT_FOUND: blah
(pycli3) kschopmeyer@leonard:~/pywbem/pywbemtools(ks/#588-support-wbemsvrstats)$

[...]

PR #899, opened 2021-02-10
Added end2end testcase for ResourceWarning about unclosed sockets

andy-maier:
This PR is now ready to be reviewed.

**Old description:**

This is to investigate and reproduce issue #883, without having it fixed. Note that PR #900 which now got merged only prepares the fix, but does not fix it yet.

The test can be run locally with `make end2endtest` as long as the target WBEM server is in a docker container. The docker image name is by default the pegasus image name, and can be set with the env var TEST_SERVER_IMAGE. See `make help`.

**DISCUSSION: (resolved)**
* Why does the new testcase not reproduce the warning in GitHub Actions? Answer: Did not use `--warn`. It does reproduce them with `--warn`.
* Does the new testcase succeed in the environments that produced the original ResourceWarning? Answer: After using `--warn`, the `make end2endtest` reproduced it on macos. See issue #883 for details.

### DIFF
--- a/tests/end2endtest/test_errors.py
+++ b/tests/end2endtest/test_errors.py
@@ -1,0 +1,69 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


KSchopmeyer:
Karl run test with this one with existing container pegasus and without the other prs that fix the close.  ACTION:

KSchopmeyer:
I ran tests with python 3.6 and 3.9 using the current master branch.  I could not create the issue with python 3.9 but consistently created it with python 3.6.  Uploaded examples and pip lists to the corresponding issue.

PR #900, opened 2021-02-11
Close WBEMConnection when leaving interactive mode

andy-maier:
The first commit just performs some restructuring on connection related elements in the `PywbemServer` and `ContextObj` classes.

There are two temporary flags that are used to switch between old and new behavior.
```
PR900_REQUIRE_CORRECT_CONNECTION_STATUS = False
PR900_DISCONNECT_CALLS_CLOSE = False
```

TODOs for Karl when continuing the work on this PR:
* Enable the two temporary flags to enable the new behavior. That causes some tests to now fail.
* Rework how and when `PywbemServer.connect()` and `PywbemServer.disconnect()` and `PywbemServer` objects in general are used. One major verification is is when the tests succeed again.
* Extend testcases to cover all variartions on how connection related input can be specified, inherited, overwritten in command mode and interactive mode, including setting default connections and selecting interactive connections.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,7 @@ endif
 pytest_end2end_opts := -v --tb=short $(pytest_opts)
 
 ifeq ($(python_m_version),3)
-  pytest_warning_opts := -W default -W ignore::PendingDeprecationWarning -W ignore::ResourceWarning
+  pytest_warning_opts := -W default -W ignore::PendingDeprecationWarning
   pytest_end2end_warning_opts := $(pytest_warning_opts)
 else
   pytest_warning_opts := -W default -W ignore::PendingDeprecationWarning
--- a/pywbemtools/pywbemcli/_context_obj.py
+++ b/pywbemtools/pywbemcli/_context_obj.py
@@ -46,7 +46,7 @@ class ContextObj(object):  # pylint: disable=useless-object-inheritance
                  pull_max_cnt, timestats, log, verbose, pdb,
                  warn, connections_repo):
 
-        self._pywbem_server = pywbem_server
+        self.pywbem_server = pywbem_server   # has setter method
         self._output_format = output_format
         self._use_pull = use_pull
         self._pull_max_cnt = pull_max_cnt
@@ -61,7 +61,6 @@ class ContextObj(object):  # pylint: disable=useless-object-inheritance
 
         self._spinner_enabled = None  # Deferred init in getter
         self._spinner_obj = click_spinner.Spinner()
-        self._conn = None
 
     def __repr__(self):
         return 'ContextObj(at {:08x}, pywbem_server={!r}, outputformat={}, ' \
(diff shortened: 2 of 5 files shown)


KSchopmeyer:
**COMMENTS:**

1. How can we close the connection in command mode (and why does the warning not show up there). In cmd mode, 

This is all tied up in that set of code : that is the last thing the cli function does pywbmcli.py about line 791.   

```

    if ctx.invoked_subcommand is None:
        print("Debug: invoke_repl")
        ctx.invoke(repl)
        print("Debug: repl_return")
        conn = ctx.obj.conn
        if conn:
            print("Debug: Closing WBEMConnection")
            conn.close()  # close() was added in pywbem 1.2.0

Note that there no else which is the path that the cmd mode takes.  However, the repl mode also takes that path for each interactive command entered.  We could add an else and a flag indicating that this was not the first command to allow the interactive mode to execute or if it was the first use of that else, know that it was a single command.  There are multiple uses for a place in pywbemcli.py:cli to know that it is at a point in the code where it just finished command mode, not just this one.

I just added:

```
    else:
        print("Debug: invoked_subcommand %s" % ctx.invoked_subcommand)

and you can see the flow.

2. Do we need a way to reopen a wbem connection???
I don't know about pywbem itself wrt reopen but in pywbemcli, we probably don't   we close a connection at two times:

a. When we exit we close the existing condition.  You covered the repl and I think the whole else  logic could cover the cmd mode Maybe that test is a global repl_mode flag that gets set when ctxinvoke(repl) is executed.

b. When we execute connection select we possibly discard an existing connection and we need to close it as part of that process since we are not caching the PywbemServer objects at this point.

Since we do not cache the PywbemServer objects but create them new each time a new connection is selected, we don't appear to have a need for reconnection, Just create a new connection.  

Finally I am not even sure what the difference between the original connect ad a reopen connect would be. This could easily get messy because without closing, you are not sure what the connection state really is  after some time not communicationg with the connection with keep-alives, server timeouts, etc.


**QUESTION:** Would it be more logical to move the actual close into PywbemServer class since that might help cover the case of 'connection select' dropping one connection for another also and clearly tie the close to the connection itself.

PS: This is the bit of code I scratched my head over for a while which is why I added the comments in the statistics pr about what the existence of the obj in pywbemcli.py cli really means. I.e. it does not cleanly separate cmd from interactive mode.  A repl_mode global flag would do that.

andy-maier:
* On 1: I understand the flow in command mode. The problem is that we want to close *after* the command has been executed. Your print statement in the else clause is executed *before* the command is executed.
  I experimented with overloading invoke() in the PywbemcliCommand class (you still can see it in the PR's code) and with the [Click resultcallback() decorator](https://click.palletsprojects.com/en/7.x/api/?highlight=callback#click.MultiCommand.resultcallback), but both had issues. The invoke() approach had the issue that it successfully closed the connection after command execution, but it is still closed when the next command in interactive mode is executed. That's why I was thinking about adding an open() method but that felt awkward because we need to save all the input parameters either in the WBEMConnection or in the pywbemcli context. The callback approach did not work right, I forgot the details.
  
What we actually need is a way to get control not on each command, but at the very end just before the pywbemcli command leaves. We don't have a place where we can catch SystemExit, otherwise that would be one way. I was even thinking about registering an atexit function, but in that function we do not have the connection so we would need to get it in by using closures (which is not impossible).

I had not thought about connection selection, and having the close in the PywbemServer class seems to be a good idea, but that alone does not close it at the very end.

KSchopmeyer:
ACTION: Karl see why the issue.

andy-maier:
I don't see how the PywbemServer class could be used to close the connection (other than passing a close() through, but then who invokes the close()).

**DISCUSSION:**
* Do we need to close the connections in interactive mode?

andy-maier:
We need to review the whole approach for creating and closing connections.

andy-maier:
We agreed that I would update this PR to do some basic changes as dicussed, most notably to change the methods of PywbemServer to have a connect/disconnect notion and to have disconnect() call conn.close(). That has been done as of now. The tests currently fail because of the call to conn.close().

Karl would then take over this PR to straighten out the use of PywbemServer and its connect/disconnect calls.

andy-maier:
The latest commit introduced two temporary flags into the pywbem_server.py module:
```
# TODO(PR900): Activate the PR900 flags when continuing the work on PR #900;
#              Remove the PR900 flags when done.
PR900_REQUIRE_CORRECT_CONNECTION_STATUS = False
PR900_DISCONNECT_CALLS_CLOSE = False
```
They should be set to True when continuing the work on this PR. That will cause tests to error out because the use of connect()/disconnect() and managing the PywbemServer objects need to be adjusted as part of continuing the work.

andy-maier:
Karl and I agreed on the Slack channel to merge this PR.

PR #906, opened 2021-02-20
Test: Added test module for _utils.py

andy-maier:
See commit message.


andy-maier:
Marked as priority because it brings coverage above the limit of 90% again and thus will allow other PRs to succeed w.r.t. coverage.

KSchopmeyer (commented):
One comment here and approve.  Shouldn't we have note in docs/changes.rst for this even though it is in tests.

PR #907, opened 2021-02-21
Aligned client and server statistics tables; improved their docs

andy-maier:
See commit message.


pywbemtools/pywbemcli/_cmd_statistics.py
```diff
@@ -560,32 +560,31 @@ def avg_time(cimtime, num_ops):
...
                                           number_of_operations)
 
-            rows.append([number_of_operations,
-                         "{:.1f}".format(cimom_avg_elapsesd_time),
-                         "{:.1f}".format(provider_avg_elapsesd_time),
-                         "{:.1f}".format(avg_request_size),
-                         "{:.1f}".format(avg_response_size),
-                         op_type_name])
+            rows.append([op_type_name,
+                         number_of_operations,
```
KSchopmeyer, line 574: I originally put the operation as the last column for a reason, It is usually the widest column and it seemed to make reading easier if if did not take up a lot of the left side of the table.  That was maybe my preference and the way I read tables but since we are putting these things in limited width terminal envrionment it seemed logical.
**DISCUSSION**

pywbemtools/pywbemcli/_cmd_statistics.py
```diff
@@ -560,32 +560,31 @@ def avg_time(cimtime, num_ops):
...
-                   'Operation Name']
+        headers = ['Operation', 'Count', 'Server Time\n[ms]',
+                   'Provider Time\n[ms]', 'Request Size\n[B]',
+                   'Response Size\n[B]']
 
         context.spinner_stop()
         click.echo(format_table(
             rows, headers,
-            title='WBEM server CIM Operation Statistical Data',
+            title='Server statistics',
```
KSchopmeyer, line 587: Should we expand this more to be sure the reader understands the table from the results (i.e. server maintained statistics) or something similar?? In this case it is not an issue of realestate on the line since the table itself is much longer than the title.

pywbemtools/pywbemcli/_cmd_statistics.py
```diff
@@ -560,32 +560,31 @@ def avg_time(cimtime, num_ops):
...
+                         "{:.3f}".format(cimom_avg_elapsesd_time),
+                         "{:.3f}".format(provider_avg_elapsesd_time),
+                         "{:.0f}".format(avg_request_size),
+                         "{:.0f}".format(avg_response_size)])
 
-        headers = ['Operation\nCount', 'Server\ntime(usec)',
-                   'Provider\nTime(usec)', 'Request\nsize(bytes)',
-                   'Response\nsize(bytes)',
-                   'Operation Name']
+        headers = ['Operation', 'Count', 'Server Time\n[ms]',
```
KSchopmeyer, line 580: Would it be logical to pack up the headers more and into multiple lines so that the column header widths do not greatly exceed the expected column data widths making the table fit into a more narrow terminal width.  Ex. 'Server\nTime [ms]`, 'Request\nsize[b]`

docs/pywbemcli/commands.rst
```diff
@@ -1556,115 +1556,67 @@ command.
...
+    a. Each CIM-XML response from the WBEM server may include an extra header
+       field ``WBEMServerResponseTime`` with the server response time for that
+       operation. Pywbemcli puts those server response times into the client
+       statistics it maintains. The inclusion of the server response time
+       can be enabled and disabled with the ``statistics server-on`` and
+       ``statistics server-off`` commands.
+
+    b. Statistical information on operation execution in the WBEM server and
+       its providers may be gathered and maintained and by the WBEM server.
+       This server statistics is completely independent of the client statistics
```
KSchopmeyer, line 1592: Probably should be These server statistics are

docs/pywbemcli/commands.rst
```diff
@@ -1556,115 +1556,67 @@ command.
...
+       all clients working with that server.
+       The gathering of server statistics can be enabled and disabled with the
+       ``statistics server-on`` and ``statistics server-off`` commands.
+       The server statistics can be retrieved and displayed with the
+       ``statistics server-show`` command.
+
+    The mock environment implemented by pywbemcli does not support server
+    statistics.
+
+    The capabilities for managing and retrieving server statistics may be
```
KSchopmeyer, line 1603: Change may be to 'is'

docs/pywbemcli/commands.rst
```diff
@@ -1556,115 +1556,67 @@ command.
...
+    supported only in some WBEM server implementations. While these capabilities
+    were documented in the DMTF CIM Schema, they were never included as part of
+    a DMTF or SNIA management profile, so the implementations may vary across
+    WBEM server implementations. Pywbemcli makes a best effort to manage the
```
KSchopmeyer, line 1607: I like this sentence.  Best effort works well

docs/pywbemcli/commands.rst
```diff
@@ -1748,26 +1700,86 @@ the XML response from the WBEM server
 The ``statistics show`` command displays the current pywbemcli managed
 statistics as a table.
 
-The following example shows the use of ``--timestats`` option in
-the interactive mode to display statistics after the enumerate command
-is executed.
+Using this command only makes sense in interactive mode. In interactive mode,
+the pywbemcli managed statistics is maintained for the entire interactive
+session, and executing commands in the interactive session causes the
```
KSchopmeyer, line 1705: how about "executing commands that communicate with the server"

pywbemtools/pywbemcli/_context_obj.py
```diff
@@ -288,88 +288,41 @@ def execute_cmd(self, cmd):
             if not self.pdb:
                 self.spinner_stop()
 
-            # Issue statistics if required. Note: uses _conn in order
-            # not to create the connection if not created.
-            if self.timestats and self._conn:
+            # Issue statistics if required.
+            if self.timestats:
```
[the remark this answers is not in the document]
KSchopmeyer: My original idea was to never show statistics after commands that did not communicate with the server but that meant a lot of minor changes to understand which commands communicate with the server and seemed to be a lot of work for minor gain but not displaying stats before we have communicated with a server makes sense.

pywbemtools/pywbemcli/_context_obj.py
```diff
@@ -288,88 +288,41 @@ def execute_cmd(self, cmd):
...
-
-        # build list of column names
-        header = ['Op\nCnt', 'Exc\nCnt', 'Op Time(S)\nAvg/Min/Max']
-        if include_svr_time:
-            header.append("Server Time(S)\nAvg/Min/Max")
-        if include_lengths:
-            header.extend(['RequestLen\nAvg/Min/Max', 'ReplyLen\nAvg/Min/Max'])
-
-        # build table rows from snapshot of OperationStatistics
+        header = ['Operation', 'Count', 'Errors',
```
KSchopmeyer, line 308: should we compress these headers in to columns that more closely match the data wiidths and use multiple lines for the headers??? Also same question about the operation name to the lefl vs the right column,.

QUESTION: Should we consider sorting thee rows here? We could sort by names, number of operations, etc.

KSchopmeyer (commented):
This time I made a number of comments but not so many to the documentation.  Many of them are questions or maybe items for discussion

pywbemtools/pywbemcli/_context_obj.py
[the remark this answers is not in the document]
andy-maier: I understand the goal, yet the test for `self._conn` does not reach that goal. Where it falls short is in most of the "server" commands. For example, "server brand" does communicate with the server, but for reasons I have not investigated, the `self._conn` attribute is not set after it has been executed.
**DISCUSSION**

docs/pywbemcli/commands.rst
```diff
@@ -1748,26 +1700,86 @@ the XML response from the WBEM server
...
+
+    pywbemcli> server brand
+    OpenPegasus
+
+    pywbemcli> statistics show
+    Client statistics
+    Operation                 Count    Errors    Client Time    Server Time    Request Size    Response Size
+                                                        [ms]           [ms]             [B]              [B]
+    ----------------------  -------  --------  -------------  -------------  --------------  ---------------
+    EnumerateInstanceNames        2         1         33.174          1.407             333             6225
```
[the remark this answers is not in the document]
andy-maier: There are multiple goals here: You are suggesting to round to the same relative precision, but there is also a need for maintaining the same absolute precision. If people are using this for calculations, they will add large and small times, and then they should have the same absolute precision. Or when people calculate the time spent on the network by subtracting server time from client time, these values may be differently large too, so maintaining the same absolute precision would be helpful. Finally, the measurement error is an absolute error, i.e. a large time is really also precise to the same digit as a small number.

Bottom line, I think we should stick with rounding to an absolute precision and not make it relative.
**DISCUSSION**

docs/pywbemcli/commands.rst
```diff
@@ -1748,26 +1700,86 @@ the XML response from the WBEM server
...
+close to the network interface, the time includes only a very small fraction of
+the time spent in code of the pywbemcli client program.
+
+The **Server Time** column shows the total elapsed time the operation took from
+a perspective of the WBEM server. It has the same meaning as the Server Time
+value shown in the server statistics. This time includes time spent in the
+CIM object manager code and time spent in its providers. The Server Time is
+obtained from the header field ``WBEMServerResponseTime`` in the CIM-XML
+response message; if the WBEM server does not support returning this field or
+has it disabled, the Server Time is shown as 0. That is why the Server Time
```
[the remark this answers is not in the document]
andy-maier: The 0 is not created by pywbemcli: In case of the client statistics, it is created in pywbem. In case of the server statistics, it is created in the WBEM server (e.g. for class operations, Provider Time is 0).

**DISCUSSION**: Do we really want to change this?

docs/pywbemcli/commands.rst
```diff
@@ -1781,36 +1793,56 @@ The ``statistics server-show`` command displays the current WBEM server gathered
...
-    |           5 |       7372.6 |          0   |         595.2 |        3192   | OpenEnumerateInstances |
-    +-------------+--------------+--------------+---------------+---------------+------------------------+
+    $ pywbemcli -s http://localhost statistics server-show
+
+    Server statistics
+    Operation                 Count    Server Time    Provider Time    Request Size    Response Size
+                                              [ms]             [ms]             [B]              [B]
+    ----------------------  -------  -------------  ---------------  --------------  ---------------
+    GetClass                    315          0.239            0                 399             4124
+    GetInstance                   4          0.495            0.07              862             3128
```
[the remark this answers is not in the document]
andy-maier: Showing them blows up the width quite a bit.
I have never used them.
**DISCUSSION**

docs/pywbemcli/commands.rst
```diff
@@ -1748,26 +1700,86 @@ the XML response from the WBEM server
...
+such errors is not necessarily a problem, depending on the logic in the
+pywbemcli client program.
+
+All time and size values in this report are average values across the number
+of operations executed, rounded to the precision shown.
+
+The **Client Time** column shows the total elapsed time the operation took from
+a perspective of the pywbemcli client program. This time includes network time
+and server time; since the measurement point in the client program is very
+close to the network interface, the time includes only a very small fraction of
```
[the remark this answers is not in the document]
andy-maier: You are right, it is not so simple. In pywbem, the time is measured as close to the pywbem API as possible, so the entire conversion of the request to CIM-XML and parsing of the response CIM-XML is done during the client time. The entire pywbemcli code however is not measured.

I have changed this to: 
```
This time includes network time
and server time and most of the time spent in the pywbemcli command. More
specifically, the client time is measured by the statistics support of the
pywbem library directly after the API for executing an operation, so it does
include the creation of the CIM-XML for the request and the parsing of the
CIM-XML for the response, but it does not include any processing in the
code of the pywbemcli command above the pywbem API.
```

docs/pywbemcli/commands.rst
```diff
@@ -1781,36 +1793,56 @@ The ``statistics server-show`` command displays the current WBEM server gathered
...
+in the provider from a perspective of the CIM object manager portion of the
+WBEM server that calls the provider. WBEM servers typically only report values
+for instance providers, so the values for class and qualifier operations are
+typically shown as 0. The provider time typically also includes the time spent
+in "up-calls" from the provider back to the CIM object manager portion of the
+WBEM server (and possibly down to other providers).
+
+The **Request Size** and **Response Size** columns show the size of the CIM-XML
+request and response messages, respectively. These values typically include
+only the size of the HTTP bodies but do not include the size of the HTTP header
```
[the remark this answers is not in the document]
andy-maier: Thanks for catching this. I have changed the text to:

```
For OpenPegasus and WBEM servers
that followed the description in CIM_CIMOMStatisticalData.mof, these values
include the size of the HTTP bodies and the size of the HTTP header fields.
```
DONE

docs/pywbemcli/commands.rst
```diff
@@ -1781,36 +1793,56 @@ The ``statistics server-show`` command displays the current WBEM server gathered
...
+
+The **Count** column shows the number of operations executed.
+
+All time and size values in this report are average values across the number
+of operations executed, rounded to the precision shown.
+
+The **Server Time** column shows the total elapsed time the operation took from
+a perspective of the WBEM server. It has the same meaning as the Server Time
+value shown in the client statistics. This time includes time spent in the
+CIM object manager code and time spent in any providers.
```
andy-maier, line 1828: After reading the description of CIMOMElapsedTime in CIM_CIMOMStatisticalData.mof, it turns out it says that it does not include the provider time.
Can you please double check what Pegasus implemented for it?
**DISCUSSION**

docs/pywbemcli/commands.rst
```diff
@@ -1781,36 +1793,56 @@ The ``statistics server-show`` command displays the current WBEM server gathered
...
+
+The **Provider Time** column shows the total elapsed time the operation spent
+in the provider from a perspective of the CIM object manager portion of the
+WBEM server that calls the provider. WBEM servers typically only report values
+for instance providers, so the values for class and qualifier operations are
+typically shown as 0. The provider time typically also includes the time spent
+in "up-calls" from the provider back to the CIM object manager portion of the
```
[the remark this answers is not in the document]
andy-maier: To fix that, and to improve the statement about instance providers, changed that text to:
```
... WBEM server that calls the provider. OpenPegasus only reports values for
instance providers and reports the values for class and qualifier operations
as 0. Other WBEM servers would typically also do that. The provider time
includes the time spent for performing any actions in the managed system
and also the time spent in any "up-calls" from the provider back to the CIM
object manager portion of the WBEM server (and possibly down to other providers).
```
DONE

docs/pywbemcli/commands.rst
```diff
@@ -1748,26 +1700,86 @@ the XML response from the WBEM server
...
+resulted in an error at the level of the CIM-XML protocol. The occurrence of
+such errors is not necessarily a problem, depending on the logic in the
+pywbemcli client program.
+
+All time and size values in this report are average values across the number
+of operations executed, rounded to the precision shown.
+
+The **Client Time** column shows the total elapsed time the operation took from
+a perspective of the pywbemcli client program. This time includes network time
+and server time; since the measurement point in the client program is very
```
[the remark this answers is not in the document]
andy-maier: Duplicate of comment on line 1768, resolved there.

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: If the display is wider than the terminal width, each line spills over, and I find it easier in that case to locate the right row when it starts with the operation name.

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: We should use one consistent term and then use it anywhere in the docs and also as the display title. So far, it seems we have not agreed to one single term, I see all kinds of related but different terms in the docs.

**DISCUSSION**: What should our single term be, for client statistics and server statistics?

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: There are multiple factors to consider: Readability and width.

With the current implementation, the client statistics is 104 chars wide and the server statistics is 96 chars wide. That is below our standard width of 120 characters.

If we optimize the column headers for with as you suggest, we get down to 91 chars for the client statistics and to 81 chars for the server statistics. That is still above the 80 characters I suppose you try reaching.

So I would go for the better readability.
**DISCUSSION**

docs/pywbemcli/commands.rst
andy-maier: DONE

docs/pywbemcli/commands.rst
andy-maier: DONE

docs/pywbemcli/commands.rst
andy-maier: :-)

docs/pywbemcli/commands.rst
andy-maier: DONE, added that as proposed.

pywbemtools/pywbemcli/_context_obj.py
andy-maier: On optimizing column headers: See [this comment](https://github.com/pywbem/pywbemtools/pull/907#discussion_r580009846)
On operation name left or right: See [this comment](https://github.com/pywbem/pywbemtools/pull/907#discussion_r580006287)

Now on sorting the rows:

The current code does the following:
* For the client statistics: Sorted by decreasing average client time (see `ContextObj.format_statistics()`)
* For the server statistics: Not sorted on client side, displays enumerated instances as received (see `cmd_statistics_server_show()`)

**DISCUSSION**: How to sort by default? Should we provide options for users to select sorting?

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
KSchopmeyer: I think this comes down to readability in the table vs use in computations.  i.e. Either side could be argued for.  It is not a major issue to me.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
KSchopmeyer: I agree with your option 1.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
KSchopmeyer: 1. I agree with issue 1.
2. Issue 2.  I actually do to see if there are any outliers or if the average actually means something. 
3. Issue 3. Yes that was never defined in the spec so never part of any implementations that followed the model.
4. In reality we could really have gone for a statistical measure like variance.
5. In some cases there are extreme variances (i.e. EnumerateInstances where the delivery time and respose size can vary by enormous values and the average sort of has no meaning.  i.e. you request 100 time an enumerate that returns one object and 1 time an enumerate that returns 100,000 objects.  Average probably means nothing.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
KSchopmeyer: OK

pywbemtools/pywbemcli/_cmd_statistics.py
KSchopmeyer: I am tending to agree that our goal should probably be 80 for very simple tables that make sense in 80 but 120 in general for tables that do not make sense for 80 columns.  You convinced me.  I was just too lazy to expand my standard terminal width to 120 columns.

geneally keeping these down to 120 columns would be logical to me.

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: Darn, I forgot that I did to that sort. Maybe we should sort the server stats the same way we sort the client stats. In particular, since we do not have any idea of the order received from the server it might make sense to make the order independent of the server order.

docs/pywbemcli/commands.rst
KSchopmeyer: Will check

pywbemtools/pywbemcli/_context_obj.py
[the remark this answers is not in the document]
andy-maier: We agreed that we would try to find what the proper way is to find out that a command is local-only, and not to show statistics when local-only.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
andy-maier: Karl agreed to leave it as it is now (as per this PR). Resolved.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
andy-maier: We agreed to not show min/max. Resolved.

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: We agreed that 120 is fine for now. Resolved.

pywbemtools/pywbemcli/_context_obj.py
andy-maier: We agreed that for now, we would sort the server statistics also by average server time.

andy-maier: The latest update implements that. DONE.

pywbemtools/pywbemcli/_context_obj.py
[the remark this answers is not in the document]
andy-maier: I removed the separate attribute `self._conn` and derived the `self.conn` property it from `self.wbem_server.conn`, and added a new `is_connected` property. This cleaned up the data housekeeping sufficiently to now be consistent across all commands. So the client statistics are now shown when requested and when `self.is_connected` is True. That worked correctly for all commands I tested with.
DONE.

docs/pywbemcli/commands.rst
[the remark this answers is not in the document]
andy-maier: Option 1 will need to be implemented in the pywbem statistics support. I have created issue https://github.com/pywbem/pywbem/issues/2603 for that. 
DONE here.

docs/pywbemcli/commands.rst
KSchopmeyer: **CONCLUSION:** Yes, OpenPegasus does treat the time in the CIM Object Manager coorectly by subtracting the provider time if there is any before saving the server time.  It is nod in Common/CIMMessage.cpp method endServer()

pywbemtools/pywbemcli/_cmd_statistics.py
KSchopmeyer: No issue. DONE

pywbemtools/pywbemcli/_cmd_statistics.py
KSchopmeyer: Since both of these entities have server statistics, that is not complete enough.  That it why I have tried to use server maintained/managed statistics vs pywbemcli maintained/managed statistics.
We can standardize to one of those temrs (ex. server maintained statistics vs pywbemcli maintained statistics)

If you have a better term I have no objection.

andy-maier: Ok, I understand your argument now. However, I'm not sure I like it.

The client statistics has a server time value (not server statistics). Does that warrant the use of a more complex term for it?

docs/pywbemcli/commands.rst
andy-maier: Ok, great! That means we can go by the description of CIMOMElapsedTime for all CIMOM brands, which the current code does. RESOLVED.

andy-maier:
I have created PR #914 for the one remaining discussion point about the terms we use. That allows to let this PR go in.

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: I have created PR #914 for the one remaining discussion point about the terms we use. RESOLVED.

PR #908, opened 2021-02-22
Removed pinning of Pylint to <2.5 and fixed issues

andy-maier:
See commit message.


KSchopmeyer (commented):
I don't have any issues with the changes themselves other than that the raise_pywbem... was used just to capture the exception in a single place and avoid possible issues where the raise was forgotten (But that is not really an issue). However, changes to satisfy does sort of bother me.

i.e. I don't have any issues with the code changes themselves. noting that at this point the code does not pass pylint.

The real question is whether the execution times of pylint 2.7.0 are going to be an issue with pywbemcli as we currently have that under discussion in a pywbem issue

Issue #885, opened 2021-01-29
Support deletion of non-empty namespaces


andy-maier:
The new `namespace delete` command errors out if the target namespace is not empty. It would be good to have a simple way to get rid of a namespace. One option would be to add a `--force` or `--including-content` option to `namespace delete` which would delete the objects in the namespace as well.

**DISCUSSION (RESOLVED):**
* When deleting instances in the namespace, is that done with DeleteInstance so that instance providers get control?
* Do instances need to be deleted in a particular order?
* Does the command immediately error out at the point an object cannot be deleted?

KSchopmeyer:
1. If delete namespace uses delete class, that has a — force option.
2. I can see wher we have to think out issues here. 1. Should associations be deleted first? 2. What about classes that support providers?  Class and no deletion are useful but clearly not production tools, ever. 

3. I think that there. Is a risk in some environments of errors but I would think that since this not a production use capability, we have to live with deleting what is possible and user lives with result.
4. Should we just disallow deleting interop namespace?

KSchopmeyer:
Discussion:

1. Use of --force option.  Should we make it fail or successful completion?  Should we use the work force or is there a better word for what we are trying.  Lets change the wording to --include-instances for remove class and --remove-objects for for the delete namespace.
2.

andy-maier:
Before implementing it, I try to summarize the design:
* We will continue to reject the deletion of an Interop namespace.
* The option for enabling the deletion of objects in the namespace in 'namespace delete' will be '--include-objects'.
* For consistency, we will add an option '--include-instances' to the 'class delete' command and deprecate its '--force' option'.
* When deleting objects in a 'namespace delete' command, the algorithm will be:
  * Determine all top-level classes via EnumerateClasses(classname=None, DeepInheritance=False, IncludeQualifiers=True).
  * Based on the Association qualifier, divide the top-level classes in associations, indications and ordinary classes, in order to delete them in the correct order.
  * For each top-level indication class, perform the internal equivalent of 'class delete --force'.
  * For each top-level association class, perform the internal equivalent of 'class delete --force'. We assume that cross-namespace associations will be handled correctly by not leaving a dangling half-instance in the other namespace(s).
  * For each top-level ordinary class, perform the internal equivalent of 'class delete --force'.

andy-maier:
[...]

PR #909, opened 2021-02-23
Changed --force / -f option of class delete command to --include-instances

andy-maier:
See commit message.
This is a first part of the resolution for PR #885, that can be moved ahead independently.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_class.py
+++ b/pywbemtools/pywbemcli/_cmd_class.py
@@ -47,7 +47,7 @@ from ._common_options import add_options, propertylist_option, \
     help_option
 from ._displaytree import display_class_tree
 from ._click_extensions import PywbemcliGroup, PywbemcliCommand
-
+from ._utils import pywbemcliwarn
 
 #
 #   Common option definitions for class group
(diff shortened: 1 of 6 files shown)


pywbemtools/pywbemcli/_cmd_class.py
```diff
@@ -186,8 +186,13 @@ def class_get(context, classname, **options):
                      options_metavar=CMD_OPTS_TXT)
 @click.argument('classname', type=str, metavar='CLASSNAME', required=True,)
 @click.option('-f', '--force', is_flag=True, default=False,
+              help=u'Same as --include-instances. The -f / --force option has '
+                   'been deprecated and will be removed in a future version.')
+@click.option('--include-instances', is_flag=True, default=False,
               help=u'Delete any instances of the class as well. '
-                   'Some servers may still reject the class deletion. '
+                   'WARNING: Deletion of instances will cause all implemented '
```
KSchopmeyer, line 193: I do not understand the all in "cause all implemented". Might be better with "Deletion of instances can cause side-effects such as ..."

KSchopmeyer (commented):
One comment and otherwise approved.

pywbemtools/pywbemcli/_cmd_class.py
andy-maier: I wanted to express that these effects are intentionally implemented in the server and even mandated in the mgmt profile if that was followed. I did not do a good job on that, though and should have gone much further.

That these "side-effects" happen is really the sole purpose of the CIM instrumentation. Calling these intended effects "side-effects" is really strange. This still comes from the notion that we are dealing with a database that "may have" some "side-effects". No, they are the main effects and totally intentional.

Feel free to propose a sentence that expresses that. And that ideally does not contain "side-effects" or "may". ;-)

andy-maier: I have changed the text to now say:

```
    WARNING: Deletion of instances will cause the removal of corresponding
    resources in the managed environment (i.e. in the real world).
```
**DISCUSSION**

KSchopmeyer: How about changing will to can.  Will means absolutely this will happen and "depending on the CIM Instrumentation".   I hesitate to say will be cause there may be whole namespaces that do not deal with real instrumentation and which do not cause any failures and others for which deleting instances is a total disaster to the managed environment.

andy-maier: Agreed that the text above is ok, but that more text should be added that clarifies that some classes do not even allow instance deletion.

andy-maier: Add the sentence:
```
Some instances may not be deletable.
```
Also, added the warning about deletion of instances to the heklp text of the "instance delete" command.

PR #913, opened 2021-02-27
Support for deleting non-empty namespaces (part 3)

andy-maier:
See commit message.


KSchopmeyer (approved):
Approved of the part 3 component.  I had some comments on the part 2 component

PR #915, opened 2021-02-28
Fixes #912: Refactor ContextObj and PywbemServer classes - Part1

### LINKED ISSUE #912: Refactor pywbemcli and contextobj classes
Refactor these classes to:
Part 1

1. Eliminate the use of some of the properties in ContextObj such as conn and wbemserver since they are already used in PywbemServer. Specifically this means putting wbem_server and conn in PywbemServer and removing them from ContextObj. Note that this means changing all references to these two attributes which will cause changes to almost all command action methods (i.e. those that use conn or wbemserver) as art of the server connection

2.PywbemServer objects have a connect/disconnect functionality (although the methods are named create_connection() and reset()), which we could use to properly set up and tear down connections. They way we use this functionality currently is not very straight forward, for example "connection select" creates a new context object and globally patches that in, or commands in interactive mode make a deep copy of PywbemServer and modify it, but do not disconnect properly.

Part2

Refactor the pywbemcli code that rebuilds the PywbemServer object in interactive mode. (This will be a separate pr)

KSchopmeyer:
1. Separate out the use of conn and wbem_server from their context.

2. Remove conn and wbem_server from ContextObj and define them in
PywbemServer.

3. Modify several tests that no longer worked.

4. Cleaned uo the code in the DIAGNOSTIC display of ctx and ctx.obj
as part of a diagnostics displayzzzmerged .

5. Merged in pywbemserver connect and disconnect

6. Extended select to disconnect any currently connected PywbemServer object.

Part 2 - fix disconnect

Part 3 - SeparatePR Refactor the code that modifies PywbemServer for interactive command that modifies PyebemServer object

1. Test all variables and do not modify if no changes.
2. Move entire functionality to separate function
3. . Review under what conditions we must reconnect.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_connection.py
+++ b/pywbemtools/pywbemcli/_cmd_connection.py
@@ -324,7 +324,10 @@ def is_default_connection(context, connection):
 def is_current_connection(context, connection):
     """Returns True if connection named name is the default connection"""
 
-    current_connection = context.pywbem_server or None
+    if context.pywbem_server_exists():
+        current_connection = context.pywbem_server
+    else:
+        current_connection = None
     if current_connection and current_connection.name == connection.name:
         return True
     return False
@@ -398,7 +401,8 @@ def get_current_connection_name(context):
     Return the name of the current connection or None if there is no
     current connection
     """
-    return context.pywbem_server.name if context.pywbem_server else None
+    return context.pywbem_server.name if context.pywbem_server_exists() \
+        else None
 
 
 def pick_connection(name, context, connections_repo):
@@ -465,6 +469,7 @@ def test_pull_operations(context, test_class):
         'NOT SUPPORTED, or the exception if any exception other than
         CIMError(CIM_ERR_NOT_SUPPORTED)
     """
+
     def cimcall(request, *pargs, **kwargs):
         try:
             request(*pargs, **kwargs)
@@ -476,8 +481,10 @@ def test_pull_operations(context, test_class):
 
             return er
 
+    conn = context.pywbem_server.conn
+
     # Find a valid instance to use in tests.
-    test_instances = context.conn.EnumerateInstanceNames(test_class)
+    test_instances = conn.EnumerateInstanceNames(test_class)
     if not test_instances:
         raise click.ClickException("No instances found for test class {}".
                                    format(test_class))
@@ -485,25 +492,25 @@ def test_pull_operations(context, test_class):
 
     # execute each command and append the results to rows as a tuple
     rows = []
-    result = cimcall(context.conn.OpenEnumerateInstances, test_class)
+    result = cimcall(conn.OpenEnumerateInstances, test_class)
     rows.append(('OpenEnumerateInstances', result))
 
-    result = cimcall(context.conn.OpenEnumerateInstancePaths, test_class)
+    result = cimcall(conn.OpenEnumerateInstancePaths, test_class)
     rows.append(('OpenEnumerateInstancePaths', result))
 
-    result = cimcall(context.conn.OpenAssociatorInstances, test_instance)
+    result = cimcall(conn.OpenAssociatorInstances, test_instance)
     rows.append(('OpenAssociatorInstances', result))
 
-    result = cimcall(context.conn.OpenAssociatorInstancePaths, test_instance)
+    result = cimcall(conn.OpenAssociatorInstancePaths, test_instance)
     rows.append(('OpenAssociatorInstancePaths', result))
 
-    result = cimcall(context.conn.OpenReferenceInstances, test_instance)
+    result = cimcall(conn.OpenReferenceInstances, test_instance)
     rows.append(('OpenReferenceInstances', result))
 
-    result = cimcall(context.conn.OpenReferenceInstancePaths, test_instance)
+    result = cimcall(conn.OpenReferenceInstancePaths, test_instance)
     rows.append(('OpenReferenceInstancePaths', result))
 
-    result = cimcall(context.conn.OpenQueryInstances, "DMTF:CQL",
+    result = cimcall(conn.OpenQueryInstances, "DMTF:CQL",
                      "SELECT * FROM CIM_ManagedElement")
     rows.append(('OpenQueryInstances', result))
 
@@ -600,24 +607,24 @@ def cmd_connection_test(context, options):
     operation. Even if class operations are not supported, a return from the
     server such as "unsupported" indicates the server exists.
     """
-
+    conn = context.pywbem_server.conn
     output_format = validate_output_format(context.output_format, ['TABLE',
                                                                    'TEXT'])
 
     try:
-        classnames = context.conn.EnumerateClassNames()
+        classnames = conn.EnumerateClassNames()
         context.spinner_stop()
     except CIMError as ce:
         # class request failed, try instance request
         try:
             test_classname = 'CIM_ManagedElement'
             if ce.status_code == 'CIM_ERR_NOT_SUPPORTED':
-                context.conn.EnumerateInstances(test_classname)
+                conn.EnumerateInstances(test_classname)
         except Error as er:
             raise click.ClickException('Cannot interact with WBEM Server '
                                        '{}. EnumerateInstanceNames exception '
                                        '{} and EnumerateInstances {} exception '
-                                       ' {} '.format(context.conn.url,
+                                       ' {} '.format(conn.url,
                                                      ce.status_code,
                                                      test_classname,
                                                      er))
@@ -631,7 +638,7 @@ def cmd_connection_test(context, options):
             test_class = classnames[0]
     else:
         raise click.ClickException('No classes found to test in namespace {}.'.
-                                   format(context.conn.defaultnamespace))
+                                   format(conn.defaultnamespace))
 
     if options['test_pull']:
         result = test_pull_operations(context, test_class)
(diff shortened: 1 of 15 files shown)


pywbemtools/pywbemcli/_context_obj.py
```diff
@@ -276,6 +239,24 @@ def verbose(self):
...
+        Returns:
+            :class": `boolean True if an instance of PywbemServer is defined.
+        """
+        return self._pywbem_server is not None
+
+    def set_connection(self, connection):
+        """ Set the connection parameter as the current connection object and
+            establish the new connection
+        """
+        self._pywbem_server = connection
```
andy-maier, line 258: There is a pywbem server setter method. I don't think there should be an additional `set_connection()` method that does the same thing.

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -164,23 +164,24 @@ def __init__(self, server=None, default_namespace=DEFAULT_NAMESPACE,
 
         # WBEMServer object when connected; None when disconnected.
         self._wbem_server = None
+        self._conn = None
```
andy-maier, line 167: `self._ conn` is redundant to `self._wbem_server.conn`.

I cannot see a reason to have this redundant attribute and suggest to not introduce it again (after PR #900 just eliminated it).

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -418,16 +419,35 @@ def conn(self):
...
+        password is saved in the context, so the password is prompted for only
+        once (e.g. in interactive mode).
+        """
+        ctx = click.get_current_context()
+        if self._wbem_server is None:
+            self.get_password(ctx.obj)
+            self.connect(
+                log=ctx.obj.log,
+                use_pull=ctx.obj.use_pull,
+                verbose=ctx.obj.verbose)
```
andy-maier, line 450: The corresponding disconnect()  is missing, so we are not closing properly.

KSchopmeyer: This one confuses me.  We do have a disconnect method in PywbemServer (line 518).  It is called from two places

1. pywbemcli line718 - This is the normal exit.
2. _cmd_connection line 694 - disconnect any existing connection when a select is executed.
3. The code to execute it when the user enters an interactive command that changes the server is not in yet.

We may be missing some exit method but I thought we were complete here

**DISCUSSION**

pywbemtools/pywbemcli/_pywbem_server.py
KSchopmeyer: Agreed. Done. Conn property exists but just uses wbem_server as the flag

pywbemtools/pywbemcli/_context_obj.py
KSchopmeyer: The good news is that this is not even used. Removed.

pywbemtools/pywbemcli/_cmd_profile.py
```diff
@@ -186,7 +186,7 @@ def cmd_profile_list(context, options):
     """
     output_format = validate_output_format(context.output_format, 'TABLE')
 
-    server = context.wbem_server
+    server = context.pywbem_server.wbem_server
```
andy-maier, line 189: I suggest to clean up the name here as well for consistency and rename `server`to `wbem_server`.

pywbemtools/pywbemcli/_cmd_server.py
```diff
@@ -250,12 +251,13 @@ def cmd_server_info(context):
...
     """
+    wbem_server = context.pywbem_server.wbem_server
     output_format = validate_output_format(context.output_format, 'TABLE')
 
     try:
-        # execute the namespaces to force contact with server before
+        # Execute the namespaces to force contact with server before
         # turning off the spinner.
-        server = context.wbem_server
+        server = wbem_server
```
andy-maier, line 260: For consistency, I suggest to eliminate the additional `server` variable and go with `wbem_server` directly.

pywbemtools/pywbemcli/_pywbem_server.py
andy-maier: Ok, I did some more investigation on this. Issues are:
* The whole purpose of `disconnect()` is to call `conn.close()`. I had prepared the `PR900_DISCONNECT_CALLS_CLOSE` flag to enable that, but it is not set to `True` so `close()` is not called. It needs to be set to True.
TODO(Andy): More to come

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/_cmd_profile.py
KSchopmeyer: I agree.  Cleaned up.

pywbemtools/pywbemcli/_cmd_server.py
KSchopmeyer: Done

tests/unit/test_general_options.py
```diff
@@ -229,8 +229,7 @@ class       Command group for CIM classes.
      {'general': ['-s', 'httpx://blah'],
       'cmdgrp': 'class',
       'args': ['get', 'blah']},
-     {'stderr': ['Error: Invalid scheme on server argument. httpx://blah Use '
-                 '"http" or "https"'],
+     {'stderr': ["Error: Unsupported scheme 'httpx' in URL 'httpx://blah"],
```
andy-maier, line 232: Is this change done in the code?

tests/unit/test_connection_cmds.py
```diff
@@ -701,7 +701,8 @@ def GET_TEST_PATH_STR(filename):  # pylint: disable=invalid-name
     ['Verify connection command export no current conection. SEQ 1.7',
      {'general': [],
       'args': ['export']},
-     {'stderr': ['No server currently defined as current'],
+     {'stderr': ['No server specified for command "export" that requires '
+                 'a WBEM server'],
```
andy-maier, line 705: Is this change done in the code?

tests/unit/test_general_options.py
KSchopmeyer: There was a change in the code somewhere that causes this.  The message is from pywbem.  Something got dropped somewhere because the pywbemcli _validate_url is no longer being called and the exception from pywbem is what you see.  Somewhere in the move to connect we dropped the PywbemServer. server setter and that is where the call to the _validate_url is.  I will sort that out this afternoon

tests/unit/test_connection_cmds.py
KSchopmeyer: Yes in _context_obj.py pywbem_server property.  Modified slightly since this should be the current server rather than specified.

KSchopmeyer: DONE

KSchopmeyer:
Part 1 ready for re-review.  The current code (part 1) includes:

1. Eliminating conn and wbem_server from ContextObj
2. Refactor of Pywbemcli.cli to move the create_server_instance method from inner to outside cli
3. Cleanup of a number of issues because of that
4. Update to merge the disconnect() pr into this pr

Part2 will fix issue with disconnect

Part 3 will refactor the code that updates PywbemServer object to separate function.

pywbemtools/pywbemcli/_pywbem_server.py
andy-maier: Resolved in the other PR.
DONE

tests/unit/test_connection_cmds.py
```diff
@@ -973,7 +974,7 @@ def GET_TEST_PATH_STR(filename):  # pylint: disable=invalid-name
       'rc': 1,
       'test': 'innows',
       'file': {'before': 'none', 'after': 'none'}},
-     None, RUN],
+     None, OK],
```
andy-maier, line 977: In this PR you changed it to OK, in the part 2 PR you changed it back to RUN. I suggest to remove this change here.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -679,6 +686,12 @@ def cmd_connection_select(context, name, options):
     # Update the root context making this context the basis for future
     # commands in the current interactive session
     ContextObj.update_root_click_context(new_ctx)
+
+    # close any existing connection.
+    # TODO/KS: Validate that the first test is required.
```
andy-maier, line 691: Given the check in ContextObj.is_connected (both the property in this PR and the method in the second PR), it is not needed.

pywbemtools/pywbemcli/_cmd_statistics.py
```diff
@@ -224,7 +224,10 @@ def test_conn_exists(context):
     Exception:
        click.Exception if no current server.
     """
-    if not context.conn:
+    # TODO/KS: Is this necessary since we should be protecting with classes
+    #          or does the pywbem_server access provide same protection??
```
andy-maier, line 228: Accessing ContextObj.pywbem_server already checks for a server, so this additional check is not needed.

andy-maier (changes requested):
See comments.

In addition, I suggest that the two commits from this PR are squashed into one.

tests/unit/test_connection_cmds.py
KSchopmeyer: Marked OK here and now OK in next pr

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: Modified in part 2 so I did not update this..

pywbemtools/pywbemcli/_cmd_statistics.py
KSchopmeyer: Removed in the Part 2 PR

pywbemtools/pywbemcli/_cmd_connection.py
andy-maier: ok

pywbemtools/pywbemcli/_cmd_statistics.py
andy-maier: ok

andy-maier (approved):

andy-maier:
Rebased and squashed into one commit.

PR #921, opened 2021-03-05
Fixes #912: Refactor pywbemcli to issue disconnect part2

### LINKED ISSUE #912: Refactor pywbemcli and contextobj classes

KSchopmeyer:
See commit message

NOTE: This PR also removes the temporary PR900 flagged code/comments and actually does the close, etc. and corrects the
issue where the disconnect was not issued under all circmstances of the combination of general options and use of stdin

### DIFF
--- a/pywbemtools/pywbemcli/_pywbem_server.py
+++ b/pywbemtools/pywbemcli/_pywbem_server.py
@@ -33,11 +33,6 @@ from .config import DEFAULT_URL_SCHEME, DEFAULT_CONNECTION_TIMEOUT, \
     DEFAULT_NAMESPACE, MAX_TIMEOUT
 from ._pywbemcli_operations import PYWBEMCLIConnection, PYWBEMCLIFakedConnection
 
-# TODO(PR900): Activate the PR900 flags when continuing the work on PR #900;
-#              Remove the PR900 flags when done.
-PR900_REQUIRE_CORRECT_CONNECTION_STATUS = False
-PR900_DISCONNECT_CALLS_CLOSE = False
-
 WBEM_SERVER_OBJ = None
 
 PYWBEMCLI_LOG = 'pywbemcli.log'
@@ -443,6 +438,14 @@ class PywbemServer(object):
                 verbose=ctx.obj.verbose)
         return self._wbem_server
 
+    @property
+    def connected(self):
+        """
+        Current connected state. Returns True if connected and
+        False if not connected.
+        """
+        return bool(self._wbem_server)
+
     def password_prompt(self, ctx):
         """
         Request password from console.
@@ -519,20 +522,18 @@ class PywbemServer(object):
 
         Must be connected to the server when calling this method.
         """
-        # print("Debug: PywbemServer disconnect")
+        assert self._wbem_server is not None  # Must be connected
 
-        # TODO(PR900): Remove the PR900 flag test and else path once done.
-        if PR900_REQUIRE_CORRECT_CONNECTION_STATUS:
-            assert self._wbem_server is not None  # Must be connected
-        else:
-            # Old code to be removed once done
-            if self._wbem_server is None:
-                return
+        ctx = click.get_current_context()
+        if ctx.obj:
+            if ctx.obj.verbose:
+                if self._mock_server:
+                    server_txt = "mock environment {}".format(self._mock_server)
+                else:
+                    server_txt = "WBEM server {}".format(self._server)
+                click.echo("Disconnecting from {}".format(server_txt))
 
-        # TODO(PR900): Remove the PR900 flag test and else path once done.
-        if PR900_DISCONNECT_CALLS_CLOSE:
-            self._wbem_server.conn.close()
-
+        self._wbem_server.conn.close()
         self._wbem_server = None
 
     def connect(self, log=None, use_pull=None, verbose=None):
@@ -551,15 +552,17 @@ class PywbemServer(object):
           ValueError: Other issues with server-related attributes of this
             object that are considered programming errors.
         """
-        # print("Debug: PywbemServer connect")
 
-        # TODO(PR900): Remove the PR900 flag test and else path once done.
-        if PR900_REQUIRE_CORRECT_CONNECTION_STATUS:
-            assert self._wbem_server is None  # Must be disconnected
-        else:
-            # Old code to be removed once done
-            if self._wbem_server is not None:
-                return
+        assert self._wbem_server is None  # Must be disconnected
+
+        ctx = click.get_current_context()
+        if ctx.obj:
+            if ctx.obj.verbose:
+                if self._mock_server:
+                    server_txt = "mock environment {}".format(self._mock_server)
+                else:
+                    server_txt = "WBEM server {}".format(self._server)
+                click.echo("Connecting to {}".format(server_txt))
 
         if self._mock_server:
             conn = PYWBEMCLIFakedConnection(
--- a/tests/unit/test_general_options.py
+++ b/tests/unit/test_general_options.py
@@ -33,6 +33,9 @@ from .cli_test_extensions import CLITestsBase, PYWBEM_0, PYWBEM_1
 from .common_options_help_lines import CMD_OPTION_HELP_HELP_LINE
 
 TEST_DIR = os.path.dirname(__file__)
+CONNECTION_REPO_TEST_FILE_PATH = os.path.join(TEST_DIR,
+                                              'tst_general_options.yaml')
+CONN_REPO_FP_STR = str(CONNECTION_REPO_TEST_FILE_PATH)
 
 
 def GET_TEST_PATH_STR(filename):  # pylint: disable=invalid-name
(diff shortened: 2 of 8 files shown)


tests/unit/test_connection_cmds.py
```diff
@@ -1227,7 +1222,7 @@ class TestSubcmdClass(CLITestsBase):
         "desc, inputs, exp_response, mock, condition",
         TEST_CASES)
     def test_connection(self, desc, inputs, exp_response, mock, condition,
-                        default_connections_file_path):
+                        default_connections_file_path, verbose=True):
```
andy-maier, line 1225: The `verbose` argument causes this Pylint issue:
```
************* Module tests.unit.test_connection_cmds
tests/unit/test_connection_cmds.py:1225:55: W0613: Unused argument 'verbose' (unused-argument)
```

andy-maier:
When testing it, I found only one issue, when the connection is specified with -n, -s, or -m in interactive mode:

```
$ echo "-n pegasus server brand" | pywbemcli -v
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
Current connection: "None"
Error: No  current server for command "server brand" that requires a WBEM server. Specify a server with the "--server", "--mock-server", or"--name" general option, by setting the corresponding environment variables, or in interactive mode use "connection select" to define a target server
```

The same command in command mode works:
```
$ pywbemcli -v -n pegasus server brand
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
-Debug: PywbemServer connect to https://localhost:5989
OpenPegasus
Debug: PywbemServer disconnect from https://localhost:5989
```

The same command in interactive mode when the connection is specified on the command line also works fine:
```
$ echo "server brand" | pywbemcli -v -n pegasus
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
-Debug: PywbemServer connect to https://localhost:5989
OpenPegasus
Debug: PywbemServer disconnect from https://localhost:5989
```

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -443,6 +443,14 @@ def wbem_server(self):
                 verbose=ctx.obj.verbose)
         return self._wbem_server
 
+    @property
+    def connected(self):
+        """
+        Current connected state. Returns True if connected and
+        False if not connected.
+        """
+        return self._wbem_server
```
andy-maier, line 452: This does not return a boolean but the PywbemServer object. Change to using `bool()` around it.

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -519,7 +527,11 @@ def disconnect(self):
 
         Must be connected to the server when calling this method.
         """
-        # print("Debug: PywbemServer disconnect")
+        # TODO(PR900): Remove when disconnect fully tested
+        ctx = click.get_current_context()
+        if ctx.obj:
+            if ctx.obj.verbose:
+                print("Debug: PywbemServer disconnect")
```
andy-maier, line 534: I wonder if we should leave the connect/disconnect messages in, in verbose mode.

andy-maier (changes requested):
See comments

andy-maier:
In addition, the test coverage goes down by 1.4%. This can be seen once the Pylint error is fixed, or in the coverage report of PR #899 : https://coveralls.io/builds/37676122

tests/unit/test_connection_cmds.py
KSchopmeyer: Whoops forgot to remove that one after a test.

pywbemtools/pywbemcli/_pywbem_server.py
KSchopmeyer: Actually I think we should, at least for this release.

pywbemtools/pywbemcli/_pywbem_server.py
andy-maier: DONE

tests/unit/test_connection_cmds.py
andy-maier: DONE

KSchopmeyer:
Note this corrects the issue Andy found where the 3 forms of  a simple request delivered different results and adds test specifically for that.

andy-maier:
Rebased the PR on the updated PR #915.

andy-maier:
It has gotten better, but is not yet good:

Ok:
```
$ pywbemcli -v -n pegasus server brand
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
-Debug: PywbemServer connect to https://localhost:5989
OpenPegasus
Debug: PywbemServer disconnect from https://localhost:5989
```

Ok:
```
$ echo "server brand" | pywbemcli -v -n pegasus
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
-Debug: PywbemServer connect to https://localhost:5989
OpenPegasus
Debug: PywbemServer disconnect from https://localhost:5989
```

Disconnect missing:
```
$ echo "-n pegasus server brand" | pywbemcli -v
Connection repository with 5 connection definitions and connections file "/Users/maiera/.pywbemcli_connections.yaml" (exists)
Current connection: "None"
-Debug: PywbemServer connect to https://localhost:5989
OpenPegasus
```

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -519,20 +522,15 @@ def disconnect(self):
...
-            assert self._wbem_server is not None  # Must be connected
-        else:
-            # Old code to be removed once done
-            if self._wbem_server is None:
-                return
+        ctx = click.get_current_context()
+        if ctx.obj:
+            if ctx.obj.verbose:
+                click.echo("Disconnect from server {}".
+                           format(self._mock_server or self._server))
```
andy-maier, line 529: The term "server" is ambiguous here. I suggest to be specific about mock environment vs. WBEM server, e.g. by changing the lines with the click.echo() call to:
```
            if self._mock_server:
                server_txt = "mock environment {}".format(self._mock_server)
            else:
                server_txt = "WBEM server {}".format(self._server)
            click.echo("Disconnecting from {}".format(server_txt))
```

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -551,15 +549,13 @@ def connect(self, log=None, use_pull=None, verbose=None):
           ValueError: Other issues with server-related attributes of this
             object that are considered programming errors.
         """
-        # print("Debug: PywbemServer connect")
+        ctx = click.get_current_context()
+        if ctx.obj:
+            if ctx.obj.verbose:
+                click.echo("Connect to server {}".
+                           format(self._mock_server or self._server))
```
andy-maier, line 556: The term "server" is ambiguous here. I suggest to be specific about mock environment vs. WBEM server, e.g. by changing the lines with the click.echo() call to:
```
            if self._mock_server:
                server_txt = "mock environment {}".format(self._mock_server)
            else:
                server_txt = "WBEM server {}".format(self._server)
            click.echo("Connecting to {}".format(server_txt))
```

pywbemtools/pywbemcli/_pywbem_server.py
```diff
@@ -551,15 +549,13 @@ def connect(self, log=None, use_pull=None, verbose=None):
...
 
-        # TODO(PR900): Remove the PR900 flag test and else path once done.
-        if PR900_REQUIRE_CORRECT_CONNECTION_STATUS:
-            assert self._wbem_server is None  # Must be disconnected
-        else:
-            # Old code to be removed once done
-            if self._wbem_server is not None:
-                return
+        assert self._wbem_server is None  # Must be disconnected
```
andy-maier, line 558: I have some second thoughts on this. If in some more complex scenario we would actually close twice, it is not helpful for a user if we raise the assertion error. This is great during development but not so great during actual usage. Maybe we can introduce an env var that indicates we are running in test.

andy-maier (changes requested):
See comments.

In addition:

* You said there are now test cases that verify the correct connect/disconnect by checking for the verbose messages. I did not find those testcases in this PR.

* The first 3 commits in this PR sum up to no change and can be eliminated, by deleting or squashing them during interactive rebase.

tests/unit/test_general_options.py
```diff
@@ -1307,6 +1313,48 @@ class       Command group for CIM classes.
       'rc': 0,
       'test': 'innows'},
      None, OK],
+
+    #
+    # Test of 3 ways to execute same simple command. Creates a mock
+    # and tests combination of single command,
+    #
```
andy-maier, line 1320: These testcases depend on each other, i.e. on the previously generated connections file. That is a general problem, and we should not do that. See also issue #923.

pywbemtools/pywbemcli/_pywbem_server.py
KSchopmeyer: DONE

tests/unit/test_general_options.py
KSchopmeyer: Why not. **DISCUSSION**

andy-maier: Ok. Let's put this one in, since we have issue #923 for the .bak files.

PR #932, opened 2021-03-21
Notify workflow: RUn only when scheduled; structure changes

andy-maier:
Signed-off-by: Andreas Maier <andreas.r.maier@gmx.de>


andy-maier:
The failures are due to the recent PR in pywbem that changes the checks for ModifyClass. Merging.

PR #934, opened 2021-03-21
Fixed test failures in server cmds with ModifyClass

andy-maier:
See commit message.


KSchopmeyer (commented):
One comment. Could you and description of why the EMPY_MOCK_MODEL is used in the tests.  That could confuse others.  Other than that I approve your changes to the PR and I assume we can commit it.

Issue #395, opened 2019-09-26
"connection save" in interactive mode requires current connection


andy-maier:
Example:
```
$ pywbemcli
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli.
pywbemcli> --server http://localhost --user usr1 --password blah connection save testconn3
Error: No current connection connection. Use "connection list" to see more connection information
pywbemcli> connection list
WBEM server connections: (#: default, *: current)
+-----------+------------------+-------------+--------+-----------+----------+----------------------------------------+
| name      | server           | namespace   | user   |   timeout | verify   | mock-server                            |
|-----------+------------------+-------------+--------+-----------+----------+----------------------------------------|
| bla       | http://blahblah  | root/cimv2  |        |        30 | True     |                                        |
| local     | http://localhost | root/cimv2  | me     |        30 | True     |                                        |
| mockfuzzy |                  | root/cimv2  |        |        30 | True     | tests/unit/simple_mock_model.mof       |
|           |                  |             |        |           |          | tests/unit/simple_mock_invokemethod.py |
| mymock    |                  | root/cimv2  |        |        30 | True     | tests/unit/simple_assoc_mock_model.mof |
| testconn  | http://localhost | root/cimv2  | usr1   |        30 | True     |                                        |
| testconn2 | http://localhost | root/cimv2  | usr1   |        30 | True     |                                        |
+-----------+------------------+-------------+--------+-----------+----------+----------------------------------------+
pywbemcli>
```

Issues:
1. The "connection save" command in interactive mode as currently implemented seems to use the current connection **of the session** and overwrites that with its connection related general options. It fails because there was no current connection in the interactive session. I think that should be changed so that it saves **its own** current connection (i.e. the result of parameters inherited from the session, and specified on its own command line).
2. Issue in error message "Error: No current connection connection".

KSchopmeyer:
I agree with issue.  Will sort out.

andy-maier:
Fixed in PR #403.

PR #958, opened 2021-05-02
Fixes issue# 935,  Confusing exception display from server add_mof and remove_mof()

KSchopmeyer:
Proposal to fix the pywbemcli component of issue #395.

Modified server add_mof and remove_mofto only display the exception received from the call to
the compiler compile_string() exception if it is NOT MOFCompileError
exception.  The compile_string() method creates a log entry with any exception from calling the compiler and then raises the exception again so we were duplicating the exception display

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -70,6 +70,9 @@ Released: not yet
   and displaying of PYWBEMCLI environment variables during testing in verbose
   mode.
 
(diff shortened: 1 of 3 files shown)


andy-maier (approved):

andy-maier:
To me, this proposal looks good.

andy-maier (approved):

PR #962, opened 2021-05-04
[DROP] Added listener command group to pywbemcli

andy-maier:
**DROP:** This PR has been abandoned in favor of PR #971, but we keep it around until that PR is merged.

Initial stab at listener support. This PR adds a 'listener' command group to pywbemcli. At this point, the listener does nothing with any received indications.

Ready for a concept review and early test.

TODO:

* Add documentation
* Add tests
* Add support for using the general logging options to log the output of 'listener run'. (see discussion item, below)
* Change --certfile and --keyfile to use the corresponding global options instead. (see discussion item, below)
* Add support for processing indications: display (subject to being logged), log to file (separate from logging), call Python function. (see discussion item, below)

**DISCUSSION:**

* How to treat stdout/stderr of the running listeners. Should they log to a file?
  KS/COMMENT: I think we have a real discussion about how we log/display/count, etc. received indications.
  Result: We go with the general logging options, for now.
  Status: **OPEN**

* The general options --certfile, --keyfile and --ca-certs could be used instead of adding similar options at the level of the listener run/start commands.
  Status: **OPEN**

* What types of indication processing would we support: display, log to file, call Python function, execute command?
  Result: For now, we do: display (subject to being logged), log to file (separate from logging), call Python function.
  Status: **OPEN**

* Default listener protocol? (currently http)
  KS/COMMENT: This one does not really bother me other than the default for the server is https so this is "different".
  Andy: Changed the default to https for security by default and for consistency.
  Result: Agreed to https as a default.
  Status: DONE

* Default listener port? (currently 5988)
  KS/COMMENT: I think that Jim did set it up that way and certainly we did not claim any standard port. However 5988 always bothered me because there was always the real chance that the server was on the same system. That is why I always pick some high number like 50000.
  Andy: Changed to 25989 to have a more unique value with less chance of collisions.
  Result: Agreed to that port as a default.
  Status: DONE

* Need for a certificate trust store: Somehow the single (!) server certificate to be presented during SSL/TLS handshake needs to be determined. How would that be done with multiple certificates in a trust store? The underlying WBEMListener class takes certfile and keyfile and it passes them on to `ssl.SSLContext.load_cert_chain()`. That is straight forward. What would a trust store improve?
  KS/COMMENTS: Our listener may be more primitive but the listener should be capable of handling certs from a CA tree, possibly even multiple  CAs since, in all probably the same cert store as the server(s) with which it is defining subscriptions. For our listener we need to determine how much of that we want to do since we do not have an implemented multiple certificate store that we can simply attach to.  I should have recognized this much earlier in the project but (to be honest) I tried to avoid the whole certificate subject as long as possible (i.e. until I had to fix pegasus).
  Andy: This does not answer the question why a *server* would need more than one server certificate?
  Result: We agree that for the server certificate presented by the listener, only one certificate is needed. However, when supporting mutual TLS, a set of CA certs is needed, which Karl meant with "trust store". This item is now about adding support for mutual TLS verification.
  Status: DEFERRED to issue #965

* Do we want a `send` or `test` command?  KS: I like the idea of an indication sender that we can use in tests but is part of pywbemcli.
  Result: For now, we do only "send". It is not clear what the addiitonal value of test would be given that we cannot really test the indication was processed.
  Status: DEFERRED to issue #966

* Do we want to show the general options in the 'show' command? They are not generically available but are properties on the context object.
  Result: We want to show them. We have to, if we use the general options --certfile etc. It is also easier than I thought because they come from parsing the command line of the run commands, not from the context of the currently running list/show command.
  Status: DEFERRED to issue #967

* Does the WBEMListener class need to be extended with:
  - support for keyfile password? (see [ssl.SSLContext.load_cert_chain()](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain))
    Result: Yes
    Status: DONE - it already did that, by letting OpenSSL perform the prompt. PR https://github.com/pywbem/pywbem/pull/2695 improves that a little.
  - support for restricting version of TLS protocol? (see [Protocol versions](https://docs.python.org/3/library/ssl.html#protocol-versions))
    Result: Yes
    Status: DEFERRED to issue #964
  - support for setting the ciphers? (see [Cipher selection](https://docs.python.org/3/library/ssl.html#cipher-selection))
    Result: Yes
    Status: DEFERRED to issue #964
 
* Idea: During startup of the "run" command, have its stdout/err/in redirected to/from the parent process, in order to get success and error messages in the parent process, and also the potential prompt for keyfile password. Needs investigation.
  Status: DONE - all messages of the run process are now displayed on the correct stream before the start process terminates, and the start process exit code is also correct (i.e. shows error if the child process had an error). The password prompt is also properly displayed and is not subject to redirection.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_listener.py
+++ b/pywbemtools/pywbemcli/_cmd_listener.py
@@ -1,0 +1,833 @@
+# (C) Copyright 2021 Inova Development Inc.
+# All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Click Command definition for the listener command group which includes
+cmds to manage indication listeners that are separate invocations of
+pywbemcli representing a listener.
+
+NOTE: Commands are ordered in help display by their order in this file.
+"""
+
+from __future__ import absolute_import, print_function
+
+import os
+import subprocess
+import signal
+import atexit
+import threading
+import argparse
+from time import sleep
+from datetime import datetime
+import click
+import psutil
+import six
+
+from pywbem import WBEMListener, ListenerError
+
+from .pywbemcli import cli
+from ._common import CMD_OPTS_TXT, GENERAL_OPTS_TXT, SUBCMD_HELP_TXT, \
+    format_table
+from ._common_options import add_options, help_option
+from ._click_extensions import PywbemcliGroup, PywbemcliCommand
+
+# Print debug messages related to process handling
+DEBUG_PROCESS = False
+
+# Signals used for having the 'listener run' command signal startup completion
+# back to its parent 'listener start' process.
+# The default handlers for these signals are replaced.
+# The success signal can be any signal that can be handled.
+# In order to handle keyboard interrupts during password prompt correctly,
+# the failure signal must be SIGINT.
+try:
+    # Unix/Linux/macOS
+    # pylint: disable=no-member
+    SIGNAL_RUN_STARTUP_SUCCESS = signal.SIGUSR1
+except AttributeError:
+    # native Windows
+    # pylint: disable=no-member
+    SIGNAL_RUN_STARTUP_SUCCESS = signal.SIGBREAK
+SIGNAL_RUN_STARTUP_FAILURE = signal.SIGINT
+
+# Status and condition used to communicate the startup completion status of the
+# 'listener run' process between the signal handlers and other functions in the
+# 'listener start' process.
+RUN_STARTUP_STATUS = None
+RUN_STARTUP_COND = threading.Condition()
+
+# Timeout in seconds for the 'listener run' command starting up. This timeout
+# also ends a possible prompt for the password of the private key file.
+RUN_STARTUP_TIMEOUT = 60
+
+DEFAULT_LISTENER_PORT = 25989
+DEFAULT_LISTENER_PROTOCOL = 'https'
+
+LISTEN_OPTIONS = [
+    click.option('--port', type=int, metavar='PORT',
+                 required=False, default=DEFAULT_LISTENER_PORT,
+                 help=u'The port number the listener will open to receive '
+                 'indications. This can be any available port. '
+                 'Default: {}'.format(DEFAULT_LISTENER_PORT)),
+    click.option('--protocol', type=click.Choice(['http', 'https']),
+                 metavar='PROTOCOL',
+                 required=False, default=DEFAULT_LISTENER_PROTOCOL,
+                 help=u'The protocol used by the listener (http, https). '
+                 'Default: {}'.format(DEFAULT_LISTENER_PROTOCOL)),
+    click.option('--certfile', type=str, metavar='FILE',
+                 required=False, default=None,
+                 help=u'Path name of a PEM file containing the certificate '
+                 'that will be presented as a server certificate during '
+                 'SSL/TLS handshake. Required when using https. '
+                 'The file may in addition contain the private key of the '
+                 'certificate.'),
+    click.option('--keyfile', type=str, metavar='FILE',
+                 required=False, default=None,
+                 help=u'Path name of a PEM file containing the private key '
+                 'of the server certificate. '
+                 'Required when using https and when the certificate file '
+                 'does not contain the private key. '
+                 'Default: Certificate file.'),
+]
+
+
+class ListenerProperties(object):
+    """
+    The properties of a running named listener.
+    """
+
+    def __init__(self, name, port, protocol, certfile, keyfile, pid, created):
+        self._name = name
+        self._port = port
+        self._protocol = protocol
+        self._certfile = certfile
+        self._keyfile = keyfile
+        self._pid = pid
+        self._created = created
+
+    def show_row(self):
+        """Return a tuple of the properties for 'listener show' command"""
+        return (
+            self.name,
+            str(self.port),
+            self.protocol,
+            self.certfile,
+            self.keyfile,
+            str(self.pid),
+            self.created.strftime("%Y-%m-%d %H:%M:%S"),
+        )
+
+    @staticmethod
+    def show_headers():
+        """Return a tuple of the header labels for 'listener show' command"""
+        return (
+            'Name',
+            'Port',
+            'Protocol',
+            'Certificate file',
+            'Key file',
+            'PID',
+            'Created',
+        )
+
+    def list_row(self):
+        """Return a tuple of the properties for 'listener list' command"""
+        return (
+            self.name,
+            str(self.port),
+            self.protocol,
+            str(self.pid),
+            self.created.strftime("%Y-%m-%d %H:%M:%S"),
+        )
+
+    @staticmethod
+    def list_headers():
+        """Return a tuple of the header labels for 'listener list' command"""
+        return (
+            'Name',
+            'Port',
+            'Protocol',
+            'PID',
+            'Created',
+        )
+
+    @property
+    def name(self):
+        """string: Name of the listener"""
+        return self._name
+
+    @property
+    def port(self):
+        """int: Port number of the listener"""
+        return self._port
+
+    @property
+    def protocol(self):
+        """string: Protocol of the listener"""
+        return self._protocol
+
+    @property
+    def certfile(self):
+        """string: Path name of certificate file of the listener"""
+        return self._certfile
+
+    @property
+    def keyfile(self):
+        """string: Path name of key file of the listener"""
+        return self._keyfile
+
+    @property
+    def pid(self):
+        """int: Process ID of the listener"""
+        return self._pid
+
+    @property
+    def created(self):
+        """datetime: Point in time when the listener process was created"""
+        return self._created
+
+
+@cli.group('listener', cls=PywbemcliGroup, options_metavar=GENERAL_OPTS_TXT,
+           subcommand_metavar=SUBCMD_HELP_TXT)
+@add_options(help_option)
+def listener_group():
+    """
+    Command group for WBEM indication listeners.
+
+    This command group defines commands to manage WBEM indication listeners.
+    Each listener is a process that executes the `pywbemcli listener run`
+    command to receive WBEM indications sent from a WBEM server.
+
+    A listener process can be started with the `pywbemcli listener start`
+    command and stopped with the `pywbemcli listener stop` command.
+
+    There is no central registration of the currently running listeners.
+    Instead, the currently running processes executing the
+    `pywbemcli listener run` command are by definition the currently running
+    listeners. Because of this, there is no notion of a stopped listener nor
+    does a listener have an operational status.
+
+    In addition to the command-specific options shown in this help text, the
+    general options (see 'pywbemcli --help') can also be specified before the
+    'connection' keyword.
+    """
+    pass  # pylint: disable=unnecessary-pass
+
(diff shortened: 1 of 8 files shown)


pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+from ._common_options import add_options, help_option
+from ._click_extensions import PywbemcliGroup, PywbemcliCommand
+
+DEFAULT_LISTENER_PORT = 5988
+DEFAULT_LISTENER_PROTOCOL = 'http'
+
+LISTEN_OPTIONS = [
+    click.option('--port', type=int, metavar='PORT',
+                 required=False, default=DEFAULT_LISTENER_PORT,
+                 help=u'The port number listened on. '
```
KSchopmeyer, line 48: How about port number on listener will open to receive indications.

Also note that this can be any available port.  To be honest I never understood whey we set 5988 as defualt

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+                 'Default: {}'.format(DEFAULT_LISTENER_PORT)),
+    click.option('--protocol', type=click.Choice(['http', 'https']),
+                 metavar='PROTOCOL',
+                 required=False, default=DEFAULT_LISTENER_PROTOCOL,
+                 help=u'The protocol used by the listener (http, https). '
+                 'Default: {}'.format(DEFAULT_LISTENER_PROTOCOL)),
+    click.option('--certfile', type=str, metavar='FILE',
+                 required=False, default=None,
+                 help=u'Path name of a PEM file containing the certificate '
+                 'that will be presented as a server certificate during '
```
KSchopmeyer, line 58: This one is going to be fun.  We will have to clarify the server cert, etc. since in this case the listener is the server and the indication exporter is the client.  Note that we will eventually have to make this more than just a certificate since this really requires a certificate trust store.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+                 'SSL/TLS handshake. Required when using https. '
+                 'The file may in addition contain the private key of the '
+                 'certificate.'),
+    click.option('--keyfile', type=str, metavar='FILE',
+                 required=False, default=None,
+                 help=u'Path name of a PEM file containing the private key '
+                 'of the server certificate. '
```
KSchopmeyer, line 65: again server vs listener, etc.  This also probably will become a cert trust store but lets not get messy now.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+    """
+    context.execute_cmd(lambda: cmd_listener_start(context, options))
+
+
+@listener_group.command('stop', cls=PywbemcliCommand,
+                        options_metavar=CMD_OPTS_TXT)
+@click.argument('name', type=str, metavar='NAME', required=True)
+@add_options(help_option)
+@click.pass_obj
+def listener_stop(context, name):
```
KSchopmeyer, line 192: This one could be extended to present a list of running listeners from which one is selected to stop

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+    """
+    context.execute_cmd(lambda: cmd_listener_stop(context, name))
+
+
+@listener_group.command('show', cls=PywbemcliCommand,
+                        options_metavar=CMD_OPTS_TXT)
+@click.argument('name', type=str, metavar='NAME', required=True)
+@add_options(help_option)
+@click.pass_obj
+def listener_show(context, name):
```
KSchopmeyer, line 212: This could also show from a list of running listeners.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+    A listener with that name must be running, otherwise the command fails.
+
+    Examples:
+
+      pywbemcli listener stop lis1
+    """
+    context.execute_cmd(lambda: cmd_listener_show(context, name))
+
+
+@listener_group.command('list', cls=PywbemcliCommand,
```
KSchopmeyer, line 225: I had thought about combining list and show (i.e. list with a GLOB for names to show). However, I assume that we will want to present more information about number of indications received, etc. and that can be done through show so combining is an open questions.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+        """Return the object as a tuple of strings"""
+        return (str(self._pid), self._name, str(self._port), self._protocol,
+                ' '.join(self._general_options))
+
+    @property
+    def pid(self):
+        """int: Process ID of the listener"""
+        return self._pid
+
+    @property
```
KSchopmeyer, line 93: Suggest adding start time as property on listener.

KSchopmeyer (commented):
Made some particular comments and a couple of general thoughts.  Generally I like this. Fairly simple and straight forward but covers the options.

KSchopmeyer:
Made first overview review but have not yet run tests against OpenPegasus.  Will do that Wednesday.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+@cli.group('listener', cls=PywbemcliGroup, options_metavar=GENERAL_OPTS_TXT,
+           subcommand_metavar=SUBCMD_HELP_TXT)
+@add_options(help_option)
+def listener_group():
+    """
+    Command group for WBEM indication listeners.
+
+    This command group defines commands to manage WBEM indication listeners.
+    Each listener is a running process that executes the
+    `pywbemcli listener run` command acting as a listener.
```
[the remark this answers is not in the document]
andy-maier: DONE.
We do have a term for the indication client - it is "WBEM server". We defined that e.g. in section 6.1 of DSP0200: "A WBEM server receives and processes CIM-XML operation requests and issues CIM-XML operation responses. A WBEM server also issues CIM-XML export requests and receives and processes CIM-XML export responses."

Changed it to "to receive WBEM indications sent from a WBEM server."

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+
+    These processes can be started manually by the user by invoking the
+    `pywbemcli listener run` command in the background, or by using the
```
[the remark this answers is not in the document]
andy-maier: DONE - please re-review.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+    `pywbemcli listener start` command which does exactly that.
+
+    There is no persisted data about the currently running listeners, instead
```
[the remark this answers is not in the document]
andy-maier: DONE - changed it to "There is no central registration of the currently running listeners."

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+           subcommand_metavar=SUBCMD_HELP_TXT)
+@add_options(help_option)
+def listener_group():
+    """
+    Command group for WBEM indication listeners.
+
+    This command group defines commands to manage WBEM indication listeners.
+    Each listener is a running process that executes the
+    `pywbemcli listener run` command acting as a listener.
+
```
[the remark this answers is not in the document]
andy-maier: DONE - Added that to the start command.

pywbemtools/pywbemcli/_cmd_listener.py
```diff
@@ -0,0 +1,473 @@
...
+    context.spinner_stop()
+    # pylint: disable=consider-using-with
+    p = subprocess.Popen(run_args, start_new_session=True)
+    click.echo("Started listener {} as detached process {}".format(name, p.pid))
+
+
+def cmd_listener_stop(context, name):
+    """
+    Show the properties of a named listener.
```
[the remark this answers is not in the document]
andy-maier: DONE

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: DONE - changed help to "The port number the listener will open to receive indications. This can be any available port.". The default port is a discussion item.

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: On listener assuming server role in HTTPS: That's why I explained that this is presented by the listener as a server certificate. I agree that this requires some more words, but that would be in a security section of the two be written documentation for the listener command group.
DEFERRED to future update of this PR.

On certificate trust store:  The underlying WBEMListener class takes exactly these two arguments, and it directly passes them on to `ssl.SSLContext.load_cert_chain()`. We need to discuss this. Added it to the discussion items.
DEFERRED to future update of this PR.

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: DEFERRED - see previous comment.

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: Is that really needed? How many listeners do you expect?

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: You mean a user can interactively pick one? Is that really needed? How many listeners do you expect?

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: Right now, they both show the same information because it is short enough. Once we add properties like what action is taken, where is the output logged, etc, it may become different.

pywbemtools/pywbemcli/_cmd_listener.py
andy-maier: Good idea! Added to TODO items. DEFERRED to future update of this PR.

andy-maier:
I resolved most of your comments, some by adding an item to the TODO list. Please review that, and particularly the comments not yet marked as resolved. Also, please review the new help text of the listener group again.

pywbemtools/pywbemcli/_cmd_listener.py
KSchopmeyer: Lets ignore my comment at least until we better understand the listener

pywbemtools/pywbemcli/_cmd_listener.py
KSchopmeyer: Ignore until we better understand how people use listeners.

pywbemtools/pywbemcli/_cmd_listener.py
KSchopmeyer: OK as it stands.

andy-maier:
DISCUSSION:** Should we move the listener functionality into a separate command?

A major consideration for this are the general options:

Options only for targeting a WBEM server:
```
  -n, --name NAME
  -m, --mock-server FILE
  -s, --server URL
  -u, --user TEXT
  -p, --password TEXT
  -t, --timeout INT
  -U, --use-pull [yes|no|either]
  --pull-max-cnt INT
  -d, --default-namespace NAMESPACE
  -T, --timestats / --no-timestats
  -C, --connections-file FILE PATH
```

Common options with different meanings:
```
  --verify / --no-verify
  --ca-certs CACERTS
  -c, --certfile FILE
  -k, --keyfile FILE
```

Common options with same meaning:
```
  -o, --output-format FORMAT
  -l, --log COMP[=DEST[:DETAIL]],...
  -v, --verbose / --no-verbose
  --warn / --no-warn
  --pdb
  --version
  -h, --help
```

andy-maier:
**DISCUSSION:** If a new 'pywbemlistener' command is the direction (see PR #971), we can abandon this PR.

PR #968, opened 2021-05-11
Prepared development environment for multiple pywbemtools commands

andy-maier:
See commit message.

This is a first stab at supporting multiple commands, but there may be more coming as we actually do it.


Makefile
```diff
@@ -141,17 +141,9 @@ endif
...
-
-# Name of the Python package modules
-# TODO future: expand so multiple modules are allowed. Right now we have
-# one module and directly reference it in the Makefile
-pywbemcli_module_path := $(package_name)/$(module_name)
-
-pywbemcli_module_import_name := $(package_name).$(module_name)
+# Names of the commands in the package. The command names are used in
+# certain directory names.
+command1 := pywbemcli
```
KSchopmeyer, line 146: Just to make more explanable could we chage to commandpywbemcli or something rather than 1

docs/changes.rst
```diff
@@ -23,8 +26,13 @@ Released: not yet
 
 **Enhancements:**
 
+* Added support for managing WBEM indication listeners by adding a command
```
KSchopmeyer, line 29: We should also state that we moved a number of utility functions from pywbemcli to a _utils module.

KSchopmeyer (commented):
I went throught the files.  Made a lot simpler when I blocked out all the ones that just changed name.  Couple of comments but I agree with the changes. Couple of questsion.  Should the third package be utils or common or should there be a utils and a common..  Right now everything fits into utils but will there be a need for a group of things that are really common code between apps and not just utilities for the various commands? Not sure right now so it is just a question for thought.

KSchopmeyer:
Not many comments for 74 changed files.

docs/changes.rst
andy-maier: DONE - added a sentence to the existing item in the Cleanup section. Also, since this is supposed to just be the preparation, removed the entry about the listener in the Enhancements section.

Makefile
andy-maier: NO CHANGE - When searching for direct uses of 'pywbemcli' it helps if it does not show up in the variable names.

andy-maier:
I have addressed the comment on teh change log entries but did not address the one about the Make variable name. See above for reasons. Merging.

PR #974, opened 2021-05-13
Moved _click_extensions.py module to be common across commands

andy-maier:

### DIFF
--- a/pywbemtools/_click_extensions.py
+++ b/pywbemtools/_click_extensions.py
@@ -1,32 +1,35 @@
 """
-This file contains extensions to Click specifically for pywbemcli
+This file contains extensions to Click for pywbemtools.
 """
+
 import sys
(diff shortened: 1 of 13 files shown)


pywbemtools/_click_extensions.py
```diff
@@ -1,47 +1,50 @@
 """
-This file contains extensions to Click specifically for pywbemcli
+This file contains extensions to Click specifically for pywbemtools.
```
KSchopmeyer, line 2: Lets get rid of the word specifically.  It is redundant. My original error

KSchopmeyer (commented):
Only one comment and approved.

pywbemtools/_click_extensions.py
andy-maier: DONE

PR #976, opened 2021-05-14
Moved options related definitions into a common _options.py module

andy-maier:


KSchopmeyer (commented):
This creates a file at the pywbemtools level named _options.py.  Would it be logical to have a more precise name for this file so the user did not interpret it as all options code. (Ex. options_general, options_common, etc.)

KSchopmeyer (commented):
My only comment was about the name of the file. _options.py. Otherwise approved.

PR #977, opened 2021-05-14
Moved formatting-related funcs/consts into common _output_formatting.py module

andy-maier:


KSchopmeyer (approved):

KSchopmeyer:
Approved

PR #978, opened 2021-05-14
Renamed common functions pywbemcliwarn* to pywbemtools_warn*

andy-maier:

### DIFF
--- a/pywbemtools/_utils.py
+++ b/pywbemtools/_utils.py
@@ -111,19 +111,19 @@ def _formatwarning(message, category, filename, lineno, line=None):
     return "{}: {}\n".format(category.__name__, message)
 
 
-def pywbemcliwarn(*args, **kwargs):
+def pywbemtools_warn(*args, **kwargs):
     """
-    Pywbemcli monkey patch for the warnings.warn function; substitutes our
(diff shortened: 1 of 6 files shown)


pywbemtools/_utils.py
```diff
@@ -111,18 +111,18 @@ def _formatwarning(message, category, filename, lineno, line=None):
     return "{}: {}\n".format(category.__name__, message)
 
 
-def pywbemcliwarn(*args, **kwargs):
+def pywbemtools_warn(*args, **kwargs):
     """
-    Pywbemcli monkey patch for the warnings.warn function; substitutes our
+    Pywbemtools monkey patch for the warnings.warn function; substitutes our
```
KSchopmeyer, line 116: Might clarify by explaining what and why of "our own formatter". Repeated at least one other time.  Otherwise approved

KSchopmeyer (commented):
One comment and approved

pywbemtools/_utils.py
andy-maier: DONE - changed description to "Pywbemtools version of the warnings.warn() function, with replaced formatting."

PR #979, opened 2021-05-14
Moved PYWBEMCLI_* envvars from PywbemServer class to become constants

andy-maier:
See commit message.


KSchopmeyer (commented):
Why do this one.  Are these really configuration variables? I am not sure.  Originally I put them in the class because that kept the definition close to where the corresponding internal usage was.  I will go either way but at least, I propose adding a comment in pywbem server reminding the programmer to review these enironment variables if changes are made.  

For the change itself I approve except for my question of why.

KSchopmeyer:
There are two bugs in the code that do not allow run.  

a) namep pywbemcli_warn used in place of pywbemtools_warn
b) All of the env_var definitions in the options decorartors  of pywbemlistener.py define entities in pywbemcli, not in pywbemlistener.

KSchopmeyer:
Temporarily fixed and will do tests.

andy-maier:
The reason the variables are moved is that they have nothing to do with PywbemServer, and they are also not used in PywbemServer. Their only usage is from outside of PywbemServer.

Our "config" module is not really a config module in the sense that users can configure things there. We just copied it from pywbem, where users can actually change configuration, but in a CLI usrers do not program, so they could not change the configuration anyway. For all practical purposes, "config" is a module with global constants. Which pretty much fits the nature of the PYWBEMCLI_* variables (aka constants).

KSchopmeyer:
Agree.  Approved

KSchopmeyer (approved):

PR #987, opened 2021-05-21
Add new pywbemcli command group subscriptions

KSchopmeyer:
This PR adds the new command group subscriptions. This command group includes commands to add, remove, and list subscriptions, filters, and listener destinations.  It also adds the corresponding tests and documentation for the new command group.

### Status of PR
1. Code complete for all of the subscription methods that add/list/remove destinations, filters, and subscriptions and remove server command.
2. A couple of items I marked Future  in _cmd_subscription.py. Would not expect to do these in current PR.
3. Test coverage 89%. Test coverage for errors incomplete,  List of about 15 more tests possible in test.subscription_cmds.py file.
4. Code restarts SubscriptionManager for each new command even in interactive mode. This means it gets instances from the server for each command executed. It would seem logical to pass the Subscription manager from command to command through either the pywbem_server or context object to avoid having to get the subscriptions from the server for each subscription command.  Propose second pr see issue #1067.   Note that this approach also has issues in that we would be reusing the list of instance previously retrieved from the server while some other client (or even the server could be changing the instances on the server). There is a DISCUSSION item on this work because keeping the subscription between commands leads to issues of caching.
5. Documentation complete but I am the only person who has reviewed it.  Have requested couple of other reviews but no responses yet.

### New Discussion Items:

None

### TODOS:
*  Add more tests. Have list of sever more tests in the test_subscriptions_cmd.py file. Current coverage is 89%
* Add server tests based on pegasus container (Make this separate branch)
* Make changes in accord with the discussions above.
   b. Add ability to retain SubscriptionManager between commands in interactive mode to avoid reloading subscriptions from server for each command. See issue #1056. This will be separate PR
  c. Extend the table display so that rows that are properties based on the Values qualifier include both components.  Right now
     the show only the numeric value of the property. This will be separate PR.
  d.  Make persistence type on add-destination an option.  **Proposal:** No for this PR.  We will wait for requirements to do this.

### Previous Resolved Design discussion items:

1. **DISCUSSION**  **RESOLVED BY going to IDENTITY ARGUMENT, see above**.I rewrote the subscription commands to handle the new destination form and API.  This makes tthe whole thing more consistent but leads to a real difference.
   a. To be consistent in all the commands I used the options --destination-id, --destination-name, --filter-id, --filter-name. While only the --name is needed for add-filter and add-destination, in order to refer to both the filter and the destination in add-subscription you need all 4 of parameters and options is logical.  However the result is very hard for the user to understand In think as well as being very long.  Here are some working examples:

```
# NOTE: I do not include the --owned parameter here. The code selects that from the use of ..._id, vs ...name
subscription add-destination -l http://blah:5000 --destination-id=dest1
subscription add-destination -l https://blah:5001 --destination-id=dest2
subscription add-destination -l http://blah:5003 --destination-name dperm1
subscription add-destination -l https://blah:5004 --destination-name dperm2
subscription add-destination -l http://blah:5005 --destination-name dperm3
subscription add-destination -l https://blah:5006 --destination-name dperm4

subscription add-filter --filter-id filter1 -q "SELECT * from CIM_Indication"
subscription add-filter --filter-id filter2 --query="SELECT fred from CIM_Indication" --owned
subscription add-filter --filter-id filter3 -q "SELECT * from CIM_Indication" --query-language DMTF:CQL --owned
subscription add-filter --name perm1 -q "SELECT * from CIM_Indication" --query-language DMTF:CQL --permanent
subscription add-filter --name perm2 -q "SELECT * from CIM_Indication" --query-language DMTF:CQL --permanent

subscription add-subscriptions --destination-id dest1 --filter-id filter1 --owned
subscription add-subscriptions --destination-id dest2  --filter-id filter1
subscription add-subscriptions --destination-name dperm1 --filter-name perm2 --permanent

```
    The reason for using destination-name and filter-name as the option names is because those distinctions are required for add-subscription (--name would exist for both destination and filter).  

    I have considered an option for the parameters of these commands:

    a. for filter and destination use argument to set the name component and --owned to clarify whether it is owned.  thus add-filter would look like 
     
 ```
     subscription add-filter <id-string> <options.
     subscription add-filter permfilter1 --permanent --query "..."
     subscription add-filter ownedfilter1 --owned  --query "..."  

```
    This eliminates the --filter-id, --destination_id but would mean that the name/id concept would not appear in add-filter and add-destionation.  Further since these names are used when adding subscriptions, things really get messy since the subscription must provide id for both filter and destination. Thus, subscriptions would have to have something like :+1: 

```
    subscription add-subscription filter-id ownedfilter1 filter-type owned destination-id dest1 dest-type owned

```
   This means that the user has possibly 3 parameters involving ownerhip, the filter owner, the destination owner, and the subscription owner.
   Because there is no uniqueness requirement between the filter_id string and the filter_name string so the unique identity of a filter cannot be determined just from the string itself but by the combination of name and ownership. 
 

1.  **DISCUSSION:** How we pick destinations to delete and to create the subscription.  
   c) Destinations - We currently use the destination property as our  identification  for removing owned destinations.  Therefore, there can be only a single owned destination for each destination url to keep the urls unique since there is no model uniqueness requirement for urls (They probably are unique since the combination of scheme, hostname, port should be unique). For --permanent destinations, there are no limits but that means that we have a real issue defining subscriptions since there could be multiple handlers with the same destination url.  NOTE: We reuse a owned destination if a destination object exists with the same url.
 **CONCLUSION:**  **DONE**I think this is adequate but need to test some more.

2. **DISCUSSION:** We are not using the capability to have multiple servers in a single subscription context since pywbemcli focuses on a single server. Alternative could be a "global" Subscription context in interactive mode but that makes no sense to me.  **PROPOSAL:** This is correct, so no change required.
**CONCLUSION:**  **DONE** We will not use this capability for now.

3. **DISCUSSION:** **DISCUSSION DONE** The SubscriptionManager allows for multiple SubscriptionManager Ids.  I put the parameter into the subcommands but with a default "Pywbemcli_sub_mgr" but commented it out for the moment.  What should we consider the goal of the this variable, identify pywbemcli, identify a particular client user, etc.? Some possibilities include:
**CONCLUSION:** we want something where the user can define a subscription manager ID with some level of persistence as the default for a series of operations and NOT be tied to typing it in with each command. 
We want to get this out to the level of something like environment variable.

   **Andy:** (Updated 09/04) The subscription_manager_id is described "to help the user identify these instances in a WBEM server.". Where it is used in instance property values, it is always accompanied by the client hostname, and pywbem is also identified in the property values. I think it should be defaulted to a standard string such as "default" and have overrides with an env var and an option.

  **Karl.**  * Are we opening any security doors if we make the user name public by making it part of every subscription???? 
                 * Note that the user name is (strangely one of the more difficult things to acquire correctly in python (may mean adding new requirement ( ex. psutil) to cover all supported OS's. **COMMENT/KS** We agreed that the user name is probably not the best approach and will continue with a default name and the ability to set an alternate name.

4. **DISCUSSION:** **The filter_id vs name mechanism in adding filters.**  **DONE** With the latest change to pywbem, the filter_id is used only with owned filters and the name only with permanent filters. Therefore use of the filter_id as an option in create-  filter implies owned and use of --owned implies that the options --filter_id will become our own format for the Name property.  
**CONCLUSION**:   FIXED with pywbem PR https://github.com/pywbem/pywbem/pull/2764.   We need to separate out the three pieces (filter_id, name, owned) so the user has a clean choice with these options in the various commands.  Right  now we have both filter_id and name options and the owned option.
**PROPOSALS:**
   a. Use just --name and --owned.  If --name and --owned, we use the name as filter_id.  If --name without owned it is used as the string supplied with the option.
  b. Use just --name and --filter_id with create-filter and remove-filter and do not include the --owned option for add_filter.  It becomes owned if --filter_id is used.  We could then drop the --owned.--permanent option from add_filter.    **I prefer this proposal**  The question is also, why do we include the --owned parameter in pywbem for add-filter since ownership is  determined by the use of the --filter-id/--name parameters.

   **Andy:** There are 3 possibilities for specifying the `add_filter()` parameters, and not just 2. See [my comment in pywbem issue 2757](https://github.com/pywbem/pywbem/issues/2757#issuecomment-906093611)
   **Update:KS** This was changed meanwhile so that we now have only two options: owned filters with filter_id, and permanent filters with name.  The code has been updated on 7 Sept to reflect the new form for  the name property. **DONE**

  **Karl** I noted in issue pywbem 2757 that there is an issue with the filter-id / name code vs. documentation.  The documentation appears to state that only owned filters can have filter_id. The code allows both owned and not-owned filters to have filter_id unless I am reading something wrong.  I looked at the documentation and not the code the first time.

**Karl** I am keeping the two parameters --filter-id and --name throughout the cmd line options.  Code updated to reflect the change to pywbem. **DONE**

5. **DISCUSSION:**  **Selecting filters to be removed:**  **DONE** The code currently allows you to define a filter to be removed by --filter-id or --name.  When --name is used, we assume that the the --name is unique since the name property is a key property.  However, since the filter_id is not guaranteed to be unique, this can result in multiple filters in the list to remove as shown in the following example:

   **Andy:** There is no filter-id in the filter instance. This value exists only in order to construct the "Name" property, and there is a UUID created in the "Name" property when it is used to create filter instances. Therefore, filter-id cannot be used to identify existing filter instances. I think we should use the "Name" property to identify the filter, and have some mechanism to deal with its possible but unlikely non-uniqueness. For example, we could allow using the key list portion of a WBEM URI to deal with it:

   * `name1` -> just "Name", reject if non-unique
   * `Name="name1",CreationClassName="class1",...`

   Since non-uniquenes is unlikely, it does not matter much if the fallback mechanism to deal with it is not so nice.
```
**Karl Reply**.  Actually, there can be the concept of filter-id as far as the user is concerned in that the filter_id is a component of a particular formatting of the name property.  We require filter-id parameter to add an owned filter so, therefore, for the user there is a concept of filter-id. Thus if we request remove of a filter with filter_id = filter1, the code searches for filters that have a Name with our structure of the value and the fiilter_id component of filter1.  The name property for owned filters is not pretty, for example.  

    "pywbemfilter:owned:leonard:pywbemcliSubMgr:filter2:f1e85edf-b2d5-4be2-b71d-fe3d6b1c761e";

The only part of this that is useful for the user is a) the filter-id part, and b) the guid only if they would have to select between two filters with the same filter-id

```
Presenting the paths  to the user to pick a filter to remove leaves the user with a guessing game in determining which filter to remove. There is not enough information in the path for the user to decide.  We need to present information that will allow the user to separate out which filter to remove.  Thus, it would appear we should present a. the query language, b) the query, c) the source namespaces since any of these could separate the filters.

Further, this is unworkable in any automated environment since the user does not know when the remove just works or returns with a select statement.  

**Possible solutions:**
  a,  pywbemtools insures that the filter_ids are unique when filters are created.  **COMMENT** We agreed that this was what we wanted and this is in the pywbem master now.
  b. Add a flag that would control behavior if multiple filters were selected. In the normal mode the command would generate an exception.  This flag (ex. --select) would present the list and the user would pick the filter to remove. 
  c. Release all filters with that filter_id, under assumption that user knows what they want to do.  
6. **DISCUSSION** Add one more option to remove-subscriptions, (`--remove-associated-instances`) that would attempt to remove also the filters and destinations associated with a subscription being removed (if they had no other associations). I think this is logical other than the very long name for the option
   **Andy:** I think automatic removal of the referenced filters and destinations should only be done for filters and destinations that are owned by the current client. That is the whole point of the ownership concept: to be sure that they are yours. The option could be named `--remove-owned`. 
  **Karl** - I agree.  It would be better to make this easier for owned destinations, etc. and harder for permanent destinations, etc. I will add the remove option but it should indicate that we are removing associations like `remove-associated-instances` or `remove-everything`, etc.  This would only remove associations if they were not associated with other subscriptions and if they  werre owned, themselves as well as the subscription being owned.  Thus, an owned subscription with owned destination but permanent filter would only remove the subscription and destination, but not the filter.

**CODE STATUS:** The basic structure of the code is complete and all of the commands implemented. However, the output formats are purely samples, and the mechanisms for choosing filter_id/name to delete or as option in lists is being revised now.
6. **DISCUSSION**  Use of GUID with name in destination path.  Do we agree that the current form of the CIM_IndicationDestination Name propery will be kept the same with the form shown below or will we modify this one also?

     "pywbemdestination:permanent:leonard:defaultpywbemcliSubMgr:d9a6b7aa-aaa4-45ab-bfcd-469524ec7a40";

This form has tthree issues:
1. It is very long with the guid. While the guid is truly unique it is awkward to use in a terminal environment.
2. It includes the client host name.
3. It does not give the user any indication of what the destination really is (i.e what is the url or other identity info) so using --select to pick one to release is a guessing game. Currently I create a "short" string with the Name, url, if one of multiple destinations is to be selected.

### DIFF
--- a/docs/appendix.rst
+++ b/docs/appendix.rst
@@ -73,6 +73,16 @@ This documentation uses a few special terms to refer to Python types:
       non-embedded instances of the class (not including subclasses) within the
       namespace.
 
+   Interop namespace
+      The Interop namespace is a :term:`CIM namespace` that exists in a WBEM
+      server as a point of interoperatbility with WBEM clients because it has a
+      well known name. A WBEM server may have only a single Interop namespace
+      and the names are limited normally to ``interop``, ``root/interop``.
+      Pywbemcli adds ``root/PG_Interop`` as a third alternative when it looks
+      for any Interop namespaces on a WBEM server.  The object manager,
+      profile, namespace information and indication subscription is available
+      through the interop namespace.
+
    connections file
       A file maintained by pywbemcli and managed by the
       :ref:`Connection command group`.
@@ -208,11 +218,47 @@ This documentation uses a few special terms to refer to Python types:
       if a :term:`default connection`) has been defined. The current connection can be
       changed in the interactive mode with the :ref:`connection select command`.
 
+   default namespace
+      The :term:`CIM namespace` set as the current namespace for CIM operations
+      if pywbemcli is started with a connection defined but without a specific
+      namespace defined.  Pywbemcli uses ``root/cimv2`` as the default
+      namespace.
+
    CQL
       CQL (CIM Query Language) is a query language defined by DMTF for use
       by query operations against WBEM servers. In operation parameters that
       define the use of a query language, it is specified with the string
-      ``DMTF:QCL``. CQL is described in DMTF standard :term:`DSP0202`.
+      ``DMTF:CQL``. CQL is described in DMTF standard :term:`DSP0202`.
+
+   WQL
+      WQL (WBEM Query Language) is a query language defined by Microsoft for use
+      by query operations against WBEM servers. In operation parameters that
+      define the use of a query language, it is specified with the string
+      ``WCL``. CQL is described in informal specifications generally available
+      on the internet. WQL is not a DMTF standard and there are differences
+      in implementations partly because of different versions of the WQL specification.
+      A number of WBEM servers implement WQL query processors because WQL was
+      the first query language defined for CIM query operations.
+
+   query language
+      A language defined to support query operations against WBEM servers
+      specifically the ``ExecuteQuery`` operation and its corresponding pull
+      operation. A WBEM server may implement any desired query language but the
+      DMTF specifies the characteristics of the :term:`CQL` the :term:`WQL`
+      query language is specified by Microsoft documents.
+
+   filter query language
+      A query language defined to support the DMTF pull instance enumeration
+      operation (ex. OpenEnumerateInstances, PullEnumerateInstances) specifically
+      so that filtering of responses based on a query statement can be integrated
+      into the enumeration of the instances.  The DMTF defined the filter query
+      language ``FQL`` (defined in DMTF specification (:term:`DSP0212`))as one
+      possible filter query language
+
+      See :term:query languages for query languages that support the ``ExecuteQuery``
+      operation and its corresponding pull  operation.
+
+      Query languages and filter query languages are not interchangable.
 
    schema
       A schema (named CIM schema in the DMTF specification :term:`DSP0004`) in
@@ -229,6 +275,37 @@ This documentation uses a few special terms to refer to Python types:
       number (ex. version 2.41.0). The DMTF CIM Schemas can be retrieved from
       the DMTF web site at: `<https://www.dmtf.org/standards/cim>`_ .
 
+   CIM indication subscription
+   indication subscription
+      A CIM indication subscription is the definition of :term:`listener destination`
+      and :term:`indication filter` contained within an association class
+      (``CIM_IndicationSubscription`` class) that defines for a WBEM server the
+      characteristics of indications to be generated and where they are to be
+      sent. The  listener destination is defined by the ``Handler`` reference
+      property (class name ``CIM_ListenerDestination``) and the filter is
+      defined by the ``Filter`` reference property(class name
+      ``CIM_IndicationFilter``). The DMTF indication profile and is defined in
+      :term:`DSP0154`
+
+   CIM indication destination
+   listener destination
+      A listener destination is an instance of a CIM class that maintains a
+      reference to a listener within an implementation. Specifically within
+      the scope of pywbemcli it is an instance of the class
+      ``CIM_ListenerDestinationCIMXML`` where the listener reference is the
+      ``Destination`` property which contains the URL of the target
+      indication listener.
+
+   CIM indication filter
+   indication filter
+      An indication filter is an instance of a CIM Class
+      (``CIM_IndicationFilter``) that defines characteristics of indications
+      that the WBEM server will generate. The basis for the indication
+      characteristics is the query language and query statement properties of
+      the instance.
+
+   dynamic indication filter
+      An :term: indication filter whose lifecycle is controlled by a client
 
 .. _`Profile advertisement methodologies`:
 
@@ -372,6 +449,9 @@ References
 
    DSP1033
       `DMTF DSP1033, Profile Registration Profile, Version 1.1 <https://www.dmtf.org/standards/published_documents/DSP1033_1.1.pdf>`_
+
+   DSP0154
+      `DMTF DSP1054, Indications Profile, Version 1.2.2, <https://www.dmtf.org/sites/default/files/standards/documents/DSP1054_1.2.2.pdf>`_
 
    RFC3986
       `IETF RFC3986, Uniform Resource Identifier (URI): Generic Syntax, January 2005 <https://tools.ietf.org/html/rfc3986>`_
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -102,6 +102,13 @@ Released: not yet
 * Enhanced test matrix for push-driven runs on GitHub Actions to add
   Python 3.5 on macOS, and removing Python 3.5 minimum on Windows.
 
+* Implement command group subscription that manages the creation, viewing and
+  removal of indication subscription on WBEM servers. This creates a new command
+  group 'subscription' and new commands for adding, removing, and displaying
+  (list) indication destination, filter, and subscription instances on target
+  WBEM servers. It includes the code for the new commands, a set of tests
+  and the documentation for the new commands. (see issue #4)
+
 **Cleanup:**
 
 * Prepared the development environment for having more than one pywbemtools
--- a/docs/pywbemcli/cmdshelp.rst
+++ b/docs/pywbemcli/cmdshelp.rst
@@ -144,16 +144,17 @@ Help text for ``pywbemcli``:
       -h, --help                      Show this help message.
 
     Commands:
-      class       Command group for CIM classes.
-      instance    Command group for CIM instances.
-      namespace   Command group for CIM namespaces.
-      profile     Command group for WBEM management profiles.
-      qualifier   Command group for CIM qualifier declarations.
-      server      Command group for WBEM servers.
-      statistics  Command group for WBEM operation statistics.
-      connection  Command group for WBEM connection definitions.
-      help        Show help message for interactive mode.
-      repl        Enter interactive mode (default).
+      class         Command group for CIM classes.
+      instance      Command group for CIM instances.
+      namespace     Command group for CIM namespaces.
+      profile       Command group for WBEM management profiles.
+      qualifier     Command group for CIM qualifier declarations.
+      server        Command group for WBEM servers.
+      statistics    Command group for WBEM operation statistics.
+      subscription  Command group to manage WBEM indication subscriptions.
+      connection    Command group for WBEM connection definitions.
+      help          Show help message for interactive mode.
+      repl          Enter interactive mode (default).
 
 
 .. _`pywbemcli class --help`:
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

**Comment by KSchopmeyer** [comment]
Help output showing current options.

```
Usage: pywbemcli [COMMAND-OPTIONS] subscription COMMAND [ARGS]...

  Command group to manage WBEM subscriptions.

  This group uses the pywbem subscription manager to view and manage CIM
  Indication subscripitons for a WBEM Server.

  In addition to the command-specific options shown in this help text, the
  general options (see 'pywbemcli --help') can also be specified before the
  command. These are NOT retained after the command is executed.

Command Options:
  -h, --help  Show this message and exit.

Commands:
  add-destination      Create a new listener destination.
  add-filter           Create a new subscription based on destination and...
  add-subscription     Create a new subscription based on destination and...
  delete-destination   Delete a destination
  delete-filter        Delete a subscription filter
  delete-subscription  Delete a subscription
  list-destinations    list subscription destinations objects List the...
  list-filters         list subscription filters List existing CIM indication...
  list-subscriptions   list indication subscriptions.
  test_indication      Generate test indications.
Usage: pywbemcli [COMMAND-OPTIONS] subscription add-destination

  Create a new listener destination.

  The listener destination defines the location of a WBEM indication listener
  and includes all the parameters to define that listener

  The listener destination is automatically created on the target WBEM
  connection.

Command Options:
  -d, --destination destination   Add a listener destination url for
                                  indications. this includes the schema (http or
                                  https and host name.

  -d, --listener-port listener_port
                                  Listener destination port to receive
                                  indications

  -i, --submgr-id TEXT            Optional subscription manager id that can be
                                  used to separate subscriptions created by
                                  pywbem. Default: pywbemcliSubMgr

  --owned                         Limit the operation to owned subscriptions,
                                  listener destinations or filters. Default:
                                  False

  -h, --help                      Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription add-filter

  Create a new subscription based on destination and filter provided as input

Command Options:
  -q, --query filter       Filter query definition. This is a SELECT statement
                           in the query language defined in the filter-query-
                           language parameter

  --query-language TEXT    Filter query language for this subscription The query
                           languages normally implemented are "DMTF:CQL" and
                           "WQL" .  Default: WQL

  --source-namespace TEXT  The namespace for which the query is defined. If not
                           defined, the default namespace of the server is used

  --name TEXT              Value for the name parameter.
  --filter-id TEXT         Text that is used to create the name parameter.
  -i, --submgr-id TEXT     Optional subscription manager id that can be used to
                           separate subscriptions created by pywbem. Default:
                           pywbemcliSubMgr

  --owned                  Limit the operation to owned subscriptions, listener
                           destinations or filters. Default: False

  -h, --help               Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription add-subscription

  Create a new subscription based on destination and filter provided as input

Command Options:
  --select              Select filter and destination to include in this
                        subscription.

  -i, --submgr-id TEXT  Optional subscription manager id that can be used to
                        separate subscriptions created by pywbem. Default:
                        pywbemcliSubMgr

  --owned               Limit the operation to owned subscriptions, listener
                        destinations or filters. Default: False

  -h, --help            Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription delete-destination

  Delete a destination

Command Options:
  -n, --path destination  Delete by path name of the filter.
  --select                Select filter to delete from select list of filters.
  -h, --help              Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription delete-filter

  Delete a subscription filter

Command Options:
  -n, --path destination  Delete by path name of the filter.
  --select                Select filter to delete from select list of filters.
  -h, --help              Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription delete-subscription

  Delete a subscription

Command Options:
  -n, --path destination  Delete by path name of the filter.
  --select                Select filter to delete from select list of filters.
  -h, --help              Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription list-destinations

  list subscription destinations objects

  List the existing CIM indication destinations on the current connection

Command Options:
  -s, --summary         If True, show only summary count
  -i, --submgr-id TEXT  Optional subscription manager id that can be used to
                        separate subscriptions created by pywbem. Default:
                        pywbemcliSubMgr

  --owned               Limit the operation to owned subscriptions, listener
                        destinations or filters. Default: False

  -h, --help            Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription list-filters

  list subscription filters

  List existing CIM indication subscriptions on the current connection

Command Options:
  -p, --paths           Show paths of filters.
  -d, --details         Show detailed instance for each filter.
  -s, --summary         If True, show only summary count of filters
  -i, --submgr-id TEXT  Optional subscription manager id that can be used to
                        separate subscriptions created by pywbem. Default:
                        pywbemcliSubMgr

  -h, --help            Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription list-subscriptions

  list indication subscriptions.

  List the existing CIM indication subscriptions on the defined connection

Command Options:
  -d, --detail          Show details of the subscriptions.
  -s, --summary         If True, show only summary count of subscriptions,
                        filters, and destinations.

  -i, --submgr-id TEXT  Optional subscription manager id that can be used to
                        separate subscriptions created by pywbem. Default:
                        pywbemcliSubMgr

  -h, --help            Show this help message.
Usage: pywbemcli [COMMAND-OPTIONS] subscription test_indication

  Generate test indications.  This command executes a WBEM server dependent
  set of requests to generate test indications.

  Today, it only works for the OpenPegasus WBEM server and only if specific
  classes are implemented in a provider on the server.  It creates a
  subscription and sends a method that tells the server to generate  a number
  of test indications.

  If the owned option is set, the command will issue a message and wait for
  the user and then complete which will close the subscription.  If not owned,
  the user must later delete the subscription.

Command Options:
  -d, --destination host name     Listener destination host for indications.
                                  Default:localhost

  --count COUNT                   Number of test indications to be requested.
                                  Default:10.

  -d, --listener-port listener_port
                                  Listener destination port to receive
                                  indications

  -i, --submgr-id TEXT            Optional subscription manager id that can be
                                  used to separate subscriptions created by
                                  pywbem. Default: pywbemcliSubMgr

  --owned                         Limit the operation to owned subscriptions,
                                  listener destinations or filters. Default:
                                  False

  -h, --help                      Show this help message.

KSchopmeyer:
examples of usage (Note that there are no removes in these examples)

```


pywbemcli -m tests/unit/pywbemcli/testmock/wbemserver_mock_class.py

# The following hides the display of the compile
pywbemcli> class enumerate --no --di

pywbemcli> -v subscription add-destinations -l http://blah:5000 -l https://blah:5001
Added destinations instances:
instance of CIM_ListenerDestinationCIMXML {
   CreationClassName = "CIM_ListenerDestinationCIMXML";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   PersistenceType = 3;
   Name =
      "pywbemdestination:owned:leonard:pywbemcliSubMgr:d4f21d32-65e7-4911-a37b-"
      "391eca524550";
   Destination = "http://blah:5000";
   Protocol = 2;
};

instance of CIM_ListenerDestinationCIMXML {
   CreationClassName = "CIM_ListenerDestinationCIMXML";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   PersistenceType = 3;
   Name =
      "pywbemdestination:owned:leonard:pywbemcliSubMgr:64713054-bd4e-4dcb-8d58-"
      "fde6dab57910";
   Destination = "https://blah:5001";
   Protocol = 2;
};

#
# Adding a second destination with same urls.  This fails since duplicates
# previous owned destinations
#
pywbemcli> -v subscription add-destinations -l http://blah:5000 -l https://blah:5001
-Owned URL http://blah:5000 already exists. Ignored
Owned URL https://blah:5001 already exists. Ignored
Error: Add destinations aborted. No destinations to add.

#
## Adding a filter
#

 -v subscription add-filter -q "SELECT * from CIM_Indication" --filter-id filter1
Added filter instance:
instance of CIM_IndicationFilter {
   CreationClassName = "CIM_IndicationFilter";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   Name =
      "pywbemfilter:owned:leonard:pywbemcliSubMgr:filter1:65a15096-45e1-458a-b9"
      "82-7521c6f65630";
   SourceNamespaces = { "root/cimv2" };
   Query = "SELECT * from CIM_Indication";
   QueryLanguage = "WQL";
   IndividualSubscriptionSupported = true;
};

pywbemcli> subscription list-filters
Instances: CIM_IndicationFilter
+---------------------------------+------------------------------+-----------------+--------------------+
| Name                            | Query                        | QueryLanguage   | SourceNamespaces   |
|---------------------------------+------------------------------+-----------------+--------------------|
| pywbemfilter:owned:leonard:pywb | SELECT * from CIM_Indication | WQL             | root/cimv2         |
| emcliSubMgr:filter1:65a15096-45 |                              |                 |                    |
| e1-458a-b982-7521c6f65630       |                              |                 |                    |
+---------------------------------+------------------------------+-----------------+--------------------+

#
# NOTE: Adding the same filter a second time results in more filters because we dictated that we did
# not want the filter_id to be required to be unique.
#

pywbemcli> -v subscription add-filter -q "SELECT * from CIM_Indication" --filter-id filter1
|RAdded filter instance:
instance of CIM_IndicationFilter {
   CreationClassName = "CIM_IndicationFilter";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   Name =
      "pywbemfilter:owned:leonard:pywbemcliSubMgr:filter1:b848760a-6af1-4c8c-a3"
      "8d-4ea702662976";
   SourceNamespaces = { "root/cimv2" };
   Query = "SELECT * from CIM_Indication";
   QueryLanguage = "WQL";
   IndividualSubscriptionSupported = true;
};

pywbemcli> subscription list-filters
Instances: CIM_IndicationFilter
+---------------------------------+------------------------------+-----------------+--------------------+
| Name                            | Query                        | QueryLanguage   | SourceNamespaces   |
|---------------------------------+------------------------------+-----------------+--------------------|
| pywbemfilter:owned:leonard:pywb | SELECT * from CIM_Indication | WQL             | root/cimv2         |
| emcliSubMgr:filter1:65a15096-45 |                              |                 |                    |
| e1-458a-b982-7521c6f65630       |                              |                 |                    |
| pywbemfilter:owned:leonard:pywb | SELECT * from CIM_Indication | WQL             | root/cimv2         |
| emcliSubMgr:filter1:b848760a-6a |                              |                 |                    |
| f1-4c8c-a38d-4ea702662976       |                              |                 |                    |
+---------------------------------+------------------------------+-----------------+--------------------+


# Add subscription based on existing filter and destinations

pywbemcli> -v subscription add-subscription --destination https://blah:5001 --filter-id filter1

1 subscriptions added

#
#  There is a simplified list command "subscription list" that thus shows counts of what we understand
# to be the instances of each type.
#
pywbemcli> subscription list
WBEM server indication instances for server_id http://FakedUrl:5988
+-------------------------------+---------+-------------+-------+
| CIM_class                     |   owned |   permanent |   all |
|-------------------------------+---------+-------------+-------|
| CIM_IndicationSubscription    |       1 |           0 |     1 |
| CIM_IndicationFilter          |       1 |           0 |     1 |
| CIM_ListenerDestinationCIMXML |       2 |           0 |     2 |
+-------------------------------+---------+-------------+-------+


pywbemcli> subscription list-destinations

Instances: CIM_ListenerDestinationCIMXML
+----------------------------------+-------------------+-------------------+-------------+
| Name                             | Destination       | PersistenceType   | Protocol    |
|----------------------------------+-------------------+-------------------+-------------|
| pywbemdestination:owned:leonard: | http://blah:5000  | 3 (Transient)     | 2 (CIM-XML) |
| pywbemcliSubMgr:526bfd97-ed3c-4d |                   |                   |             |
| 69-bfcd-db5fb55a669a             |                   |                   |             |
| pywbemdestination:owned:leonard: | https://blah:5001 | 3 (Transient)     | 2 (CIM-XML) |
| pywbemcliSubMgr:2848ca33-334b-40 |                   |                   |             |
| 09-9a8c-bb6fd404fdcb             |                   |                   |             |
+----------------------------------+-------------------+-------------------+-------------+

pywbemcli> subscription list-filters

Instances: CIM_IndicationFilter
+----------------------------------+------------------------------+-----------------+--------------------+
| Name                             | Query                        | QueryLanguage   | SourceNamespaces   |
|----------------------------------+------------------------------+-----------------+--------------------|
| pywbemfilter:owned:leonard:pywbe | SELECT * from CIM_Indication | WQL             | root/cimv2         |
| mcliSubMgr:filter1:daa1010a-b61f |                              |                 |                    |
| -48e3-8f83-55c166253a23          |                              |                 |                    |
+----------------------------------+------------------------------+-----------------+--------------------+


#
# You can also display the mof of the destinations, filters or subscriptions.
# EX: mof of filters.
#
pywbemcli> -o mof subscription list-filters

instance of CIM_IndicationFilter {
   CreationClassName = "CIM_IndicationFilter";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   Name =
      "pywbemfilter:owned:leonard:pywbemcliSubMgr:filter1:daa1010a-b61f-48e3-8f"
      "83-55c166253a23";
   SourceNamespaces = { "root/cimv2" };
   Query = "SELECT * from CIM_Indication";
   QueryLanguage = "WQL";
   IndividualSubscriptionSupported = true;
};

#
#  For the subscription it shows the mof of the association and components
#  largely because the subscription mof really means very little by itself
#  other than the properties other than Filter and Handler
pywbemcli> -o mof subscription list

instance of CIM_IndicationSubscription {
   Filter =
      "/interop:CIM_IndicationFilter.CreationClassName=\"CIM_IndicationFilter\""
      ",Name=\"pywbemfilter:owned:leonard:pywbemcliSubMgr:filter1:daa1010a-b61f"
      "-48e3-8f83-55c166253a23\",SystemCreationClassName=\"CIM_ComputerSystem\""
      ",SystemName=\"MockSystem_WBEMServerTest\"";
   Handler =
      "/interop:CIM_ListenerDestinationCIMXML.CreationClassName=\"CIM_ListenerD"
      "estinationCIMXML\",Name=\"pywbemdestination:owned:leonard:pywbemcliSubMg"
      "r:2848ca33-334b-4009-9a8c-bb6fd404fdcb\",SystemCreationClassName=\"CIM_C"
      "omputerSystem\",SystemName=\"MockSystem_WBEMServerTest\"";
   SubscriptionStartTime = "20210808111034.324721-300";
   TimeOfLastStateChange = "20210808111034.324778-300";
   OnFatalErrorPolicy = 2;
   RepeatNotificationPolicy = 2;
   SubscriptionState = 2;
};

instance of CIM_IndicationFilter {
   CreationClassName = "CIM_IndicationFilter";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   Name =
      "pywbemfilter:owned:leonard:pywbemcliSubMgr:filter1:daa1010a-b61f-48e3-8f"
      "83-55c166253a23";
   SourceNamespaces = { "root/cimv2" };
   Query = "SELECT * from CIM_Indication";
   QueryLanguage = "WQL";
   IndividualSubscriptionSupported = true;
};

instance of CIM_ListenerDestinationCIMXML {
   CreationClassName = "CIM_ListenerDestinationCIMXML";
   SystemCreationClassName = "CIM_ComputerSystem";
   SystemName = "MockSystem_WBEMServerTest";
   PersistenceType = 3;
   Name =
      "pywbemdestination:owned:leonard:pywbemcliSubMgr:2848ca33-334b-4009-9a8c-"
      "bb6fd404fdcb";
   Destination = "https://blah:5001";
   Protocol = 2;
};

#
#  Listing subscriptions table gets key properties from filter and handler
#  and presents them as a single table.
#

Subscriptions
+------------------+------------------------------+------------+---------------------------+
| Destination      | Query                        | Query      | StartTime                 |
|                  |                              | language   |                           |
|------------------+------------------------------+------------+---------------------------|
| http://blah:5000 | SELECT * from CIM_Indication | WQL        | 20210808122522.519211-300 |
+------------------+------------------------------+------------+---------------------------+

andy-maier:
Please rebase to master since the two dependent PRs have been merged.

andy-maier (approved):

KSchopmeyer:
Andy proposed that we not let the server resolve dupliactes of add-destination and add-filter --permanent but that we test and not allow any duplication of the Name property on these commands with permanent filters and destinations.

DONE.

andy-maier (approved):

PR #992, opened 2021-06-13
Fixes issue in class invokemethod with cmd namespace option (-n)

KSchopmeyer:
The option was being ignored with the command

    pywbemcli class invoke method ... -n <namespacename>

1 Modifies process method to assure only  CIMClassName and
CIMInstanceName on input

2. Add code to set namespace in objectname if the option exists.

3. Adds test for this option on invokemethod

4. Adds tests of other class commands with multiple namespaces.

This pr does NOT add tests across namespaces for the other commands (instance, qualifier) that use the --namespace command option.  That will be a separate issue. NOTE: All the other class commands did correctly handle the cross namespace execution.

### DIFF
--- a/pywbemtools/pywbemcli/_common.py
+++ b/pywbemtools/pywbemcli/_common.py
@@ -609,7 +609,8 @@ def split_str_w_esc(astring, delimiter, escape='\\'):
     return ret
 
 
-def process_invokemethod(context, objectname, methodname, options):
+def process_invokemethod(context, objectname, methodname, namespace,
+                         parameters):
     # pylint: disable=line-too-long
     """
     Process the parameters for invokemethod at either the class or instance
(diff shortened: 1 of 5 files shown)


pywbemtools/pywbemcli/_common.py
```diff
@@ -655,14 +655,17 @@ def create_params(classname, cim_method, kv_params):
...
+    classname = objectname.classname
 
+    cmd_namespace = options['namespace']
     cim_class = conn.GetClass(
         classname,
-        namespace=options['namespace'], LocalOnly=False)
+        namespace=cmd_namespace, LocalOnly=False)
+
+    if options['namespace']:
+        objectname.namespace = cmd_namespace
```
andy-maier, line 668: This modifies the object that is passed in as an argument. That is a side effect I think we should avoid if possible, even if it is just an internal method.

In the instance-level invoke method, a get_instancename() is invoked that returns a CIMInstance object with namespace set from the options, and that is passed to process_invokemethod(), so that one does not need to inject the namespace. Maybe we can do it similarly for the class-level invoke method, i.e. pass a CIMClassName with namespace set. We would still reduce the objectname parameter like you did, to no longer accept a string.

andy-maier (changes requested):
See comment

pywbemtools/pywbemcli/_common.py
KSchopmeyer: Fixed. DONE.  Change all to avoid passing options to this method

PR #998, opened 2021-07-03
Added documentation for the pywbemlistener command

andy-maier:


docs/pywbemlistener/commands.rst
```diff
@@ -0,0 +1,240 @@
...
+See :ref:`pywbemlistener stop --help` for the exact help output of the command.
+
+
+.. _`pywbemlistener run command`:
+
+``pywbemlistener run`` command
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``pywbemlistener run`` command is running the WBEM indication listener code
+in a loop that never ends. While it is possible to start this command in the
```
KSchopmeyer, line 235: The "While ... " is confusing. Should we even mention it since the easy path is pywbemlistener start

docs/pywbemlistener/commands.rst
```diff
@@ -0,0 +1,240 @@
...
+
+This section describes each of the pywbemlistener commands including examples.
+
+The pywbemlistener commands are:
+
+* :ref:`pywbemlistener list command` - List the currently running named WBEM indication listeners.
+* :ref:`pywbemlistener show command` - Show a named WBEM indication listener.
+* :ref:`pywbemlistener start command` - Start a named WBEM indication listener in the background.
+* :ref:`pywbemlistener stop command` -  Stop a named WBEM indication listener.
+* :ref:`pywbemlistener run command` - Run as a named WBEM indication listener.
```
KSchopmeyer, line 29: How did we end up with an order where start and run are not adjacent. in this list.. I think the order should be run, start, stop

docs/pywbemlistener/commands.rst
```diff
@@ -0,0 +1,240 @@
...
+
+See :ref:`pywbemlistener stop --help` for the exact help output of the command.
+
+
+.. _`pywbemlistener run command`:
+
+``pywbemlistener run`` command
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``pywbemlistener run`` command is running the WBEM indication listener code
```
KSchopmeyer, line 234: change "is running" to "runs"

docs/pywbemlistener/generaloptions.rst
```diff
@@ -0,0 +1,78 @@
...
+
+
+.. index:: single: pywbemlistener logging
+
+.. _`pywbemlistener logging`:
+
+Logging
+"""""""
+
+Logging is performed only by the ``pywbemlistener run`` command. The
```
KSchopmeyer, line 25: Is this important?  We are discussing the internals when what is important is that the logging logs indications received and state changes of the listener.

docs/pywbemlistener/generaloptions.rst
```diff
@@ -0,0 +1,78 @@
...
+``pywbemlistener run`` command it starts as a background process. All other
+``pywbemlistener`` commands ignore the logging settings.
+
+Logging is enabled by specifying the ``-l DIR`` or ``--logdir DIR`` general
+option, or by setting the ``PYWBEMLISTENER_LOGDIR`` environment variable to
+``DIR``. The general option has precedence over the environment variable.
+``DIR`` is the path name of the directory where the log files will
+reside in.
+
+The log files are named ``pywbemlistener_NAME.log`` where ``NAME`` is the
```
KSchopmeyer, line 36: Add there is a log file created for each indication listener initiated (or something like that).

docs/pywbemlistener/index.rst
```diff
@@ -20,15 +20,15 @@ Pywbemlistener command
 
 The pywbemlistener command can run and manage WBEM listeners.
 
-Each listener is a process that executes the `pywbemlistener run`
+Each listener is a process that executes the ``pywbemlistener run``
 command to receive WBEM indications sent from a WBEM server.
 
-A listener process can be started with the `pywbemlistener start`
-command and stopped with the `pywbemlistener stop` command.
+A listener process can be started with the ``pywbemlistener start``
```
KSchopmeyer, line 26: Should we be discussing the run command here also and the fact that the difference between the run and the start is whether the indication listener is a subprocess of the initiating pywbemlistener.

KSchopmeyer (commented):
I made some comments, mostly on documentation.  Note that some comments apply to the documentation in the doc string.

docs/pywbemlistener/commands.rst
andy-maier: It is an externally visible and documented command, so we need to explain it at least to the extent that people understand what it does and why not to use it directly.

docs/pywbemlistener/commands.rst
andy-maier: run is the one we don't want users to use directly, that's why I put it at the end.

docs/pywbemlistener/commands.rst
andy-maier: DONE

docs/pywbemlistener/generaloptions.rst
andy-maier: `--logdir` is a global option, for consistency with pywbemcli. However, it is ignored on all commands but this one. That deserves some explanation.

docs/pywbemlistener/generaloptions.rst
andy-maier: DONE: Added "There is a log file created for each listener that is running."

docs/pywbemlistener/index.rst
andy-maier: I don't think so. We do not want users to use it directly, and this is just the overview here.

andy-maier:
I updated the PR and marked those comments I have resolved as proposed, as resolved.

The other ones I responded to but left them unresolved. Please ready my responses and add comments or mark them as resolved.

docs/pywbemlistener/commands.rst
KSchopmeyer: Except that on the general help it shows up as the second on the list with pywbemlistener -h

```
  list   List the currently running named WBEM indication listeners.
  run    Run as a named WBEM indication listener.
  show   Show a named WBEM indication listener.
  start  Start a named WBEM indication listener in the background.
  stop   Stop a named WBEM indication listener.

As the help, etc. works, That is the one the user would pick first.   Is it that the codes is such that we absoluteily do not want the user to execute the run or just that it creates a dependent process that can only be killed by ctrl-c or something similar?  If we absolutely do not want them to use run, I think we should hide the command. 

 Click has a hidden option on the commands class that I think hides the help but allows the command to run.  Would that be a solution???

docs/pywbemlistener/generaloptions.rst
KSchopmeyer: What we are trying to say is that only the run command actually adds entries to the log.  Or more general, the log is used only to provide information on the pywbemlistener operation based on python standard loging filters (ex: INFO, DEBUG, WARNING, ERROR, etc.)

docs/pywbemlistener/commands.rst
andy-maier: I still hang on to the possibility that users have their own mechanisms for using the run command. For example, it could be turned into an OS service or a job. So I would not want to hide the run command.

The order of commands in the general help is alphabetical. Can that be influenced?

docs/pywbemlistener/generaloptions.rst
andy-maier: Yes, only the run command adds log entries.
No, the standard Python severities are not used, and not even the Python logging mechanism is used.

docs/pywbemlistener/commands.rst
KSchopmeyer: Let me look at the ordering.  At one point we had set the order as the order in the _cmd)....py file but I think we created a subclass that sorted the order in our click customizations.

docs/pywbemlistener/index.rst
KSchopmeyer: I still think somewhere we should state these differences and why the run command is public.  As it is, the user is faced with run and start and the first question is which do I use.

docs/pywbemlistener/commands.rst
andy-maier: DONE
* Used move_to_end on click.group() to move the run command to the end of the help list.
* Added the following text to the help text for the run command: 
  "Note: The `pywbemlistener start` command should be used to start listeners,
    and it starts a `pywbemlistener run` command as a background process.
    Use the `pywbemlistener run` command only when you need to have control
    over how exactly the process runs in the background."

PR #1004, opened 2021-07-05
Added 'test' command to listener

andy-maier:
Ready to be reviewed and tested!

**TEST Instructions:**

1. Check out this branch and reinstall pywbemtools and pywbem to pick up the required pywbem changes:
    ```
    git checkout andy/listener-send
    pip uninstall -y pywbem
    make clobber install
    ```

2. Start a listener with output to an indication log file:
    ```
    pywbemlistener start lis1 -s http --indi-file lis1.log
    ```

3. Send a test indication to that listener:
    ```
    pywbemlistener test lis1
    ```

4. Look at the indication log file:
    ```
    cat lis1.log
    ```

DONE:
* Added tests
* Add `ExportIndication()` to `pywbem.WBEMConnection` (done in PR https://github.com/pywbem/pywbem/pull/2734)
* Add documentation (this PR)
* Discussions (done):
  * the approach in general -> go
  * targeting only pywbemlistener based listeners -> yes
  * using a fixed alert indication -> yes
  * Should we rename the command to 'test'? -> yes

Note that this command depends on support in pywbem for the ExportIndication export message. See  https://github.com/pywbem/pywbem/issues/2729 -> Has been implemented in PR https://github.com/pywbem/pywbem/pull/2734 -> Merged.

### DIFF
--- a/docs/pywbemlistener/cmdshelp.rst
+++ b/docs/pywbemlistener/cmdshelp.rst
@@ -62,6 +62,7 @@ Help text for ``pywbemlistener``:
       show   Show a named WBEM indication listener.
       start  Start a named WBEM indication listener in the background.
       stop   Stop a named WBEM indication listener.
+      test   Send a test indication to a named WBEM indication listener.
       run    Run as a named WBEM indication listener.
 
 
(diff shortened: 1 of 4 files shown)


andy-maier:
We discussed this idea and agree. Maybe the command should be called "test" instead of "send".

KSchopmeyer:
We agreed to use test as the command in place of send.

andy-maier:
**DISCUSSION:** Is further test needed?

andy-maier:
Karl will do a test with it.

KSchopmeyer (commented):
I agree with this and worked with it for a while.  I have one suggestion,  Add a second parameter, 'repeat' or 'count' that would allow it to send multiple indications to the listener where the default was 1, (without this optional parameter) it would only send a single indication.  That would allow testing load of our listener.

Andy: The scope of this test is to ensure the indication gets to the listener, and that can be ensured with a single indication. Also, this is a slippery slope: What about support for process indications? Or for changing the indication properties?

**DISCUSSION**

Andy: We discussed this and agreed to only add support for a repetition count, but not more.

KSchopmeyer (commented):
In retrospect, I have a second comment.  I was wondering about using this across container and host to test indications and realized that by allowing only the listener name, the test can only send locally.  I know that we agreed that this was to test our listener but would it be logical to extend it to alternate of port and schema options for that case (testing indication path from container to host or host to container? This is probably more discussion than proposal at this point.

andy-maier:
I added a `-c`/ `--count` option for the n umber of indications to send.

KSchopmeyer (approved):

PR #1007, opened 2021-07-05
Pywbemlistener: Improved output messaging; accomodated grand child in Windows

andy-maier:
See commit message.

### DIFF
--- a/pywbemtools/pywbemlistener/_cmd_listener.py
+++ b/pywbemtools/pywbemlistener/_cmd_listener.py
@@ -140,6 +140,14 @@ LISTEN_OPTIONS = [
 ]
 
 
+def print_out(line):
+    """
+    Print a line to stdout, and flush stdout.
+    """
+    print(line)
+    sys.stdout.flush()
+
+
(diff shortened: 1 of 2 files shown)


pywbemtools/pywbemlistener/_cmd_listener.py
```diff
@@ -278,6 +294,11 @@ def created(self):
 
 @cli.command('run', cls=PywbemtoolsCommand, options_metavar=CMD_OPTS_TXT)
 @click.argument('name', type=str, metavar='NAME', required=False)
+@click.option('--start-pid', type=str, metavar='PID',
```
KSchopmeyer, line 297: This one may require more explanation here or in doc as to why it is required.

andy-maier: DONE
Added this sentence:
"Note: The --start-pid option is needed because on Windows, the
    `pywbemlistener run` command is not the direct child process of the
    `pywbemlistener start` command starting it."

PR #1008, opened 2021-07-07
[WIP] Fixes for listener on Windows

andy-maier:


docs/pywbemlistener/commands.rst
```diff
@@ -220,6 +220,11 @@ Example:
     $ pywbemlistener stop lis1
     Shut down listener lis1 running at http://localhost:25989
 
+On Windows, the stop of the listener process is immediate, without giving the
```
KSchopmeyer, line 223: grammar change.  On windows the listener process is stopped immediately without giving

andy-maier: DONE, also in next sentence.

PR #1029, opened 2021-07-19
Ks/#1023/logging change interactive

KSchopmeyer:
Adds logic to enable, disable and modify the log configuration for each interactive --log option

Note: Added a verbose msg output showing the --log change.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -43,6 +43,16 @@ Released: not yet
 
 * Fix issue where the general help for '--log' was unclear. (see issue #1025)
 
+* Fixed an error that resulted in exception traceback when instance commands
+  used the instance wildcard (e.g. 'CIM_ManagedSystemElement.?') and the
+  enumerate instances operation failed for some reason. (issue #963)
+
+* Fix issue with --log general option where the log was left enabled when the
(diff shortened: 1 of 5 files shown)


andy-maier:
See the deepcopy errors on py27 and py34. There is your lock error again.
I wonder why there is a `thread.lock` object on the `pywbem_server` object. Also, is the `_io.TextIOWrapper` object caused by logging?
Finally, I find it strange that `deepcopy()` issues error messages that use the terms "pickle" and "serialize".

KSchopmeyer:
1. Apparently deepcopy actually uses pickle methods which explains the pickle messages.   

2.  The Threadlock comes from the logger and only on old python versions so I can assume that there was something in these vesions of the logger where locks were set into the copiable data but that was moved, modified, etc. somewhere in python logger because of this type of issue.

andy-maier (approved):
Please rebase and resolve the merge conflict in changes.rst. Otherwise, approved.

KSchopmeyer:
Issues from  in merge resolved.

andy-maier:
Needs It still contains a commit named after a PR that is already merged. Please verify whether you want the change from that commit, or not.

PR #1036, opened 2021-07-26
Remove restrictions on additional general options with --name general option

KSchopmeyer:
NOTE: This pr has NOT be built on top of the ones for issue 1023 and 1025

Originally pywbemcl restricted the use of general options such as --user, etc. when the cmd line --name option was used.  This pr removes this restriction both for cmd line and interactive mode.  See issue # 1034.

Additionally, it  allowied general options that apply only to the WBEMConnection to be included on the command line with the --mock-server option.  This pr generates an exception for those cases both for cmd line mode and interactive mode. See issue: 1035

Most of the changes were in pywbemcli and we are still at 92% coverage for that code.


docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -225,6 +226,25 @@ user):
 
     pywbemcli> !EXTERNAL-COMMAND
 
+The general options may be included on the interactive command line to
+modify the general options entered in the initial command line for pywbemcli.
```
andy-maier, line 230: The use of "modify" may be interpreted such that the changes persist. That is not the case however, I suggest the use of "override" instead.

docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -225,6 +226,25 @@ user):
...
+Thus, a user can define a server on the command line and modify elements
+of that definition with commands in the interactive mode.
```
andy-maier, line 232: Again, change "modify" to "override".

docs/pywbemcli/cmdlineinterface.rst
```diff
@@ -225,6 +226,25 @@ user):
...
+c
```
andy-maier, line 233: Superflous "c"

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -255,14 +255,87 @@ def _validate_no_server_option_conficts(server, connection_name, mock_server):
...
-                            certfile, keyfile, resolved_ca_certs,
-                            resolved_mock_server, connections_repo,
-                            conditional_options,
-                            verbose):
+                            certfile, keyfile, ca_certs,
+                            resolved_mock_server,
+                            connections_repo,
+                            verbose,
+                            default_namespace, use_pull, pull_max_cnt,
+                            timeout, verify):
```
andy-maier, line 338: This layout is a pain when debugging. I suggest to split already the first parameter into a new line indented by 8, and then fill the lines to the end to minimize the number of lines.

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -701,11 +759,20 @@ def cli(ctx, server, connection_name, default_namespace, user, password,
...
+                                                ca_certs,
                                                 resolved_mock_server,
                                                 connections_repo,
-                                                conditional_options,
-                                                verbose)
+                                                verbose,
+                                                # used for test in function
+                                                default_namespace,
+                                                use_pull, pull_max_cnt, timeout,
+                                                verify)
```
andy-maier, line 769: Similarly, this is a pain when debugging.

pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -255,14 +255,87 @@ def _validate_no_server_option_conficts(server, connection_name, mock_server):
...
+        opts = [user, password, timeout, verify is not None, certfile,
+                keyfile, ca_certs]
+        if any(opts):
+            opt_names = ["--user", "--password", "--timeout",
+                         "--verify", "--certfile", "--keyfile",
+                         "--ca_certs"]
+            errors = [opt_names[c] for c, i in enumerate(opts) if i]
+            raise click.ClickException(
+                "General option(s): ({0}) not allowed with --mock-server "
+                "general option'".
```
andy-maier, line 322: First, superflous trailing single quote in exception message.

Second, in case of specifying a named connection, the mock_server attribute of PywbemServer is set from the connections repo, but it was not specified in the command line, and thus the error message is misleading.

You can try that out with:
```
$ pywbemcli -n simple_interop -u myuser
Error: General option(s): (--user) not allowed with --mock-server general option'
```

To address that, I suggest changing "not allowed with --mock-server general option'" to "not allowed when using a mock server".

andy-maier (changes requested):
See comments

docs/pywbemcli/cmdlineinterface.rst
KSchopmeyer: Good catch DONE

docs/pywbemcli/cmdlineinterface.rst
KSchopmeyer: DONE

docs/pywbemcli/cmdlineinterface.rst
KSchopmeyer: DONE

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: Agree completely but it is very good for establishing that all parameters are in place.  Changed to your suggestion now that we know all are in place.

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: DONE

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: TKS. Completely missed that one.   DONE

Issue #1039, opened 2021-08-02
Connections file default connection name hard setter to find


KSchopmeyer:
Found that even I do not use the ability to create a default connection because a) it is only part of one command b) that is the connection select command which is probably used only occasionaly and c) no real documentation.

I propose:

1. Add --set-default to the connection save command.
2. Add a new command "connection set-default name" that will allow the user to just set the default connection to one of the connections defined in the connections file and addition will allow the default connection to be cleared if the user does not want a default connection defined.
3. Change option name to set-default from --default to clarify what it does

PR #1040, opened 2021-08-03
fixes issue #1039 extend set default connection command options

KSchopmeyer:
See commit  and issue #1039 for details:

Extends capability to set default connection name

1. adds --set-default to connection save command
2. adds command connection set-default
3. adds tests
4. adds documentation on the default connection name.

I did this one the other night when I realized that even I had forgotten how to use the default connection name and could not remember it.

Adding to the save is logical because that is often when you might want to set the connection you saves to default.  connection set-default command makes the whole idea public and also allows the user to clear the default if that is ever needed.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -17,6 +17,10 @@ Released: not yet
 * The PYWBEMCLI_TERMWIDTH environment variable was renamed to
   PYWBEMTOOLS_TERMWIDTH since it is common to all pywbemtools commands.
 
+* Changed option --default on command ``connection select`` to ``set-default``.
+  to be compatible with other commands that touch the default connection
+  definition.
+
 **Deprecations:**
 
 **Bug fixes:**
@@ -62,6 +66,12 @@ Released: not yet
 * Generate exception when general options such as --user, --password, etc.
   that apply only to the server are used with the --mock-server general
   option. (see issue #1035)
+
+* Extend the capability to set the default connection in a connections file to
+  the connection save command and a specific command that will set or clear the
+  default.  Since the ability to set the default connection was only an
+  option in the connection select command it was difficult to find.  This makes
+  the functionality more visible and more usable.
 
 
 **Cleanup:**
--- a/pywbemtools/pywbemcli/_cmd_connection.py
+++ b/pywbemtools/pywbemcli/_cmd_connection.py
@@ -37,7 +37,8 @@ from .config import PYWBEMCLI_SERVER_ENVVAR, \
     PYWBEMCLI_CERTFILE_ENVVAR, PYWBEMCLI_KEYFILE_ENVVAR, \
     PYWBEMCLI_CA_CERTS_ENVVAR, PYWBEMCLI_TIMEOUT_ENVVAR, \
     PYWBEMCLI_USE_PULL_ENVVAR, PYWBEMCLI_CONNECTIONS_FILE_ENVVAR
-from ._common import pick_one_from_list, pywbem_error_exception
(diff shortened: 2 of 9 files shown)


docs/changes.rst
```diff
@@ -17,6 +17,10 @@ Released: not yet
 * The PYWBEMCLI_TERMWIDTH environment variable was renamed to
```
andy-maier, line 17: The PR contains two commits, the description of one of which does not match the PR at all. however, the overall PR content is ok, so I think what is needed is a rebase to master, squashing the two commits into one, and cleaning up the resulting commit message.

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -291,6 +297,34 @@ def connection_list(context, **options):
...
+              help=u'Prompt user to verify change before changing the '
+              'default connection definition).')
+@add_options(help_option)
+@click.pass_obj
+def connection_set_default(context, name, **options):
+    """
+    Set a connection as the default connection.
+
+    Sets either the connection defined in the name argument as the default
+    current connection definition or, if there is no argument on the command
```
andy-maier, line 315: This is the command line help text, and there the name argument is called "NAME".

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -291,6 +297,34 @@ def connection_list(context, **options):
...
+    it sets the current connection (if there is one) as the default
+    connection.
```
andy-maier, line 317: If NAME is optional, that should be represented in the command line help, which currently displays:
```
$ pywbemcli connection set-default --help
Usage: pywbemcli [GENERAL-OPTIONS] connection set-default NAME [COMMAND-OPTIONS]
```

pywbemtools/pywbemcli/_cmd_connection.py
```diff
@@ -291,6 +297,34 @@ def connection_list(context, **options):
...
+
+    The character "?" may be used as the name argument to allow selecting  the
+    connection to be set as the default connection form
```
andy-maier, line 320: Typo "form"

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: Fixed DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE.  added the word  optional .. There apparently is NO way to indicate that an argument is optional so the only way to clarify that is to define the optionality in the text.

DONE

pywbemtools/pywbemcli/_cmd_connection.py
KSchopmeyer: DONE

docs/changes.rst
KSchopmeyer: WHOOPS.  I am not going to try to explain that one.  Removed.

KSchopmeyer:
Fixed the issues from your comments and completely retested.  Actually found one issue in where a if verbose was placed and modified that also.

docs/changes.rst
KSchopmeyer: Whoops.

andy-maier:
The merging of PR #1036 caused merge conflicts on this PR.
Please rebase to master and resolve the conflicts.

andy-maier (approved):

Issue #991, opened 2021-06-13
Cannot set the default namespace to "interop"


KSchopmeyer:
A command on the mock of multiple namespace with interop (SIMPLE_INTEROP_MOCK_SCRIPT.py) fails when a command that uses the general option -d to set the subcommand default namespace  to interop for that command is execute.

```
   pywbem -m tests/unit/pywbemcli/simple_interop_script.py
   pywbemcli> -d interop class enumerate --ni --do 
```
Note that the command:
```
class enumerate --ni --d  -n interop
```
works
   
**Example of the failure**
```
(pycli39) kschopmeyer@leonard:~/pywbem/pycli39/pywbemtools(ks/#990-invokemethod-ignores-namespace)$ ppp -m tests/unit/pywbemcli/simple_interop_mock_script.py 
Enter 'help' for help, <CTRL-D> or ':q' to exit pywbemcli or <CTRL-r> to search history, 
pywbemcli>
pywbemcli> -d interop class enumerate --no --di
-/home/kschopmeyer/virtualenvs/pycli39/lib/python3.9/site-packages/pywbem/_server.py:523: ToleratedServerIssueWarning: Server at http://FakedUrl:5988 has an Interop namespace 'interop', but does not return it when enumerating class 'CIM_Namespace' - adding it to the 'namespaces' property
  self._determine_namespaces()
Mock script tests/unit/pywbemcli/simple_interop_mock_script.py failed: 3 (CIM_ERR_INVALID_NAMESPACE): Namespace does not exist in CIM repository: 'root/cimv2'
Aborted!
```

andy-maier:
This behavior can also be reproduced without using interactive mode:

```
$ pywbemcli -m tests/unit/pywbemcli/simple_interop_mock_script.py -d interop class tree
-/Users/maiera/PycharmProjects/pywbem/pywbem/pywbem/_server.py:523: ToleratedServerIssueWarning: Server at http://FakedUrl:5988 has an Interop namespace 'interop', but does not return it when enumerating class 'CIM_Namespace' - adding it to the 'namespaces' property
  self._determine_namespaces()
Mock script tests/unit/pywbemcli/simple_interop_mock_script.py failed: 3 (CIM_ERR_INVALID_NAMESPACE): Namespace does not exist in CIM repository: 'root/cimv2'
Aborted!
```
The issue does not happen when settng the default namespace to `root/cimv2`:
```
[...]

PR #1042, opened 2021-08-09
Fixed model namespace in simple_interop_mock_script.py

andy-maier:
Ready for review and merge.

Note that this PR does not solve issue #991, it just makes the mock script independent of what that solution will be.

### DIFF
--- a/tests/unit/pywbemcli/simple_interop_mock_script.py
+++ b/tests/unit/pywbemcli/simple_interop_mock_script.py
@@ -14,8 +14,11 @@ import os
 import six
 import pywbem_mock
 
-# interop namespace used by this mock environment
+# Namespaces set up by this mock environment.
+# Must be consistent with the namespace pragmas in the main MOF file, because
+# the pywbem MOF compiler does not create any namespaces.
 INTEROP_NAMESPACE = 'interop'
+MODEL_NAMESPACE = 'root/cimv2'
 
 
 def register_dependents(conn, this_file_path, dependent_file_names):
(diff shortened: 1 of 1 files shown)


pywbemtools/pywbemcli/mockscripts/__init__.py
```diff
@@ -213,7 +213,7 @@ def script_error(file_path, exc):
     Return a MockScriptError exception object with the current traceback in its
     exception message, ready to be displayed.
     """
-    if isinstance(exc, pywbem.Error):
+    if isinstance(exc, pywbem.Error) and False:
```
KSchopmeyer, line 216: I do not understand this statement with the and False added.  I am missing something here.

tests/unit/pywbemcli/simple_interop_mock_script.py
```diff
@@ -14,8 +14,10 @@
 import six
 import pywbem_mock
 
-# interop namespace used by this mock environment
+# Namespaces used by this mock environment.
+# Must be consistent with namespace pragmas in main MOF file.
```
KSchopmeyer, line 18: Suggestion: We might explain a bit and also point back to the issue maybe since are stepping in some territory that can be confusing.

pywbemtools/pywbemcli/mockscripts/__init__.py
andy-maier: This is WIP. The statement temporarily shows `pywbem.Error` exceptions with traceback.

KSchopmeyer: Ahhha.  I was having real problems trying to determine if that and False had some other reason.

andy-maier: I have added an according TODO to remove it again.

tests/unit/pywbemcli/simple_interop_mock_script.py
andy-maier: DONE - changed to:
```
# Namespaces set up by this mock environment.
# Must be consistent with the namespace pragmas in the main MOF file, because
# the pywbem MOF compiler does not create any namespaces.
```

andy-maier:
Abandoning this PR in favor of PR #1114

PR #1048, opened 2021-09-03
Issue #4 subscriptions - Part2 Modify mockwbemserver to include subscription classes

KSchopmeyer:
This is another prerequisite to pr # 987

Adds classes for server subscription management to the definition
for mockwbemserver and adds the subscription providers.

Changed instance counts in test_server_cmds.py. Change the
instance counts to accomodate the subscription classes that
are added to mockwbemserver


KSchopmeyer:
Discussion: Fails one  test pywbem39 min with following:

     ModuleNotFoundError: No module named 'tomli'

andy-maier:
The tomli error is rerlated to the use of setuptools-scm by pip. It was first reported in https://github.com/pypa/setuptools_scm/issues/608 with setuptools-scm 6.1.1, and after some backs and forths, setuptools-scm 6.3.0 released yesterday seems to contain the final fix.

Bottom line: Re-run the PR.

andy-maier (approved):

andy-maier (approved):

PR #1066, opened 2021-10-13
Implement click extension to validate mutually exclusive options:

KSchopmeyer:
This provides for defining mutually exclusive options in the click option definition. While it is less complete than the corresponding click contrib click-option-group that contribution only works for python 3.6+ and the author was never interested in including earlier versions of python.

**Marked as waiting because I want to test this against work in pr #987, subscription command which has mutually exclusive options**

Note that this pr includes the change to the pywbemcli --connection-name, --mock-server, and --server options without appending the new text to the help command.

This allows the following syntax for the option definition:

```

    @option('--fiirst-option',
        cls=MutuallyExclusiveOption,
        help="first-option blah.",
        mutually_exclusive=["another-option"],
        show_mutually_exclusive=True)

Which defined an option first_option option that is mutually exclusive with another-option.  If the two are included in the
same command the error:
```
Error: Conflicting options: 'first-option' is mutually exclusive with options: (--another-option) will be generated.

If show_mutually_exclusive is  True in the definition the text \:

```
This option is mutually exclusive with "options: (--another-option) is appended to the help.

If show_mutually_exclusive is not included as a parameter or is False, nothing is appended to the help.

There are no special tests added for this change the same as other click extensions.  The extensions are tested with the pywbemcli code itself.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -102,12 +102,22 @@ Released: not yet
 * Enhanced test matrix for push-driven runs on GitHub Actions to add
   Python 3.5 on macOS, and removing Python 3.5 minimum on Windows.
 
+<<<<<<< HEAD
 * Implement command group subscription that manages the creation, viewing and
   removal of indication subscription on WBEM servers. This creates a new command
   group 'subscription' and new commands for adding, removing, and displaying
   (list) indication destination, filter, and subscription instances on target
   WBEM servers. It includes the code for the new commands, a set of tests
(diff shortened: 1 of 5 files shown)


andy-maier (approved):

andy-maier (changes requested):
Could you please squash the two commits into one? Otherwise, approved.

DONE, squashed.

andy-maier (approved):

Issue #1070, opened 2021-10-28
builddoc fails with python 2.7


KSchopmeyer:
Builddoc fails when building with Python 2.7 with the following information using both the pr #987 and the master branch.

```
akefile: Creating docs/pywbemlistener/cmdshelp.rst for documentation
PYWBEMTOOLS_TERMWIDTH=120 python -u tools/click_help_capture.py pywbemlistener >docs/pywbemlistener/cmdshelp.rst
Done: Created help command info for cmds: docs/pywbemlistener/cmdshelp.rst
makefile: Creating the documentation as HTML pages
rm -f build_doc/html/index.html
sphinx-build -b html -v -d build_doc/doctrees -c docs -D latex_elements.papersize=a4 docs build_doc/html
Running Sphinx v1.8.5
WARNING: Python 2 is deprecated with sphinx_rtd_theme, update to Python 3
WARNING: Sphinx 1.x is deprecated with sphinx_rtd_theme, update to Sphinx 2.x or greater
WARNING: 'html4_writer' is deprecated with sphinx_rtd_theme
making output directory...
loading intersphinx inventory from https://docs.python.org/2/objects.inv...
loading intersphinx inventory from https://docs.python.org/2/objects.inv...
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://pywbem.readthedocs.io/en/stable/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 17 source files that are out of date
updating environment: 17 added, 0 changed, 0 removed
reading sources... [  5%] appendix

Traceback (most recent call last):
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/sphinx/builders/__init__.py", line 347, in build_update
    len(to_build))
  File "/home/kschopmeyer/virtualenvs/pycli2/local/lib/python2.7/site-packages/sphinx/builders/__init__.py", line 360, in build
    updated_docnames = set(self.read())
[...]

PR #1072, opened 2021-11-01
Fixes issue #1070, Sphinx fails python 2.7

KSchopmeyer:
Fixes an issue where Sphinx fails with python < 3.5 because of an update
to docutils to 0.18. Limited docutils for python 3.5.

We also set some of the Sphinx limits from pywbem dev_requirements.txt.

### DIFF
--- a/minimum-constraints.txt
+++ b/minimum-constraints.txt
@@ -97,7 +97,8 @@ dparse>=0.5.1; python_version >= '3.5'
 tox==2.0.0
 
 # Sphinx (no imports, invoked via sphinx-build script):
-Sphinx==1.7.6
+Sphinx==1.7.6; python_version <= '3.4'
+Sphinx==3.5.4; python_version >= '3.5'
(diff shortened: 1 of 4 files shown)


minimum-constraints.txt
```diff
@@ -97,7 +97,8 @@ dparse>=0.5.1; python_version >= '3.5'
 tox==2.0.0
 
 # Sphinx (no imports, invoked via sphinx-build script):
-Sphinx==1.7.6
+Sphinx==1.7.6; python_version <= '3.4'
+Sphinx>= 3.5.4; python_version >= '3.5'
```
andy-maier, line 101: This should compare with == not with >=

andy-maier (changes requested):
see comment

minimum-constraints.txt
KSchopmeyer: Fixed. DONE

andy-maier (approved):

PR #1074, opened 2021-11-04
Safety checks; Made constraints more consistent; Relaxed wrapt pinning

KSchopmeyer:
Modified Makefile to test for security issues against list in the Makefile.  This duplicates the code in pywbem except with a slightly different list. So now pywbemcli make will fail  if safety test fails.

Also made version mods to pylint, typed-ast, and astroid to pass tests and match pywbem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -321,6 +321,82 @@ else
   cwd := $(shell pwd)
 endif
 
+# Issues reported by safety command that are ignored.
+# Package upgrade strategy due to reported safety issues:
+# - For packages that are direct or indirect runtime requirements, upgrade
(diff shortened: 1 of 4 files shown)


andy-maier (approved):
Approved, except for the rebase that is needed once the Sphinx PR is merged.

KSchopmeyer:
Marked review needed only to note that update/rebase done.

andy-maier (changes requested):
The first commit "Relaxed wrapt pinning to just py27 on windows " has a misleading title, since it also adds security ignores. It seems that the two commits should be squashed into a single one.

andy-maier (approved):

Issue #1077, opened 2021-11-13
Fix issue with pytest and more-itertools


KSchopmeyer:
See pywbem issue https://github.com/pywbem/pywbem/issues/2796 for a description of the problem.

The fix is to limit more-itertools 

  # version 8.11 requires python 3.6, See issue #1077
  more-itertools<=8.10.0; python_version <= '3.5'

PR #1078, opened 2021-11-13
Fix issue with dependency and more-itertools, issue #1077

KSchopmeyer:
Based on PR #1072- Sphinx issue

Modify dev-requirements.txt to limit version of more-itertools for
python >= 3.5. Eliminates bug where new release of more-tools did did
not limit python versions.

NOTE: I marked this one for rollback but remember that this is a pytest based issue for us so the question is whether we should rollback something that is our tests only

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -22,6 +22,10 @@ pluggy>=0.13.0
 colorama>=0.3.9,<0.4.0; python_version <= '3.4'
 colorama>=0.4.0; python_version >= '3.5'
 
+# more-itertools version 8.11 requires python 3.6, See issue #2796
+more-itertools>=4.0.0,!=8.11.0; python_version < '3.6'
+more-itertools>=4.0.0; python_version >= '3.6'
+
 # Virtualenv
 # Virtualenv 20.0.19 has an issue where it does not install pip on Python 3.4.
 # Virtualenv 20.0.32 has an issue where it raises AttributeError on Python 3.4.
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -22,6 +22,9 @@ pluggy>=0.13.0
 colorama>=0.3.9,<0.4.0; python_version <= '3.4'
 colorama>=0.4.0; python_version >= '3.5'
 
+# more-itertools version 8.11 requires python 3.6, See issue #2796
+more-itertools<=8.10.0; python_version <= '3.5'
```
andy-maier, line 26: Same comment as on pywbem:
- just exclude 8.11.0, but do not pin
- add 4.0.0 as a minimum.

minimum-constraints.txt
```diff
@@ -97,7 +97,8 @@ dparse>=0.5.1; python_version >= '3.5'
 tox==2.0.0
 
 # Sphinx (no imports, invoked via sphinx-build script):
-Sphinx==1.7.6
+Sphinx==1.7.6; python_version <= '3.4'
+Sphinx==3.5.4; python_version >= '3.5'
```
andy-maier, line 101: add more-itertools==4.0.0 as a minimum

andy-maier (changes requested):
See comments

dev-requirements.txt
KSchopmeyer: DONE

minimum-constraints.txt
KSchopmeyer: DONE

andy-maier (approved):

andy-maier:
I updated this PR to remove the first commit which was a different version of PR #1069.

PR #1091, opened 2021-12-03
Ks/#1067 save submgr between interactive commands

KSchopmeyer:
Extends pywbemcli to save the CmdSubscriptionManager object between subscription commands in interactive mode. 

See commit for more information


KSchopmeyer:
Close. broken commit.

PR #1092, opened 2021-12-04
Added support for Python 3.10

andy-maier:


minimum-constraints.txt
```diff
@@ -180,10 +193,9 @@ alabaster==0.7.9
 appnope==0.1.0
 args==0.1.0
 atomicwrites==1.2.1
-attrs==18.2.0
-Babel==2.3.4
+attrs==18.2.0; python_version <= '3.9'
+attrs==19.2.0; python_version >= '3.10'
```
KSchopmeyer, line 197: Why did Babel disappear from minimum-constraints.  Was Bavel==2.3.4 before

KSchopmeyer:
Reviewed. I tried to go through the changes and compare with pywbem versions.  I did not find anything that would  break it. However, we still have a lot differences between min versions.  Note that we also have a test_requirements.txt in pywbem but the same items are in dev-requirements in pywbemtools.  However, for righ now the goal is to just get this out.

andy-maier:
I needed to add more changes. Please review again.

KSchopmeyer (commented):
Made 1 comment.  Since I pratically went blind just reviewing this, not sure how your survived writing it.

andy-maier:
I cannot see any new comment on this PR. Please double check where the new comment is.

andy-maier:
I had to merge this to have the basis for another fix for today's failure in the all-test.

KSchopmeyer:
My only comment was that we need some time to get these requirements files in sync so we have the same thing on different python projects.  I have no problem basing it on your cookie cutter but we should try have the same structure for the requirements on a number of projects since in general, the changes are the same across projects.   It could be one minimum.. and one dev_requirements or even one per major group of test/doc requirements, what ever is simpler to keep in sync.

PR #1095, opened 2021-12-04
Fixes #1067 - cache Subscription info between commands.

### LINKED ISSUE #1067: Extend Subscription commands to keep SubscriptionManager in interactive mode
Retain SubscriptionManager object between commands in interactive mode.

This may be simple since the SubscriptionManager is capable of handling multiple wbem server within a single instance by adding and removing server_id entities.

The simplest is simply to change the code to:

a. Create a variable to hold the SubscriptionManager object in the context.
b. For each command use the existing context if it exists and add a new server if the server for this command does not exist in the context.
c. Consider possibly removing other server_ids when a new one is added but that is probably excessive at least for now.

ISSUES:

Since we will be keeping the subscription manager status between commands, there is one issue, keeping the local list of all  subscriptions up-to-date with the server.

KSchopmeyer:
Extends pywbemcli to cache subscription information between commands in
interactive mode.  While the permanent subscription information is also
cached, it is not used since pywbem enumerates subscriptions each time a
request for permanent instances is executed.

add cache of class CmdSubscriptionManager instance to PywbemServer
class and recovery when a command is called.  The instance is cached in
PywbemServer object so that there is a set of cached subscription data
for each wbem server created.

Since we do not cache the PywbemServer objects if the server is changed
in interactive mode, the subscription cache only exists where the server
definition is not changed.

This also adds documentation and corrects some documentation grammar and
formatting.

No significant to the functionality of the subscription command group

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_subscription.py
+++ b/pywbemtools/pywbemcli/_cmd_subscription.py
@@ -205,10 +205,13 @@ def subscription_add_destination(context, identity, **options):
 @click.option('--source-namespaces', type=str, metavar='TEXT',
               required=False, default=None,
               multiple=True,
-              help=u"The namespace(s) for which the query is defined. If not "
-                   u"defined, the default namespace of the server is used. "
-                   u"Define multiple namespaces by using option multiple times "
-                   u"or comma-separating namespaces names.")
+              help=u"The namespace(s) for which the query is defined. Multiple "
+                   u"values may be defined with a single comma-separated "
+                   u"string of namespaces or multiple options. If defined the "
+                   u"namespaces will be inserted into the SourceNamespaces "
+                   u"property. Otherwise the property will not be created and "
+                   u"the WBEM server typically use the Interop namespace for "
+                   u"the indication filter.")
 @add_options(ownedadd_flag_option)
 @add_options(help_option)
 @click.pass_obj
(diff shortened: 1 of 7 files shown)


pywbemtools/pywbemcli/_cmd_subscription.py
```diff
@@ -205,10 +205,13 @@ def subscription_add_destination(context, identity, **options):
...
-                   u"defined, the default namespace of the server is used. "
-                   u"Define multiple namespaces by using option multiple times "
-                   u"or comma-separating namespaces names.")
+              help=u"The namespace(s) for which the query is defined. Multiple "
+                   u"values may be defined with a single comma-separated "
+                   u"string of namespaces or multiple options. If defined the "
+                   u"namespaces will be inserted into the SourceNamespaces "
+                   u"property. Otherwise the property will not be created and "
+                   u"the WBEM server should use the interop namespace for the "
+                   u"indication filter.")
```
andy-maier, line 214: The sentence "the WBEM server should use ..." is a requirement on the WBEM server. However, we are on the client side here. So at best, we can state whether WBEM servers typically use the Interop namespace, or which standard requires them to use the Interop namespace.

pywbemtools/pywbemcli/_cmd_subscription.py
```diff
@@ -205,10 +205,13 @@ def subscription_add_destination(context, identity, **options):
...
-              help=u"The namespace(s) for which the query is defined. If not "
-                   u"defined, the default namespace of the server is used. "
-                   u"Define multiple namespaces by using option multiple times "
-                   u"or comma-separating namespaces names.")
+              help=u"The namespace(s) for which the query is defined. Multiple "
+                   u"values may be defined with a single comma-separated "
+                   u"string of namespaces or multiple options. If defined the "
+                   u"namespaces will be inserted into the SourceNamespaces "
+                   u"property. Otherwise the property will not be created and "
+                   u"the WBEM server should use the interop namespace for the "
```
andy-maier, line 213: "interop" -> "Interop".

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/_cmd_subscription.py
KSchopmeyer: That is an interesting one.  It is the model that defines this :If NULL, the namespace of the Filter registration is assumed.".  I am going with typically.

KSchopmeyer: DONE

pywbemtools/pywbemcli/_cmd_subscription.py
KSchopmeyer: DONE

KSchopmeyer:
Changes made, multiple commits squashed and ready for review/merge

andy-maier (approved):

andy-maier:
For some reason, the issue and the PR cannot be linked. This PR fixes issue #1067

PR #1096, opened 2021-12-07
Cleanup of the structure of list action cmds

KSchopmeyer:
The subscription list commands are very similar but each was written
slightly differently.  This regularizes them.

In additon it:

1. Tests for cimobject output when --names-only set and sends message if
not that.

2. Make --names-only --detail, and --summary mutually exclusive since
they are all special modes.

3. Adds a number of tests

4. Removes some exception code for return from subscription manager that
cannot be executed because the command action functions duplicate
the test with their own exception message to clarify the problem.


andy-maier:
Rebased on master and regenerated the cmdshelp.rst file to resolve its merge conflict.

PR #1101, opened 2021-12-10
Fix for issue #1079, issue with openpegasus reference property value

KSchopmeyer:
Fixes for the problem in OpenPegasus documented in issue #1079.  Changes to use the path element of the instance to get the filter and destination paths for subscription rather than getting the value from the instance itself.  That is probably the correct solution in any case but with the OP issue this is required since the value of the Filter and Handler properties in the instances is incorrect.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_subscription.py
+++ b/pywbemtools/pywbemcli/_cmd_subscription.py
@@ -1257,7 +1257,7 @@ def pick_one_path_from_instances_list(csm, instances, pick_msg):
 
 
 def resolve_instances(csm, instances, identity, obj_type_name,
-                      select_opt, action):
+                      select_opt, cmd_str):
     """
     Resolve list of instances to a single instance or exception.
     If select_opt, ask user to select.  Otherwise, generate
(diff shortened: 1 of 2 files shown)


pywbemtools/pywbemcli/_cmd_subscription.py
```diff
@@ -1327,12 +1327,16 @@ def get_insts_for_subscription_identities(csm, destination_identity,
     found or an exception if no instances for the destination and filter found.
     The identity can be either the full name as target or only the
     id suffix of an owned element
+
+    Returns: instance of destination and instance of filter if the identies
+    match. This method can return only the
```
andy-maier, line 1332: * Typo: identies
* Second sentence incomplete.

andy-maier (changes requested):
Minor comment

andy-maier (approved):

pywbemtools/pywbemcli/_cmd_subscription.py
andy-maier: Another typo, but since this is internal, good enough.

Issue #916, opened 2021-03-01
"instance count" command fails with OpenSSL error on pegasus container


andy-maier:
When performing an "instance count" command (and the default namespace) against the pegasus container, it fails with the traceback shown below, indicating "EOF occurred in violation of protocol" during SSL handshake.

Once that has happened, all subsequent pywbemcli commands error out with that same error, even commands that succeeded before. A stop/start of the pegasus container resets the condition. I have repeated this multiple times and it seems to be reproducible.

```
$ pywbemcli -n pegasus instance count
Traceback (most recent call last):
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/connectionpool.py", line 382, in _make_request
    self._validate_conn(conn)
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/connectionpool.py", line 1010, in _validate_conn
    conn.connect()
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/connection.py", line 411, in connect
    self.sock = ssl_wrap_socket(
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/util/ssl_.py", line 428, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
  File "/Users/maiera/virtualenvs/pywbemtools38/lib/python3.8/site-packages/urllib3/util/ssl_.py", line 472, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
[...]

PR #1109, opened 2021-12-17
fix issue # 1108, correct issues in instance count

KSchopmeyer:
Fixes issue where an unitialized varaible exception occurs if there is a
CIMError in handling of instance count, the count is incomplete if there is a Error exception, and there is no way to exclude
items to be counted other than selecting namespaces.

This pr:

1. Fixes the issue where we had an uninitialized variable after an exception. issue # 1108
2. Refactors the instance count to:
    a. Always output the full set of classes defined by the namespaces and class GLOB even if an error occurred. If an error occurred, the remaining rows use the string "not scanned" in place of a count.
   b. Add Error exception so we are catching the case of server failure.  In this case, the enumerate of instances
      is terminated.
   c. Add text in rows where the enumerate instance resulted in a CIMError noting the failure code.
   f. Add test in row where enumerate caused Error exception.

3. Adds a new option that defines a list of classes to be ignored when looking for instances.

This means that the command always shows all classes that either caused Error or CIMError or returned instances.
Shows the exception and CIMError as a warning.  Notes at the end of the list if there was a server failure. Even classes
that are defined as ignored are shown in the output with the count field being "ignored"

**DISCUSSION:** Should this also close issue #916, the OpenPegasus failure on instance count.  Part of it was our error in not catching the Error exception.  Now we catch this issue and the scan shows all results.  Thus pywbemcli works as best it can given that errors are coming back from server where we want to look at everything.

My proposal in the issue is to close the issue now since we are reporting what OpenPegasus is doing and that is our  job in pywbemcli

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_instance.py
+++ b/pywbemtools/pywbemcli/_cmd_instance.py
@@ -572,6 +572,17 @@ def instance_query(context, query, **options):
 @add_options(multiple_namespaces_option)
 @click.option('-s', '--sort', is_flag=True, required=False,
               help=u'Sort by instance count. Otherwise sorted by class name.')
+@click.option('--ignore-class', type=str,
+              multiple=True,
+              metavar='CLASSNAME',
+              help=u"Class names of classes to be ignored (not counted). "
+                   u"Allows counting instances in servers where instance "
+                   u"retrieval may cause a CIMError or Error exception"
+                   u"on some classes. CIN errors on particular "
+                   u"classes are ignored. Error exceptions cause scan to stop "
+                   u"and remaining classes status shown as 'not scanned'. "
+                   u"Multiple class names are allowed (one per option or "
+                   u"comma-separated).")
 @add_options(class_filter_options)
 @add_options(help_option)
 @click.pass_obj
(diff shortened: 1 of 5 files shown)


pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -572,6 +572,17 @@ def instance_query(context, query, **options):
 @add_options(multiple_namespaces_option)
 @click.option('-s', '--sort', is_flag=True, required=False,
               help=u'Sort by instance count. Otherwise sorted by class name.')
+@click.option('--ignore-class', type=str,
+              multiple=True,
+              metavar='CLASSNAME',
+              help=u"Class names of classes to be ignored (not counted). "
+                   u"Allows counting instances in servers where instance "
+                   u"retrieval may cause a CIMError or Error exception"
+                   u"on some classes. CIN errors on particular "
```
andy-maier, line 581: typo: CIN

andy-maier (changes requested):
minor comment

pywbemtools/pywbemcli/_cmd_instance.py
andy-maier: Fixed in PR #1111

andy-maier:
I have merged the PR by accident, and have fixed the one minor comment I had in PR #1111.

andy-maier:
I agree that the PR should also close issue #916.

PR #1114, opened 2021-12-21
Fixes #991: Fixed mock script and added testcases for namespace creation

andy-maier:
For details, see commit message.

TODO:
* Pull pywbem from master branch once its PR https://github.com/pywbem/pywbem/pull/2823 is merged.
* Print verbose message when creating default namespace in mock env.

**DISCUSSION:**
* This fix did not require any changes in the namespace creation processing. Is this sufficient to address issue #991?

PR #1130, opened 2022-01-02
Fix test_namespaces to match pywbem message change

KSchopmeyer:
Exception message changed in pywbem 1.4.0 and pywbemcli tests that
message on an exception for no interop namespace.

This just changes the message text tested.


andy-maier (commented):
The message was already changed as part of PR #1114. Do you still want to change the comparison mode?

andy-maier:
We agreed to abandon this one.

PR #1136, opened 2022-01-06
enhance instance table output to conditionally show null properties

KSchopmeyer:
Modify the CIM instance table output to conditionally output  information on properties where all instances have Null properties.

This adds option --detail option to instance enumerate, references, and associators.  When this option is not set properties that no value in any instance in the properties to be displayed are not presented in the output  table unless the --detail option is included.

This is part one of the prs to fix issue 1131. We will set the next pr to the issue.

Changes one parameter in _format_instances_as_rows from keyword to positional (max_cell_width) since this method should not really have power over this parameter, it is already set.

See commit for details.

### DIFF
--- a/pywbemtools/pywbemcli/_cmd_instance.py
+++ b/pywbemtools/pywbemcli/_cmd_instance.py
@@ -160,6 +160,13 @@ help_instancename_option = [              # pylint: disable=invalid-name
                  help=u'Show help message for specifying INSTANCENAME '
                       u'including use of the --key and --namespace options.')]
 
+show_null_option = [              # pylint: disable=invalid-name
+    click.option('--show-null', 'show_null', is_flag=True, required=False,
+                 help=u'In the TABLE output formats, show properties with no '
+                      u'value (i.e. Null) in all of the instances to be '
+                      u'displayed. Otherwise only properties at least '
+                      u'one instance has a non-Null property are displayed')]
+
 
 ##########################################################################
 #
@@ -201,6 +208,7 @@ def instance_group():
 @add_options(summary_option)
 @add_options(filter_query_option)
 @add_options(filter_query_language_option)
+@add_options(show_null_option)
 @add_options(help_option)
 @click.pass_obj
 def instance_enumerate(context, classname, **options):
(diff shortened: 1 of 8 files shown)


pywbemtools/pywbemcli/_cmd_instance.py
```diff
@@ -160,6 +160,12 @@
                  help=u'Show help message for specifying INSTANCENAME '
                       u'including use of the --key and --namespace options.')]
 
+detail_option = [              # pylint: disable=invalid-name
+    click.option('--detail', 'detail', is_flag=True, required=False,
+                 help=u'In the TABLE output formats, show properties with no '
+                      u'value in any instances to be displayed. Otherwise only '
+                      u'properties value at least one instance are displayed')]
+
```
andy-maier, line 168: I struggle with the option name `--detail`. It does not really show more details about the items it shows, it shows more items (properties). I think `--include-null-properties` or `--show-null-properties` would appropriately describe what it does, but it is a bit long. How about `--show-null`?

pywbemtools/pywbemcli/_display_cimobjects.py
```diff
@@ -105,14 +110,21 @@ def display_cim_objects(context, cim_objects, output_format, summary=False,
         If False, strings are not encased by quote marks in the table view for
         instance displays. The default is True so that strings are encased in
         quotes in all views.
+
+      ignore_empty_properties (:class:`py.bool`):
+        If True, cases where a property has None value in all of the instances
+        will not be included in table output. They will be included in CIM
+        object(MOF) output. If False, The properties that do not have a value
+        will be included in tables.  This allows table output to ignore
+        properties that do not add value to the table display
```
andy-maier, line 119: Can we rename this parameter to `ignore_null_properties`, because the notion of "empty" can also apply to array properties or string properties with an empty string.

andy-maier (changes requested):
See my minor comments

pywbemtools/pywbemcli/_cmd_instance.py
KSchopmeyer: I have no issue with this change and agree that --show-null works.  Making change now

KSchopmeyer: DONE

pywbemtools/pywbemcli/_display_cimobjects.py
KSchopmeyer: DONE

PR #1139, opened 2022-01-17
Release 1.0.0

andy-maier:
Please approve the release of pywbemtools 1.0.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,9 +9,9 @@ on:
     - # cron (in UTC): minute hour day_of_month month day_of_week
       cron: '30 01 * * SUN'
   push:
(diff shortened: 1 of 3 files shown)


KSchopmeyer (commented):
There is an issue here.  The end2end pr was still in wip mode and had only run the tests for enum table output a couple of times.  This time, the order of keys came up different in the table. One of the tests was for two keys in a defined order.  That is the one that failed.  I really should sort the keys before putting them into the table.  Here are my proposals.

1. Sort keys before displaying in table.
2. For right now simplify test so we just test for existence of values, not existence and order and do the order sort in for 1.1
3. Remove this PR.
I am preparing change to sort keys before we assemble table headers and rows.

KSchopmeyer (approved):

PR #1152, opened 2022-02-01
Ignored safety issue 43975

andy-maier:
No review needed.


andy-maier:
The only failure is the reporting to coveralls. Merging.

PR #1157, opened 2022-02-06
Fix issue # 1037 - allow --connections-file in interactive mode

KSchopmeyer:
Modify to allow use of --connections-file general option on a command in
interactive mode.  This is same behavior as other commands.  The defined
connections file is enabled only for that command that includes the
option.

### DIFF
--- a/tests/unit/pywbemcli/test_general_options.py
+++ b/tests/unit/pywbemcli/test_general_options.py
@@ -825,8 +825,8 @@ TEST_CASES = [
       'test': 'innows'},
      None, OK],
 
-    ['Verify -n option with non-existent connection file fails',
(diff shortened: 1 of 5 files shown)


tests/unit/pywbemcli/test_general_options.py
```diff
@@ -825,8 +825,8 @@ class       Command group for CIM classes.
       'test': 'innows'},
      None, OK],
 
-    ['Verify -n option with non-existent connection file fails',
-     {'general': ['-n', 'namedoesnotexist'],
+    ['Verify --namen option with non-existent connection file fails',
```
andy-maier, line 828: Typo in option `--namen`

andy-maier (approved):

PR #1159, opened 2022-02-10
Fix issue #1018, timeout issues

KSchopmeyer:
The issue documents  the fact that the time to timeout a WBEM operation may not be the same as the time specified in the --timeout general option.

Document in the help the fact that the actual timeout of commands is not
the same as the --timeout <int> general option value.

Further explain this in the  documentation for the general option --timeout.  

This does not
change any code since all the parameterization for retries and timeouts
is in pywbem.  A corresponding issue https://github.com/pywbem/pywbem/issues/2853
was filed for pywbem.

### DIFF
--- a/docs/pywbemcli/generaloptions.rst
+++ b/docs/pywbemcli/generaloptions.rst
@@ -393,12 +393,29 @@ any subsequent pywbemcli commands:
 ``--timeout`` general option
 """"""""""""""""""""""""""""
 
-The argument value of the ``--timeout``/``-t`` general option is an integer that
-defines the
-client side timeout in seconds. The pywbem client includes a timeout mechanism
-that closes a WBEM connection if there is no response to a request to the WBEM
-server in the time defined by this value. Pywbemcli defaults to a
-predefined timeout (normally 30 seconds) if this option is not defined.
+The argument value of the ``--timeout``/``-t`` general option is an integer
+that defines the client side read timeout in seconds. The pywbem client
+includes a timeout mechanism that closes a WBEM connection and terminates the
+current pywbemcli request if there is no response to a WBEM server request
+in the time defined by timeout with multiple retries. A read timeout
+occurs any time no bytes have been received on the underlying socket for
+timeout seconds.
+
+See ``pywbemcli --help`` for the actual retry count value.  Thus, the actual
+time to command failure is multiple times the value of this option. Therefore
+a request that does not receive any response data and with timeout value of
+5 would timeout in, for example ( 5 sec * 3 (request and retries)) = 15 seconds.
+
+Pywbemcli defaults to a predefined read timeout (normally 30 seconds) if this
+option is not defined.
+
+The connection functionality has a separate timeout time set by pywbem and
+set at 10 seconds also with retries. The connection timeout is not modifiable
+by pywbemcli.
+
+In general the timeout value should only be modified where there is a specific
+reason such as specific commands or servers that have very long delay before
+returning data.
 
 .. index:: triple: --verify; general options; verify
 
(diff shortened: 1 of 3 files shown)


pywbemtools/pywbemcli/pywbemcli.py
```diff
@@ -508,8 +508,11 @@ def _create_server_instance(
               metavar='INT',
               # defaulted in code
               envvar=PYWBEMCLI_TIMEOUT_ENVVAR,
-              help=u'Client-side timeout in seconds for operations with the '
-                   u'WBEM server. '
+              help=u'Client-side timeout (seconds) on data read for '
+                   u'operations with the WBEM server. This integer is the '
+                   u'timeout for a single server request. Pywbem retries '
+                   u'multiple times so the delay for read timeout '
+                   u'failure may be multiple times this value.'
```
andy-maier, line 515: Since the number of retries is not configurable, we could state it in the help text, by taking it from the constant that defines it.

docs/pywbemcli/generaloptions.rst
```diff
@@ -393,13 +393,24 @@ any subsequent pywbemcli commands:
...
+response to a request to the WBEM server in the time defined by this value and
+the retry count.
+
+The option value is the timeout for a single read request to the WBEM server. Since
+pywbem retries the command multiple times (normally 2 retries), the time to
+command failure may be multiple times this timeout value.  Pywbemcli defaults to a
 predefined timeout (normally 30 seconds) if this option is not defined.
 
+The connection functionality has a separate timeout defined by pywbem and normally
+set at 10 seconds also with retries.
```
andy-maier, line 408: I think we have constants for the connection timeout and its retries, and I suggest that we change the text to be specific about that.

andy-maier (changes requested):
See comments

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: I did not do that here because the variables are internal in pywbem._cim_operations.py.   I think we should make them more public than they are in pywbem itself but will make the the change here for now.

docs/pywbemcli/generaloptions.rst
```diff
@@ -393,13 +393,24 @@ any subsequent pywbemcli commands:
...
-that closes a WBEM connection if there is no response to a request to the WBEM
-server in the time defined by this value. Pywbemcli defaults to a
+The argument value of the ``--timeout``/``-t`` general option is an integer
+that defines the client side read timeout in seconds. The pywbem client
+includes a timeout mechanism that closes a WBEM connection if there is no
+response to a request to the WBEM server in the time defined by this value and
+the retry count.
+
+The option value is the timeout for a single read request to the WBEM server. Since
+pywbem retries the command multiple times (normally 2 retries), the time to
```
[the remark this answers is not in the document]
KSchopmeyer: I did not want to make it specific here because we cannot tie the value to the code itself.  Suggest we make specific in code and help and point to that from here.

docs/pywbemcli/generaloptions.rst
KSchopmeyer: Same comment as above to avoid specific values that are controlled by variables to which the documents do not have direct access.

docs/pywbemcli/generaloptions.rst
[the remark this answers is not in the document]
KSchopmeyer: Modified this to point back to the help for this option for an actual number

pywbemtools/pywbemcli/pywbemcli.py
KSchopmeyer: Fixed to show the value of the read timeout.  Note that we do not have anyplace to show any of the other timeout values (connect, etc.)

PR #1160, opened 2022-03-06
Change docker test server image used in tests

KSchopmeyer:
Changed to a new image created with definitions in github repository
OpenPegasus/OpenPegasus.  Note that the image is in
kschopmeyer/openpegasus as openpegasus-server, vesion 0.1.1.  This image
is significantly smaller and should be faster.

Note: We will drop use of the docker images from kporttech and will hopefully move the images from the kschopmeyer repo to an OpenPegasus repo if Docker approves of that repository as an open source project.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -30,6 +30,19 @@ Released: not yet
 * Extend use of general options in interactive mode to allow setting the
   connections-file for an interactive command. (see issue #1037)
 
+* Change DOCKER TEST_SERVER_IMAGE defined in Makefile to use one created from
+  OpenPegasus toolset.  See github OpenPegasus/OpenPegasusDocker repository
+  for pegasus, pegasus tools, and pegasus docker build tools.  This image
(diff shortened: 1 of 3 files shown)


docs/changes.rst
```diff
@@ -27,6 +27,8 @@ Released: not yet
 * Extend use of general options in interactive mode to allow setting the
   connections-file for an interactive command. (see issue #1037)
 
+* Change DOCKER
```
andy-maier, line 30: This change log entry could be more specific.

andy-maier (changes requested):
See comment.

docs/changes.rst
KSchopmeyer: Not sure what you want for info here.  We are trying to document the build characteristics both in the container and avalable documentation but this is the first release and the only real use for this release is the pywbemtools tests today.  I added mode documentation.

andy-maier (approved):

PR #1161, opened 2022-03-07
Remove deprecated methods server namespace, server interop

KSchopmeyer:
Remove the command definition decorators for:

1. server namespace
2. server interop

which were marked deprecated in pywbemtools 0.9.  We probably should
have removed them in pywbemtools 1.0 release but I overlooked them.

**DISCUSSION:** After removing the deprecated commands I looked at it again and wondered if we should really remove them.

1. They are logical concepts both in the namespace group and the server group (i.e   server introp is a logical concept if the user is looking to find the interop namespace without reading the command docs in detail and realizing that the command is in the namespace group.
2. They use the same code  (The server command handler calls the corresponding namespace command handler) so there is no variation in the code itself.

The question is whether to remove them, just leave them as deprecated, or change the documenttion to note that the server and namespace version do the same thing.


KSchopmeyer:
No problem.  Either choice was logical to me.  I approve this PR

andy-maier:
Please rebase to latest master and remove the new conflict (that emerged form the latest merge).

andy-maier (approved):

PR #1162, opened 2022-03-11
Added design for override of options

andy-maier:


design/option_override.md
```diff
@@ -0,0 +1,53 @@
...
+* "cmd -> int" - If there is a connection inherited from the command line to an interactive session, what should happen when the option in question is specified on an interactive command in that session. This applies regardless of how the connection at the command line level was crafted.
+
+The 'desired' columns describe the desired behavior:
+* 'update PS' means to update the currently used PywbemServer object from the option.
+* 'reset+update PS' means to reset the currently used PywbemServer object to default values and then to update it from the option. This is processed as the first option.
+* 'replace PS' means to update the currently used PywbemServer object from the new connection, replacing all connection attributes.
+
+The 'changes' columns describe the necessary changes to the existing implementation, in order to have the desired behavior.
+
+| type | general option name | desired conn -> opt | desired cmd -> int | changes conn -> opt | changes cmd -> int |
```
KSchopmeyer, line 20: Ahha. type is the connection type (real or mock)

KSchopmeyer:
NOTE: I am going through your tables and the code.  Originally hoped to be done for telecon today but still have some to do.
 I should be up to date later this week.

KSchopmeyer:
Worked my way through and tested, reviewed code for the various options and proposed changes.  My comments are:

1. | real | default-namespace   | update PS           | update PS          | remove rejection    |                    |
This one appears to be correct in the code.  I could not make it reject the change when I used --default-namespace either by command or by cmd.

2. user

TODO

3 --verify 


The command/conn usage is rejected IF the target is a mock environment because these options are not used in the mock environment.  If the target is a real server, the option is accepted and becomes part of the PS.

NOTE: The following options are rejected in interactive mode AND command line if the target object is a mock-server:

"--user", "--password", "--timeout", "--verify", "--certfile", "--keyfile",  "--ca_certs"

They are accepted for the current operation and only that operation if the server is not a mock-server.

Therefore the comment about "remove rejection" applies to user, verify, certifle, keyfile, ca-certs, timeout I think

**DISCUSSION:** Do you propose that we remove the rejection of these option for mock-server or have I missed something?  One alternate is to change exception to warning.

I could go either way. There is no harm adding these if mock-server but we are not telling the user that he is playing with options that are not used.


4. --pull-max-cnt.

This one has floated between being part of PS and not since the initial release. I think we both voted both ways at one time or another.  However, I agree that if we are saving --use-pull, we should be saving --pull-max-cnt, just for consistency.

**DISCUSSION:** Proposed resolved: Agree with comment "add to PS"


4. --password - There is an error here in that this one

6.

design/option_override.md
```diff
@@ -0,0 +1,53 @@
+These tables describe the desired behavior for different override situations (meant to be a proposal), and lists the required changes to the existing implementation (incomplete).
+
+Next steps:
+* Agree on desired behavior
+* Determine changes to existing implementation
+
+**Options with connection context:**
+
+Kinds of overrides:
+* "conn -> opt" - If the --name option is specified on a command, what should happen when the option in question is specified on the same command. This applies to both command line commands and interactive session commands.
```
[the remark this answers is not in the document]
andy-maier: It means the option that is discussed.

design/option_override.md
andy-maier: will do

KSchopmeyer:
Karl agreed to clarify comments

KSchopmeyer:
**DISCUSSION**

I went back over the table again and think I understand better but I have a question that I would guess applies to a number of the proposed changes (i.e. where marked remove rejection.  I am using --user as the example here.  The question:
I do not get a rejection with any use of --user with a real connection.  However, we do specifically reject
the --user option with mock connection.  What does your "remove rejection" mean?  By the way, I agree that the
pull-max-cnt should have the same characteristics as the use-pull.

```
| type | general option name | desired conn -> opt | desired cmd -> int | changes conn -> opt | changes cmd -> int |
|:---- |:------------------- |:------------------- |:------------------ |:------------------- |:------------------ |
| real | user                | update PS           | update PS          | remove rejection    |                    |
```

1. As an example the --user.  I do not see that we reject the request in
any case whene changing --user for a real destination as shown below:

    ppp -n dockerhttp --user new
    pywbemcli> connection show
    Connection status:
    name               value  (state)
    -----------------  ----------------------
    name               dockerhttp (current)
    server             http://localhost:15988
    default-namespace  root/cimv2
    user               new
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             True
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli> connection show
    Connection status:
    name               value  (state)
    -----------------  ----------------------
    name               dockerhttp (current)
    server             http://localhost:15988
    default-namespace  root/cimv2
    user               new
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             True
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli> --user 2ndnew connection show
    Connection status:
    name               value  (state)
    -----------------  ----------------------
    name               dockerhttp (current)
    server             http://localhost:15988
    default-namespace  root/cimv2
    user               2ndnew
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             True
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli> connection show
    Connection status:
    name               value  (state)
    -----------------  ----------------------
    name               dockerhttp (current)
    server             http://localhost:15988
    default-namespace  root/cimv2
    user               new
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             True
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli> -n dockerhttps --user 3rdnew connection show
    Connection status:
    name               value  (state)
    -----------------  -----------------------
    name               dockerhttps (current)
    server             https://localhost:15989
    default-namespace  root/cimv2
    user               3rdnew
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             False
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli> connection show
    Connection status:
    name               value  (state)
    -----------------  ----------------------
    name               dockerhttp (current)
    server             http://localhost:15988
    default-namespace  root/cimv2
    user               new
    password
    timeout            30
    use-pull
    pull-max-cnt       1000
    verify             True
    certfile
    keyfile
    mock-server
    ca-certs
    pywbemcli>

However we always reject the use of the --user option with a mock
server destination as shown below:

    (pycli39) kschopmeyer@leonard:~/pywbem/pycli39/pywbemtools(andy/option-override-design)$ ppp -n mock1 --user new
    Error: General option(s): (--user) not allowed when using a mock server. This can happen when defining a mock server with general options '--name', or '--mock-server'
    (pycli39) kschopmeyer@leonard:~/pywbem/pycli39/pywbemtools(andy/option-override-design)$ ppp -n mock1
    pywbemcli> --user new connection show
    Error: General option(s): (--user) not allowed when using a mock server. This can happen when defining a mock server with general options '--name', or '--mock-server'
    pywbemcli>

I think the same question applies to --verify, certfile keyfile ca-certs.

KSchopmeyer:
After a discussion on 6 September about the current comments we concluded:

1. The "Remove rejection" comment in the changes column for a number of the items is no longer valid.  None of the items listed (ex. --user) reject the option for real connections.  NOTE: They do reject the option for changes to the mock environment because these options are not used for mock connections.
2. The option --pull-max-cnt should become a part of the PywbemServer class to be consistent with the --use-pull option.

Karl created issue #1193 to document the proposed changes from this PR as issues and specifically note that the --pull-max-cnt is one of those issues.

KSchopmeyer:
DISCUSSION: Can we update the table in this PR and close the pr

andy-maier:
Updated the table as discussed.

Issue #1184, opened 2022-07-21
Enumerate Instance with Multiple Namespaces


FarmerMike252:
I use Enumerate Instance with multiple namespaces (-n interop,device) to find where instances of a class exist. Currently, if the class only exists in one namespace I get an exception and am bounced out of my interactive session. What I would prefer is that enumerate list the instances it found and not raise an exception.
If the results are in only one namespace, I would still like to get an indication of the namespace where it was found. If not, then a warning message saying which namespace(s) had no instances would be nice.

PR #1179, opened 2022-06-20
Fix issue using instance get with multiple namespace (-n) option

KSchopmeyer:
NOTE:  waiting on M. Walker test but ready for review.

The instance get command with multiple namespaces (-n
root/cimv2,root/cimv3) and the .? flag to display namespaces caused an
exception.  However, this lead to realizing that we were only displaying
a single instance rather than the instances from each namespace.
Corrected those issues.

I did not create an issue since this is an issue with the code in a prior PR for the the current release and the code only exists for that release.

PR #1186, opened 2022-07-28
Modify error handling on commands using --namespace with multiple namespaces

KSchopmeyer:
This PR based on pr #1179. and fixes issue #1184

The original code always returned an exception for any error on any
namespace when multiple namespaces were requested for instance get,
enumerate, references, associations.

Issue #1184 requested that what was found be returned along with warning
about errors with any of the server requests for multiple namespace
processing unless it was with all the requests or interpreted as a
error beyond the scope of a particular request .

This pr modifies the returns for the instance get, enumerate,
references, and associators table and mof views to return a warning for
each case where the request ended in exception related to existence of
the target item and then return all items found.

This is done for exceptions like NOT_FOUND, INVALID_CLASS,
INVALID_PARAMTER, and INVALID_NAMESPACE. All other exceptions are
treated as complete failure.


andy-maier (approved):

andy-maier (changes requested):
Please rebase on latest master.


KS: I will but am also modifying code to make the multiple error output match the data output (ex. table, mof comments, etc.) per the discussion from the other day.

KSchopmeyer:
Andy proposed that we  could commonize the error handling for the multi-ns calls so that the the multiple errors would appear in a form similar to the data output.   I have largely done this and am resubmitting this pr hopefully today.

PR #1189, opened 2022-08-22
Add object ordering to multi-namespace requests to allow ordering by object name

KSchopmeyer:
Add option to sort the output for the class/instance/qualifier enumerate/get/associator/reference commands from the default order of namespace, object id to order the output by object id first so that objects can be compared between namespaces.

See the commit for more information

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -36,7 +36,7 @@ Released: not yet
 * Increased minimum version of Click to 8.0.1 on Python >= 3.6 to prepare for
   new features. Adjusted testcases accordingly.
 
-  Extended class and instance enumerate/get/associators/references to allow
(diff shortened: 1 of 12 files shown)


andy-maier (changes requested):
Karl, after PR #1186 was merged, it turned out that this PR here contains the 2 commits from that other PR, which go away when rebasing it on master. However, one of the wo remaining commits has exactly the same log message as one of the commits from the other PR.

Could you please rebase this PR to master and double check whether it contains what you want.

KSchopmeyer:
1 Rebasing cleaned up.
2. Reviewed and removed one change that was not relevant.

andy-maier (approved):

PR #1201, opened 2022-10-02
Ignore security issue

KSchopmeyer:
Bypass security issues for"

1. cryptography < 39.0.0

It is development support so I added to the ignore list. This was before rebase to use PR #1199

| REPORT                                                                       |
| checked 238 packages, using default DB                                       |
+============================+===========+==========================+==========+
| package                    | installed | affected                 | ID       |
+============================+===========+==========================+==========+
| cryptography               | 3.3       | <39.0.0                  | 51159    |
+==============================================================================+
| Cryptography 39.0.0 drops support for C library "LibreSSL" < 3.4, as these   |
| versions are not receiving security support anymore.                         |
+==============================================================================+
| cryptography               | 3.2.1     | <39.0.0                  | 51159    |
+==============================================================================+
| Cryptography 39.0.0 drops support for C library "LibreSSL" < 3.4, as these   |
| versions are not receiving security support anymore.                         |
+==============================================================================+
| cryptography               | 3.4.7     | <39.0.0                  | 51159    |
+==============================================================================+
| Cryptography 39.0.0 drops support for C library "LibreSSL" < 3.4, as these   |
| versions are not receiving security support anymore.                         |
+==============================================================================+

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,6 +302,9 @@ test_log_file := test_$(python_version_fn).log
 # - 50571 dparse (user safety) 0.4.1 -> 0.5.2, 0.5.1 -> 0.5.2.  ReDos issue
 # - 50885 Pygments 2.7.4 cannot be used on Python 2.7
 # - 50886 Pygments 2.7.4 cannot be used on Python 2.7
+# - 51159 cryptography  v 39.0.0 drops support for C library "LibreSSL" < 3.4
+# - 50885 pygments Lexer infinite loop in versions Pygments versions 1.5 to 2.7.3
+# - 50886 pygments regular expressions vulnerable to ReDos, versions >=1.1,<2.7.4
 
 safety_ignore_opts := \
 	-i 38100 \
@@ -346,6 +349,8 @@ safety_ignore_opts := \
 	-i 50571 \
 	-i 50885 \
 	-i 50886 \
+
+#	-i 51159 \
 
 ifdef TESTCASES
   pytest_opts := $(TESTOPTS) -k $(TESTCASES)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -90,6 +90,9 @@ Released: not yet
 * Fixed tests that fail because XML output of classes and qualifier declarations
   return attributes not ordered before python version 3.8. (see issue #1173).
 
+* Added new entries to Makefile security ignore list for new security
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -302,6 +302,9 @@ test_log_file := test_$(python_version_fn).log
 # - 50571 dparse (user safety) 0.4.1 -> 0.5.2, 0.5.1 -> 0.5.2.  ReDos issue
 # - 50885 Pygments 2.7.4 cannot be used on Python 2.7
 # - 50886 Pygments 2.7.4 cannot be used on Python 2.7
+# - 51159 cryptography  v 39.0.0 drops support for C library "LibreSSL" < 3.4
+# - 50885 pygments Lexer infinite loop in versions Pygments versions 1.5 to 2.7.3
+# - 50886 pygments regular expressions vulnerable to ReDos, versions >=1.1,<2.7.4
```
andy-maier, line 307: We should only put those safety issues on this list that we add to the ignore list of the safety command.

Plus, the comment should say why we cannot upgrade to the version that solves the safety issue, and not repeat the safety issue.

Plus, we should always try to upgrade to the version that fixes the issue, unless we cannot do that for some Python version. Which means the minimum-constraints.txt file and the requirements file should be updated to specify the package version that fixes the safety issue. If needed, this needs to be done dependent on the Python version.

Makefile
```diff
@@ -347,6 +350,8 @@ safety_ignore_opts := \
 	-i 50885 \
 	-i 50886 \
 
+#	-i 51159 \
```
andy-maier, line 353: This safety issue is commented out. Did you want to add this safety issue to the ignore list?

andy-maier (changes requested):
See comments

Makefile
KSchopmeyer: I was testing without that change just  to be sure.

KSchopmeyer:
Apparently the cryptography safety issue has been removed from the October list since it is no longer causing a failure.  That would be logical since the version of cryptography  defined as the minimum has not yet been releases. Thi safety issue defines 39.0.0 as the minimum and as of today 38.0.1 is latest release.

KSchopmeyer:
Apparently the safety issue for cryptography and been resinded.  Logical since the new minimum version defined in the issue text has not been release.

Therefore I am closing this issue for now.

Issue #1205, opened 2022-10-10
Test failures with "instance shrub" command


andy-maier:
From https://github.com/pywbem/pywbemtools/actions/runs/3214416794/jobs/5254794418:

```
E       Traceback (most recent call last):
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/bin/pywbemcli", line 33, in <module>
E           sys.exit(load_entry_point('pywbemtools', 'console_scripts', 'pywbemcli')())
E         File "/home/runner/work/pywbemtools/pywbemtools/pywbemtools/_click_extensions.py", line 128, in __call__
E           return self.main(*args, **kwargs)
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/core.py", line 1055, in main
E           rv = self.invoke(ctx)
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
E           return _process_result(sub_ctx.command.invoke(sub_ctx))
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
E           return _process_result(sub_ctx.command.invoke(sub_ctx))
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
E           return ctx.invoke(self.callback, **ctx.params)
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/core.py", line 760, in invoke
E           return __callback(*args, **kwargs)
E         File "/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
E           return f(get_current_context().obj, *args, **kwargs)
E         File "/home/runner/work/pywbemtools/pywbemtools/pywbemtools/pywbemcli/_cmd_instance.py", line 703, in instance_shrub
E           context.execute_cmd(lambda: cmd_instance_shrub(context, instancename, options))  # noqa: E501
E         File "/home/runner/work/pywbemtools/pywbemtools/pywbemtools/pywbemcli/_context_obj.py", line 319, in execute_cmd
E           cmd()  # The pywbemcli command function call.
[...]

PR #1212, opened 2022-10-16
Fixed install of incorrect certifi version on Python 2.7

andy-maier:


andy-maier:
The test failure is due to issue #1205 . I suggest to merge this PR and then rebase PR #1206 that solves that issue, then we should be good again.

PR #1208, opened 2022-10-12
fix backport issues 1205 tabulate and #1203 pywbem 1.5.0 compatibility

KSchopmeyer:
Fixes two issues #1203 and #1205.

PR #1213, opened 2022-10-16
1.0: Fixed pywbemcli TypeError with tabulate 0.9.0

andy-maier:
This fixes issue #1205 for version 1.0.1.
Note: It does not roll back the changes in PR #1206 due to too many code changes between the versions, but still fixes the same issue.


andy-maier:
Abandoned since it duplicates PR #1208.

PR #1221, opened 2022-10-24
Restrict some pkg versions in rtd-requirements to avoid missing bullets.

KSchopmeyer:
The documentation output is not including the bullets in bullet lists. It prints the lists as if they were separate paragraphs.  Since the other instance of this were tied to specific issues of sphinx-rtd-theme and docutils, we bypassed the specific package versions.

**Note**: it is not sure this resolves the problem and we will not know until it is merged to master but seems logical and in accord with the bugs documented by docutils and sphinx-rtd-theme.

NOTE: This does not modify the version in dev-requirements.txt because these work correctly.


andy-maier (approved):
Approved.
Will merge it once the test runs succeed, so get an RTD build from master.

PR #1224, opened 2022-10-26
Set copyright in RTD documentation

andy-maier:


KSchopmeyer:
Logical since this statement showed only  something like © Copyright . Revision 57a46f6b.

andy-maier:
Well, do you know what the rule is for having two copyright holders in a single copyright statement?

KSchopmeyer:
It appears that they just append what you have in the string except that the put © Copyright ahead of it so I would assume that if we made the statement

copyright - "2022, IBM Corp. Inova Development Inc"

Note that the last period is missing because apparently they insert it.

This from https://github.com/readthedocs/sphinx_rtd_theme/issues/828 so take with grain of salt and of, course, we won't know until it gets into the master.

KSchopmeyer:
Ran quick test  using local make buildoc:

changed line in conf.py

    #copyright = u''

to

    copyright = u'2022, IBM Corp. Inova Development Inc'

This produced roduced in local build  ad the foot of each page

   © Copyright 2022, IBM Corp. Inova Development Inc.

Of course this does not prove that the ReadTheDocs build will do the same thing since the local build outputs just The word Copyright and no guid without this change.

andy-maier:
I ended up enabling the copyright with the two companies in one line, separated by comma:
```
© Copyright 2017-2022 IBM Corp., Inova Development Inc..
```

PR #1238, opened 2022-11-09
Changed development status to 'Production/Stable'

andy-maier:


KSchopmeyer (approved):
I had in my mind a question about why we are applying production/stable status to the master branch but this becomes clearer when one realizes that the primary use of the classifiers is for PyPi package reporting.

PR #1246, opened 2022-12-13
 Clarified packages needed to issues reported by safety (split out Ubuntu/Python version issue into separate PR)

KSchopmeyer:
CI tests using  ubuntu.latest (which recently was updated to ubuntu 22.04)  return an error for some tests where the os is ubuntu.latest because this version of python and/or the github action setup-python do not support python <= 3.6. This pr modifies the matrix for the github to define use of ubuntu 20.04 for tests that define use of python version <= 3.6.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,7 +310,7 @@ test_log_file := test_$(python_version_fn).log
 # - 50571 dparse (user safety) 0.4.1 -> 0.5.2, 0.5.1 -> 0.5.2.  ReDos issue
 # - 50885 Pygments 2.7.4 cannot be used on Python 2.7
 # - 50886 Pygments 2.7.4 cannot be used on Python 2.7
-# - 51499 Wheel CVE fix in version 0.38.0 yanked after release
+# - 51499 Wheel CVE fix in version 0.38.1 fixes; cannnot be used on python 2.7
 # - 51358 Safety, before 2.2.0 uses dparse with issue, python 2.7 max is 1.9.0
 # - 51457 py - Latest release has this safety issue i.e. <=1.11.0
 safety_ignore_opts := \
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -97,19 +185,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - name: Create local cache directory for Docker images
-      if: ${{ matrix.os == 'ubuntu-latest' }}
+      if: ${{ matrix.os == 'ubuntu-20.04' }}
```
andy-maier, line 188: I think this step should run if on both latest and 20.04, i.e.

```
if: ${{ matrix.os in ('ubuntu-latest', 'ubuntu-20.04') }}
```

Occurs multiple times.

andy-maier (changes requested):
See comment.

.github/workflows/test.yml
andy-maier: DONE

.github/workflows/test.yml
```diff
@@ -33,20 +33,108 @@ jobs:
...
+            \"exclude\": [ \
+              { \
+                \"os\": \"ubuntu-latest\", \
+                \"python-version\": \"2,7\", \
+                \"package_level\": \"latest\" \
+              }, \
+              { \
+                \"os\": \"ubuntu-latest\", \
+                \"python-version\": \"2.7\", \
+                \"package_level\": \"minimum\" \
```
andy-maier, line 55: Python 2.7 is supported on ubuntu-latest.

andy-maier (changes requested):
See comment

.github/workflows/test.yml
andy-maier: DONE

andy-maier:
I have split out the fixing of Ubuntu versions for Python 3.5 and 3.6 into PR #1248 , and scoped this PR to fixing the security issues.

PR #1251, opened 2022-12-22
Support for Python 3.11

andy-maier:


andy-maier:
The test runs on Python 3.11 fail because a mock script that intentionally has a syntax error is no longer handled correctly by the testcase.

KSchopmeyer (commented):
Comment: I am happy with everything except that ubuntu:latest minimum  if failing in documentation.   When that is fixed consider this approved.

andy-maier:
Fixed the docs build issue.

PR #1260, opened 2023-01-07
Fix issue with instance table column ordering

KSchopmeyer:
Fix issue where -o table instance get was not correctly ordering the …table columns

When a property list is provided, pywbemcli is expected to order the columns in the table output in the same order as the order of properties in the --pl option.  This works for enumerate, etc. but not for the instance get command.  One option was missing from the call that sets up the display so that the propert_list option was not being passed to cim_object_display.

Corrected and tests added for proper ordering of the table columns.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -18,6 +18,11 @@ Released: not yet
 
 **Bug fixes:**
 
+* Fix issue where "instance get" was not properly ordering the columns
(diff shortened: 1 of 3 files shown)


andy-maier:
Karl wants to add more tests.

andy-maier (approved):

PR #1272, opened 2023-02-06
Add a new command that shows help subjects of general interest

KSchopmeyer:
Waiting on merge of pr #1274 to make a final change to this one.

This was created in response to request for a special help in pywbemcli for the shell tab completion functionality so the user did not have to go to the docs to understand how to activate the shell tab completion.

It includes a table that defines the help for any number of subjects and the command definition and and command action method to display either a list of the help subjects or the help for any given subject.

This replaces the help function for repl and make repl one of the subjects for this command.

Right now the help includes repl and instancename (instance command group)

This was originally part of the tab-completion pr but separated out to create smaller pr

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -48,6 +48,12 @@ Released: not yet
 
 * Increased the minimum version of pywbem to 1.6.0. (issue #1244)
 
+* Add a new command that will display text on subjects that have been defined
+  for the command.  This allows defining help for subjects that are not
+  specific to a particular command.  This is created specifically to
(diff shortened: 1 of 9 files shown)


andy-maier (approved):

KSchopmeyer:
Andys comments fixed.  I marked two areas for future change 1) move all of repl to separate module.  Each command or command group should be in its own module. 2) we currently have help from repl --help and from repl help.  There should be a single help test.  But I want to get this work committed.

PR #1287, opened 2023-03-05
1.1: changes for new pylint issues with new pylint version

andy-maier:
Rolls back PR #1271 into 1.1.2.


andy-maier:
Since we are about to release 1.2.0, I am dropping this rollback PR.

PR #1300, opened 2023-05-14
Add .safety-policy.yaml file.

KSchopmeyer:
Based on PR # 3002, urllib3 version limit PR to pass tests. Waiting on commit of that PR

Add a safety policy file (.safety_policy.yaml) to replace the list of safety-ignores in the Makefile. This file based on the file in pywbem.

Update the safety ignore entries to account for the new safety issues introduced in May 2023

Note: the safety issue 52510, on Future module was removed from the ignore list since the new version (0.18.3) of that package that fixes the issue has been released in January 2023.  (pywbem issue #2976)

This pr includes changes to both minimum-constraints.txt and dev-requirements.txt to reflect the new safety issues.

NOTE: We  modified the minimum-constraints.txt and, etc. to further separate out the requirements that are specific for python 2.7 and 3.5 in to individual statements in many cases to maybe make it easier when we get to removing requirements for these python versions.

### DIFF
--- a/.safety-policy.yml
+++ b/.safety-policy.yml
@@ -1,0 +1,79 @@
+# Safety policy file
+# For documentation, see https://docs.pyup.io/docs/safety-20-policy-file
+
+# Configuration for the 'safety check' command
+security:
+
+    # Ignore certain severities.
+    # A number between 0 and 10, with the following significant values:
+    # - 9: ignore all vulnerabilities except CRITICAL severity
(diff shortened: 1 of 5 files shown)


KSchopmeyer:
The current failure is because of the problem defined in issue #1301, using urllib3.

andy-maier (approved):
Needs to wait for PR #1303 to be merged and then rebased.

PR #1303, opened 2023-05-16
Ks/set urllib3 limit

KSchopmeyer:
See commit for details.

**NOTE:**  PR #1300, (safety policy)   is based on this PR and  this one will always fail tests because it does not have the changes to the satety policy

Since the only issue is in a single test program it is questionable whether rollback is needed for this change.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -28,6 +28,16 @@ colorama>=0.4.5; python_version >= '3.5'
 importlib-metadata>=0.22,<5.0.0; python_version <= '3.7'
 importlib-metadata>=1.1.0; python_version >= '3.8'
 
+# urllib3 version < 2.0.0 required for python < 3.7
+urllib3>=1.25.9,<2.0.0; python_version == '2.7'
+urllib3>=1.25.9,<2.0.0;  python_version == '3.5'
+urllib3>=1.25.9,<2.0.0;  python_version == '3.6'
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -28,6 +28,12 @@ colorama>=0.4.5; python_version >= '3.5'
 importlib-metadata>=0.22,<5.0.0; python_version <= '3.7'
 importlib-metadata>=1.1.0; python_version >= '3.8'
 
+urllib3>=1.25.9,<2.0.0; python_version >= '2.7' and python_version <= '3.6'
+# TODO: The following limits use of urllib <= 2.0 until issues about this upgrade
```
andy-maier, line 32: I don't understand the separation between the python versions 2.7-3.6 and 3.7-3.9. The Todo comment refers to the following lines, but also applies to the previous line.

Can we merge the two sets of Python versions?

andy-maier (changes requested):

dev-requirements.txt
KSchopmeyer: This is same as pywbem.   Urllib3 version to has minimum python requirement of >= '3.7'. Thus we define a max for python 3.6 and then made the comment and the changes for python 3.7 on separate line since those are the lines where we will remove the <2.0.0 limitation when we fix urllib3.

andy-maier (approved):

PR #1305, opened 2023-05-25
Ks/container indications

KSchopmeyer:
Waiting on PR #1313

See commit message for details:

Fixes issue with pywbemlistener being bound to only "localhost" and adds an option to pywbemlistener start and run to allow the user to define the bind address for each listener created with the default value being 0.0.0.0 for which the listener will receive indications from any network interface and any host/ip address.

Adds new option to pywbemlistener test to define a host name/ip address to which indications will be sent rather than just sending them to localhost.

### DIFF
--- a/pywbemtools/pywbemlistener/_cmd_listener.py
+++ b/pywbemtools/pywbemlistener/_cmd_listener.py
@@ -93,6 +93,17 @@ LISTEN_OPTIONS = [
                  required=False, default=DEFAULT_LISTENER_SCHEME,
                  help=u'The scheme used by the listener (http, https). '
                  'Default: {}'.format(DEFAULT_LISTENER_SCHEME)),
+    click.option('-b', '--bind-addr',
+                 type=str,
+                 metavar='HOST',
+                 required=False,
+                 default=None,
+                 help=u'A host name or IP address to which this listener '
+                      'will be bound. Binding the listener defines the '
+                      'indication destination host name or IP address for '
+                      'which this listener will accept indications. The '
+                      'default accepts indications addressed to any '
+                      'network interfaces on the listener system.'),
     click.option('-c', '--certfile',
                  type=click.Path(exists=False, dir_okay=False),
                  metavar='FILE',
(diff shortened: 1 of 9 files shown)


andy-maier:
On the discussion: Given that we have quote a number of issues open for 1.3.0, I suggest backporting this whole PR.
i have set the rollback needed label.

If you agree, please rebase the PR and squash the 2 commits into one.

KSchopmeyer:
Updated to one PR.  My only issue with rollback is that we are actually adding a parameter to the command line that this something we do not normally do in a minor release.

PRO: 
1. Fixes issue that disallows anything but localhost listeners.

CON: 
1. Adds functionality and a new Option in a rollback which we normally do not do.
2. No user complaints so apparently not user issue.

I will go either way with the rollback.

andy-maier:
We decided to not roll it back because it goes beyond a pure bug fix.

pywbemtools/pywbemlistener/_cmd_listener.py
```diff
@@ -449,6 +470,11 @@ def listener_list(context):
               required=False, default=1,
               help=u'Count of test indications to send. '
               'Default: 1')
+@click.option('-l', '--listener', type=str, metavar='LISTENER HOST',
```
andy-maier, line 473: On `metavar='LISTENER HOST'`: The blank suggests these are two parameters. I suggest changing that to just "HOST".

pywbemtools/pywbemlistener/_cmd_listener.py
```diff
@@ -865,7 +892,9 @@ def parse_listener_args(listener_args):
     # Note: The following options must ne in sync with the Click command options
     parser.add_argument('--start-pid', type=int, default=None)
     parser.add_argument('--port', '-p', type=int,
-                        default=DEFAULT_LISTENER_PORT)
+                        default=None)
```
andy-maier, line 895: I don't understand this change of the port default to `None`. Normally when a default of `None` is set, there is some processing of the value None to handle this case with code, but I don't see that.

andy-maier (changes requested):

pywbemtools/pywbemlistener/_cmd_listener.py
KSchopmeyer: Done

pywbemtools/pywbemlistener/_cmd_listener.py
KSchopmeyer: Actually either works since the default value is set in the click processing but this more clearly makes the point.  Fixed

andy-maier (approved):
Approved, but please resolve the merge conflict.

KSchopmeyer:
remarked review needed just to catch your attention.  If you approve, ready for commit except issue with Python 2.7 test and pywbemcli does pass all those tests.

andy-maier (approved):

PR #1308, opened 2023-05-28
Dropped Python 3.5

andy-maier:


KSchopmeyer (commented):
That is a lot of changes to just disable use of a single version of python.  Hate to think what it is going to take to remove python 2.7

One issue:

No entry in changes.rst.

PR #1332, opened 2023-08-28
Investigated hang of 'make test' on py27

andy-maier:


andy-maier:
The failing test on Python 2.7 was disabled in PR #1340 - this PR is no longer needed.

PR #1334, opened 2023-09-08
Add an example directory with example of subscriptions/indications

KSchopmeyer:
The example generates a subscripti on specifically for OpenPegasus provider that generates test indications and requests that OpenPegasus generate indications.

This is specific to OpenPegasus because there is no standard definition for a WBEM Server indication generator

NOTE: This was created to test the indication from the WBEM server in a Docker container but I need it more generally to find out why we are not getting indications out of the container and decided that it may be useful. So I created the examples directory.

This example works and returns indications from the current OpenPegasus v 0.1.3 container.

Andy proposed that we break it into pieces  as separate scripts(ex. Create subscription, Remove subscription, etc. ) I have much of that running but while it may be a simplification to the user, it does make more work for us in defining the options, help for each new script. 

The fact that the current code works and OpenPegasus returns indications  which clarifies that the issues we had in getting indications from the container to work are fixed.

NOTE: We tried to break up the example into separate scripts but found that it was very messy for at least on parameters like the the query which is both a string with spaces and often contains the * character.  Therefore I put in the separate script but did not actually put the call into the example.

NOT DONE.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -76,6 +76,10 @@ Released: not yet
   different address/system than the system where pywbemlistener test is being
   executed. This will allow testing across multiple systems.
 
+* Added new directory ``examples`` that is available when the repository id
+  cloned and provides a first example of indication subscriptions and
+  indication testing using an OpenPegasus indication generating provider.
+
 **Cleanup:**
 
 * Change to used safety-policy-file .safety-policy-yml to keep the safety issue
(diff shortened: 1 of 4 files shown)


andy-maier:
As discussed in our meeting, I think we should have an examples directory, and it does not need to be part of the installed package.

KSchopmeyer:
I left dome cruft in what I committed mast night. Will do update today

andy-maier:
We discussed the concept of example scripts again, and agreed that we should have examples in the documentation rather than putting scripts on top of our commands.

KSchopmeyer:
We agreed that doing scripts to execute

PR #1367, opened 2024-03-18
Split safety runs into 'install' and 'all'

andy-maier:


andy-maier:
Wait for comments in https://github.com/pywbem/pywbem/pull/3124 to be resolved.

KSchopmeyer (commented):
I have exactly the same comments as the pywbem version.

1. Explain what install means in the min...install.txt file. i.e what files.
2. the optional one about putting pipdeptree into Make as target since it should be called through python.
Other than that, approved.

andy-maier:
Same resolutions as for pywbem.

PR #1403, opened 2024-07-29
Removed support for Python < 3.8

### LINKED ISSUE #1390: Remove support for Python<=3.7
Remove support for Python<=3.7
Use `pyupgrade --py38-plus` for the initial transition of the code.

andy-maier:


README.md
```diff
@@ -68,43 +67,12 @@ system as background processes. It provides functionality to:
 
 Requirements:
 
-1.  Python 2.7, 3.6 and higher
+1.  Python 3.8 and higher
```
KSchopmeyer, line 70: Probably not necessary since it is really all python versions allowed for building.

KSchopmeyer (commented):
Make only one small comment.

KSchopmeyer (commented):
Note I did not yet completely review the changes in versions in the requirements.txt files other than the removal of things less than jpython 3.8. But it works so lets commit.

README.md
andy-maier: I think a reference to "all python versions allowed for building" would not be very helpful for users who just want to install the package.

The question is therefore if we don't provide an easy to read statement like the above, where do we point to instead?
 
Note that not everybody installs from Pypi. For example, the package could be made part of a Linux distro (like for pywbem).

I think the price to change this version in the docs is very small, so I would just leave it as it is.

andy-maier: We discussed and concluded to leave the statement in.

PR #1404, opened 2024-07-29
Applied pyupgrade on source files

andy-maier:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -565,23 +565,14 @@ upload: _check_version $(dist_files)
 html: $(doc_build_dir)/html/index.html
 	@echo "Kakefile: Target $@ done."
 
-# Boolean variable indicating that Sphinx should be run
-# We run Sphinx only on Python>=3.8 because lower Python versions require too old Sphinx versions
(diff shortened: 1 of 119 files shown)


KSchopmeyer (commented):
There are some more things we should do before we release the next version  ( ex. more use of f" formatting) but for now I only have two  comments. ex. more use of  f" in place of format

1. The files tests/unit/pywbemcli/test_qualdecl_cmds.py and tests/unit/pywbemcli/test_class_cmds.py both include a test that is exclusivde to ptyhon  >= 3.8.. See the variable 
PYTHON_GE_38 in the test class file.

2. in test_display_cimobjects.py. I think the following try can be removed.
try:
    from collections import OrderedDict
except ImportError:
    from ordereddict import OrderedDict  # pylint: disable=import-error


I propose that we commit this and other things can be cleaned up after this work is committed.

andy-maier:
I think the two suggestions you have can be done in this PR.

andy-maier:
We discussed the remaining version condition flags, and decided to address them in separate PRs:
* https://github.com/pywbem/pywbemtools/issues/1406
* https://github.com/pywbem/pywbemtools/issues/1405
* https://github.com/pywbem/pywbemtools/issues/1407

PR #1408, opened 2024-08-11
Convert format() string formatting to f" formatting.

KSchopmeyer:
See commit for details.

Marked for review despite failure of one test in test-subscriptions.py.  No commit before issue fixed.

### DIFF
--- a/docs/pywbemcli/mock_support.rst
+++ b/docs/pywbemcli/mock_support.rst
@@ -902,8 +902,8 @@ of the target CIM instance and returns that property in an output parameter
                 if isinstance(localobject, CIMClassName):
                     raise CIMError(
                         CIM_ERR_INVALID_PARAMETER,
-                        "CIM method {0} must be invoked on a CIM instance".
-                        format(methodname))
+                        f"CIM method {methodname} must be invoked on a "
+                        "CIM instance")
                 return self.Method1(localobject, params)
             else:
                 raise CIMError(CIM_ERR_METHOD_NOT_AVAILABLE)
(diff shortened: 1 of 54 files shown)


KSchopmeyer:
I did not modify a number of uses because they are significantly complex and some others because the code is
actually easier to read with the format statements.  Also only modified code to date, not tests.

Fixed the click.echo(), most of click.ClickException() and a number of other uses for total of about 180 changes.  However with all the various usages there are still not changed to f" firnatting including:
* complex usage
* where a \ would be in the format string(not allowed in python 3.8 - 3.10

currently fixed
150  of total format( involved statements of over 450 existing.  About 200 of them are in tests directories.

KSchopmeyer:
Note.  Having a problem with one single test in tests_subscriptions.py.  That test reports invalid connection file.  Failure occurs one the connection test in pywbemcli.py

* fails when full test suite run.
* Fails when full subscription tests run.
* Fails when that single test run after running subscriptions tests.
* Same issue when all changes removed except for subscriptions.py changes.
* Ran test where running just this test fails but in another terminal on same virtual environment and test passes.

docs/pywbemcli/mock_support.rst
```diff
@@ -902,8 +902,8 @@ of the target CIM instance and returns that property in an output parameter
                 if isinstance(localobject, CIMClassName):
                     raise CIMError(
                         CIM_ERR_INVALID_PARAMETER,
-                        "CIM method {0} must be invoked on a CIM instance".
-                        format(methodname))
+                        f"CIM method {methodname} must be invoked on a "
+                        f"CIM instance")
```
andy-maier, line 906: The "f" prefix should not be there on this line.

KSchopmeyer: Done

andy-maier (approved):

PR #1413, opened 2024-08-25
Remove unneeded python 2.7 functionality and configuration info.

KSchopmeyer:
Removes some items that were required for older versions of python (pre 3.8 min version) but that were not caught in the initial removal:

* _click_extensions.py (TabCompletion class and TabCompletionArgument) Remove ref to TabCompletionClass from code
* _displaytree.py -  just a comment removed
* pywbemcli.py - Remove some python 2.7 tests in repl _cmd_listener - commane about tz parameter
* pywbemlistener.py - Python 2.7 test
* requirements.txt - some python 2.7 comments
* Pytest_extensions.py - Manual setup of simplified test signature.

While this pr could be considered a response to issue 1405, it was put together and tested before I started on that issue. Therefore it is submitted as an independent pr

### DIFF
--- a/pywbemtools/_click_extensions.py
+++ b/pywbemtools/_click_extensions.py
@@ -2,8 +2,6 @@
 This file contains extensions to Click for pywbemtools.
 """
 
-import sys
-from collections import OrderedDict
 import packaging.version
 
 import click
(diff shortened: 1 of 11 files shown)


KSchopmeyer:
close this pr since the code changes in this pr have been incorporated into pr #1416 since they apply to the same issue.

KSchopmeyer:
pr closed

PR #1424, opened 2024-09-22
Refactor pywbemcli tests to avoid hiding default conn file.

KSchopmeyer:
We were hiding any local user default connection file during tests by changing its name during the tests.  This code modifies pywbemcli to allow defining an alternate directory for all of the connection files (default_connection_file and mockcache directory) using an environement varaible  for tests thus completely avoiding any existing files.

This required changes to the clitestextensions.py file and also to a number of test files.


KSchopmeyer:
This pr retracted in favor of a new one.Closing

PR #1426, opened 2024-09-23
Add too-many-positional-arguments to list of ignored pylint msgs.

KSchopmeyer:
Just puts too-many-positional-arguments into the list of pylint messages to be ignored at least temporarily.

This new test generates about 70 new pylint messages when pylint is executed against pywbemtools.


KSchopmeyer:
Close as duplicate.

PR #1434, opened 2024-10-14
Upgraded Python 3.13 to final GA version

andy-maier:


KSchopmeyer (approved):

KSchopmeyer:
I noted that as with pywbem, python 3.13 does not show up in in the Readme pyversion shield.

andy-maier:
On the Python versions badge: The versions are not defined in the README file:
```
[![Supported Python](https://img.shields.io/pypi/pyversions/pywbemtools.svg?color=brightgreen)](https://pypi.python.org/pypi/pywbemtools/)
```
shields.io gets them from Pypi, and I assume that shields.io interprets the Trove classifiers for Python in the package on Pypi. There, the version that supports 3.13 is not yet released. I assume that we will see 3.13 once pywbem with 3.13 support is released.

KSchopmeyer:
I sort of assumed that was the case.  Nothing we can do about it unless we were to find another shield.  

I approve this one

PR #1437, opened 2024-10-27
Branch to just test the master

KSchopmeyer:
Executed this test to confirm if problem in another branch today with the install in windows.

This is just the current master branch so the failures in windows install are not the changes for the little py 2.7 branch


andy-maier:
This issue was fixed with PR #1438

KSchopmeyer:
Closed since the issue was found and this was just a test of master.

PR #1450, opened 2025-01-29
1.3: Added twine deps

andy-maier:
Rolls back PR #1449 into 1.3.1, but differently.


andy-maier:
Abandoning this PR, because pywbemtools 1.3 has not yet introduced support for the publish workflow, so it depends on the twine-based upload.

andy-maier:
Replaced by PR #1451

PR #1453, opened 2025-02-20
Added new entries to minimum-constraints-develop.txt

KSchopmeyer:
Fails make check_reqs.

Added entries for packages python-Levenshtein, Levenshtein, and roman-numerals-py to minimum-constraings-develop.txt to eliminate errors from make check-reqs

### DIFF
--- a/changes/noissue.6.fix.rst
+++ b/changes/noissue.6.fix.rst
@@ -1,0 +1,4 @@
+Development: Fix issue with minimum-constraints-develop.txt that causing
+failure of make check_reqs. Added two pkgs to minimum-constraints-develop.txt:
+package Levenshtein used by safety and Sphinx and roman-numerals-py used by
+Sphinx. See PR 1453 for details.
--- a/minimum-constraints-develop.txt
+++ b/minimum-constraints-develop.txt
@@ -33,6 +33,8 @@ coveralls==3.3.0
 # Safety CI by pyup.io
 safety==3.1.0
 safety-schemas==0.0.2
+# The following used by safety and Sphinx
+Levenshtein==0.25.1
 # TODO: Change to dparse 0.6.4 once released
 dparse==0.6.4b0
 ruamel.yaml==0.17.21
@@ -71,6 +73,7 @@ sphinxcontrib-serializinghtml==1.1.9; python_version >= '3.9'
 sphinxcontrib-websupport==1.2.4
 autodocsumm==0.2.12
 Babel==2.11.0
+roman-numerals-py==1.0.0 ; python_version >= '3.9'
 
 # PyLint (no imports, invoked via pylint script)
 pylint==3.0.1; python_version == '3.8'


minimum-constraints-develop.txt
```diff
@@ -71,6 +74,7 @@ sphinxcontrib-serializinghtml==1.1.9; python_version >= '3.9'
 sphinxcontrib-websupport==1.2.4
 autodocsumm==0.2.12
 Babel==2.11.0
+roman-numerals-py==1.0.0
```
andy-maier, line 77: roman-numerals-py 1.0.0 requires Python>=3.9, and is used by Sphinx>=8.2.0 which also requires Python>=3.9. Since the constraints are only used when the package is actually installed, it does not matter that the constraint has no python version marker. However, that prevents using the constraints file as a requirements file, and I think it would be more clear to add a python version marker: `; python_version >= '3.9'`.

changes/noissue.6.fix.rst
```diff
@@ -0,0 +1,4 @@
+Development: Fix issue with minimum-constraints-develop.txt that causing
+failure of make check_reqs. Added two pkgs to minimum-constraints-develop.txt:
+package Levenshtein  and python-Levenshtein used by safety and roman-numerals-py
+used by sphinx
```
andy-maier, line 4: The "python-Levenshtein" package has been added by safety 3.3.0 and will be removed again in safety 3.3.1 because it is GPL-licensed. See https://github.com/pyupio/safety/issues/682. Since "safety" is the only package using "python-levenshtein" (or its renamed package "levenshtein"), I suggest fixing this issue by excluding safety 3.3.0 instead of adding any of the "levenshtein" packages.

andy-maier (changes requested):

minimum-constraints-develop.txt
```diff
@@ -33,6 +33,9 @@ coveralls==3.3.0
 # Safety CI by pyup.io
 safety==3.1.0
 safety-schemas==0.0.2
+# The following used by safety
```
andy-maier, line 36: add "and Sphinx"

minimum-constraints-develop.txt
```diff
@@ -33,6 +33,9 @@ coveralls==3.3.0
...
+python-Levenshtein==0.25.1
```
andy-maier, line 37: I suggest to not add the "python-Levenshtein" dependency. For the reasons, see comment above.

changes/noissue.6.fix.rst
KSchopmeyer: Done

minimum-constraints-develop.txt
KSchopmeyer: Done

minimum-constraints-develop.txt
KSchopmeyer: Done

andy-maier (approved):

PR #1458, opened 2025-03-30
Dev: Simplified the release of a version and the start of a new version

andy-maier:
No review needed.


andy-maier:
Failures are caused by coveralls.io being in maintenance.

Issue #1467, opened 2025-08-18
Unit test: listener test_list_cmd.py fails on macOS


andy-maier:
They started failing on the last weekend run (2025-08-27). The week before that, they succeeded.

Test module: test_list_cmd.py

Runs with just these failures:
* 2025-08-18: https://github.com/pywbem/pywbemtools/actions/runs/17033831825
* 2025-09-28: https://github.com/pywbem/pywbemtools/actions/runs/18075923509

andy-maier:
This has been addressed as part of PR #1501

PR #1477, opened 2025-09-28
Fixed that the --nq/--no-qualifiers was ignored for class operations

andy-maier:
Note: The failures on macOS are subject of issue #1467.

### DIFF
--- a/changes/1476.fix.rst
+++ b/changes/1476.fix.rst
@@ -1,0 +1,2 @@
+Fixed that the --nq, --no-qualifiers option was ignored for class operations,
+and qualifiers were always included.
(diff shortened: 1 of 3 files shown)


KSchopmeyer:
I see that the test for the --no-qualifiers (--nq) attribute in test_class_cmds.py is simply wrong it does test for existence or non-existence of the qualifier and for some reason I combined the assciation filter and --nq together in tests.  Applies to both get and enumerate class command tests.  I should separate these tests by 1: drop the --nq from the tests for association filter, 2: add 2 tests, default value and --nq  and for --n-qualifiers to test for that GetClass and EnumerateClass properly handle that command option

andy-maier:
The failures on macOS are subject of issue #1467.

PR #1504, opened 2026-01-03
Fixed 'socket.gaierror' exception when --bind-addr option was not specified

andy-maier:


KSchopmeyer:
Approved, karl

PR #1600, opened 2026-09-14
Added sections about handling of Dependabot PRs and backporting

andy-maier:
Needs to wait until PR https://github.com/pywbem/pywbem/pull/3483 is approved, and is then also considered to be approved.


KSchopmeyer:
I made one comment on the pywbem version and that could be applied here also.

andy-maier:
PR https://github.com/pywbem/pywbem/pull/3483 is approved - made the same change here.

